PR #17, opened 2019-07-19
Fix two warnings that were occurring during development.

tgreaves:
* outDir deprecated in launch.json  (Removed - not used)
* "the options [safe] is not supported" (Updated monk to version 4)


cryogen:
Thanks!

PR #18, opened 2019-07-19
Added a few Sanctum Cards

Achocholacek:
Added [Sanctum] Bordan the Redeemed
Added [Sanctum] Lion Bautrem
Added [Sanctum] The Grey Rider

(Tested fine locally. Hope I'm doing this right!)

### DIFF
--- a/server/game/cards/01-Core/SelwynTheFence.js
+++ b/server/game/cards/01-Core/SelwynTheFence.js
@@ -5,7 +5,7 @@ class SelwynTheFence extends Card {
         this.fight({
             reap: true,
             target: {
-                activePromptTitle: 'Choose a captured aember to move to your pool.',
+                activePromptTitle: 'Choose a captured amber to move to your pool.',
                 cardCondition: card => card.hasToken('amber'),
                 controller: 'self',
                 gameAction: ability.actions.removeAmber()
--- a/server/game/cards/02-AoA/AubadeTheGrim.js
+++ b/server/game/cards/02-AoA/AubadeTheGrim.js
@@ -8,13 +8,9 @@ class AubadeTheGrim extends Card {
         this.reap({
             condition: context => context.source.hasToken('amber'),
             effect: 'move 1 amber from Aubade the Grim to the pool',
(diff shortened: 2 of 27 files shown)


cryogen:
Thanks!

Can you remove the comments after the card ids?

These are a legacy and should he removed going forwards

server/game/cards/02-AoA/TheGreyRider.js
```diff
@@ -0,0 +1,49 @@
+const Card = require('../../Card.js');
+
+class TheGreyRider extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            optional: true,
```
jeremylarner: you can just pass `fight: true, reap: true` to this.play so you don't need to do the same thing three times

Achocholacek:
Cleaned up comments as requested. Added new cards. Couple non working ones (Aubade and Abond - they both look like they SHOULD work, but no such luck.)

server/game/cards/01-Core/SelwynTheFence.js
```diff
@@ -5,7 +5,8 @@ class SelwynTheFence extends Card {
         this.fight({
             reap: true,
             target: {
-                cardType: 'creature',
+                activePromptTitle: 'Choose a captured aember to move to your pool.',
```
cryogen: If you use 'amber' it will automatically be converted to Aember and a amber icon by the client

server/game/cards/02-AoA/AubadeTheGrim.js
```diff
@@ -0,0 +1,24 @@
...
+
+class AubadeTheGrim extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            gameAction: ability.actions.capture({ amount: 3 })
+        });
+        this.reap({
+            condition: context => context.source.hasToken('amber'),
+            effect: 'move 1 amber from Aubade the Grim to the pool',
+            then:{
```
cryogen: This doesn't need to be a "then", just a straight game action

PR #20, opened 2019-07-21
Add Knuckles Bolton and fix card import process

ripred45676:
**Description**
I initially started this PR trying to get Knuckles Bolton working, but eventually found out that it didn't have skirmish in the DB. Looks like this is a bug from earlier that was just never relevant because there weren't two keywords on different lines for cards previously, so this didn't matter. Here is an example of the test data that goes into parseKeywords

`Elusive. (The first time this creature is attacked each turn, no damage is dealt.)\rSkirmish. (When you use this creature to fight, it is dealt no damage in return.)\r`

So as you can see if you just split on periods you will get an invalid keyword for Skirmish looking somethings like this `)\rSkirmish` so it won't be included in the keywords for the card.

Anyways this was a fun one to start off with, and will start diving into more cards :)

(Note) I filed a separate PR for the keyteki-json-data submodule for the fix and update JSON data

**QA Steps**
Manually confirmed that Knuckles Bolton has both skirmish and elusive working appropriately.


jeremylarner:
Where a card doesn't have any abilities which needs coding, it doesn't need a file at all. So now you've fixed the keyword parsing, Knuckles should work fine without KnucklesBolton.js

cryogen:
I'm going to close this as the fix you've already done in the data repository

As jadiel points out, cards with no text other than keywords dont need an implementation, and the other file you've changed is a deprecated file I'll be removing soon

Thanks for the fix!

ripred45676:
Great sounds good, thanks!

PR #25, opened 2019-07-22
Added some Untamed Cards. Believe to be 100% accurate.

Achocholacek:


server/game/cards/02-AoA/PunctuatedEquilibrium.js
```diff
@@ -0,0 +1,26 @@
+const Card = require('../../Card.js');
+
+class PunctuatedEquilibrium extends Card {
```
ripred45676: Curious, how should this interact with chains?

Achocholacek: That's a great question. 

Because the rulebook doesn't specifically call out a situation like this as they do Mulligans (at least not yet), I'm operating under the assumption that drawing is drawing and you shed a chain like normal if you would have shed a chain at the end of your turn.

Rulebook:
"Each time (including during setup) a player with one or more chains would
draw one or more cards to refill their hand, that player draws fewer cards
(based on their current chain level, see below) and then sheds one chain
by reducing their chain tracker by one."

PR #31, opened 2019-07-22
docker-compose.yml changes to fix issues with Docker on some systems.

tgreaves:
* Moved to version 3.3 syntax.
* Volume adjustments to only use shares as required.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,6 @@ services:
       - "4000:4000"
     volumes:
       - node_modules:/usr/src/lobby/node_modules
-      - img:/usr/src/lobby/public/img
     links:
       - mongo
     depends_on:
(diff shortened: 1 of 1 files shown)


tgreaves:
I dug into this and it seems that the issue lies from having the two Node containers have the same shared volume.  Files are not reliably shared when they are added as part of a Dockerfile.  Specifically, the public/vendor-manifest.json file.

I have changed the docker-compose.yml file so have the share specifically for the public/img directory. This will handle the cards.  I've tested that when the cards are downloaded, both Node containers can see them.

A question I have is what purpose having node_modules as a shared volume has as well, given that the build process has to build everything for each Docker image anyway -- no time is shared.  Thoughts?

tgreaves:
Paging @Achocholacek for comments too!

cryogen:
Nice work tracking this down.  Just two things:

1) The images don't need to be shared so you can take that out
2) The node_modules thing was something I found from the internet, without those lines, it ends up building, and using the wrong version of node_modules (IE for me it builds them on windows and then copies it into the linux container and they don't work)

PR #36, opened 2019-07-23
Added Tantadlin, Panpacas and Heart of the Forest.

Achocholacek:


server/game/cards/02-AoA/PanpacaAnga.js
```diff
@@ -0,0 +1,14 @@
+const Card = require('../../Card.js');
+
+class PanpacaAnga extends Card {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: card => card.controller.cardsInPlay.indexOf(card) > card.controller.cardsInPlay.indexOf(this),
```
cryogen: This works, but is not guaranteed.  As some point in the future we should make this a more guaranteed contract of where the card is on the battle line

PR #38, opened 2019-07-23
Modified Capture and Steal to allow for new effects

Achocholacek:
"limitFromPool" and "captureFromPool" to accommodate Po's Pixies
(If effect is active, captured amber or stolen amber or both comes from general pool rather than targets pool)


cryogen:
Paging @jeremylarner

PR #42, opened 2019-07-24
Add missing steps to build/run via docker

kevindigo:
When I tried to run via docker, it failed. After running those 2 commands, it worked.


cryogen:
Thanks!

PR #43, opened 2019-07-24
Update README to explain how to run the server

kevindigo:


README.md
```diff
@@ -87,6 +87,34 @@ Then for each game node (typically one per CPU/core):
...
+
+If you implement or make changes to a card, you can use manual mode 
+to add it to a deck from within a game. Use manual mode, and the command:
+```
+/add-card <card name>
+```
+
+The unit tests require cross-env to be installed:
+```
+npm install cross-env
```
cryogen: This step is not needed, it should be 'npm install' on its own

README.md
```diff
@@ -87,6 +87,34 @@ Then for each game node (typically one per CPU/core):
...
+### Running and Testing
+
+The game server should be accessible by browsing to localhost:4000. 
+
+You can register 2 or more users, to play against yourself. 
+They can have fake email addresses. 
+You can login as both users either from 2 different browsers, or by 
+using an incognito window. 
+
+There is no concept of an 'admin' user.
```
cryogen: This isn't true, but it's not necessary for development

README.md
kevindigo: Fixed

README.md
kevindigo: Reworded.

kevindigo:
I think I have addressed the PR comments. Let me know if anything else needs to be changed.

PR #47, opened 2019-07-24
Scowlycaper

saluk:


cryogen:
There's a failing test on harland mind lock :(

saluk:
Ruh roh! My guess is that because I flipped left/right back for takeControl, the harland mindlock test is incorrect. I'll check it and make sure.

PR #50, opened 2019-07-24
Implement bonerot venom

cryogen:


README.md
```diff
@@ -47,11 +47,11 @@ If you have docker installed, you can use the containerised version of the site.
 Clone the repository
 git submodule init
 git submodule update
+npm install
 docker-compose up
 
 (in another terminal)
-docker-compose exec lobby bash
```
kevindigo: @cryogen : It looks like a newline accidentally got deleted here. Or was this intentional?

cryogen: Yes, that's intentional, I combined the two steps into one

PR #54, opened 2019-07-25
Couple bug fixes

LifeIsPizza:
- Ritual of Balance is now exhausted when used by an opponent (Nexus, Remote Access) even if the condition isn't met

- Mobius Scroll is now archived in the owner's archives when used by an opponent (Nexus, Remote Access).

- Same as above for Hexpion (which can change controller)

### DIFF
--- a/server/game/cards/01-Core/RitualOfBalance.js
+++ b/server/game/cards/01-Core/RitualOfBalance.js
@@ -6,10 +6,6 @@ class RitualOfBalance extends Card {
             condition: context => context.player.opponent && context.player.opponent.amber >= 6,
             gameAction: ability.actions.steal()
         });
-        this.action({
-            condition: context => context.player.opponent && context.player.opponent.amber < 6,
-            gameAction: ability.actions.exhaust()
-        });
     }
 }
 
--- a/server/game/cards/01-Core/RitualOfBalance.js
+++ b/server/game/cards/01-Core/RitualOfBalance.js
@@ -6,9 +6,13 @@ class RitualOfBalance extends Card {
             condition: context => context.player.opponent && context.player.opponent.amber >= 6,
             gameAction: ability.actions.steal()
         });
+        this.action({
+            condition: context => context.player.opponent && context.player.opponent.amber < 6,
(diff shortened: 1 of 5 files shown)


server/game/cards/01-Core/RitualOfBalance.js
```diff
@@ -6,9 +6,13 @@ class RitualOfBalance extends Card {
             condition: context => context.player.opponent && context.player.opponent.amber >= 6,
             gameAction: ability.actions.steal()
         });
+        this.action({
```
jeremylarner: I don't think this is the correct way to fix this bug.  The problem is that the game won't allow the artifact to be used when the action condition isn't met, whereas it should allow the ability to be used, and just not permit the game action to resolve

test/server/cards/01-Core/MobiusScroll.spec.js
```diff
@@ -1,4 +1,4 @@
-describe('Mobius Scroll', function() {
+fdescribe('Mobius Scroll', function() {
```
cryogen: please remove the 'f' here as it means all test wont run

LifeIsPizza: Sorry forgot to revert it back

server/game/cards/01-Core/RitualOfBalance.js
LifeIsPizza: I'm sure there's a more clever and cleaner way to do it... Not sure how to put a condition inside the gameAction though, as there are no other cards that work in a similar way.
On top of my head, stuff like Anomaly Exploiter with no damaged creatures should exhaust just fine as the condition is on the selection...

jeremylarner: Hmmm, I'll take a look

cryogen:
@LifeIsPizza is this PR still valid?

There's a failing test (which might be correct, but ideally it needs fixing :) )

cryogen:
@jeremylarner Is this something you're still looking at?

PR #56, opened 2019-07-25
add Collector Worm, refs #6

issa:


issa:
Removed the focus in the spec

PR #62, opened 2019-07-26
Fix drummernaut according to match errata

ripred45676:
Just noting that I was not able to test this locally because of some issues logging in and registering, but thought I might as well just add the code for now.


Achocholacek:
Hideaway hole is incorrect. You're removing the keyword elusive, not adding it, and 'forRemainderOfTurn' only lasts until the end of the current players turn - Hideaway hole is meant to last until the start of the current player's *next* turn.

(I did the same thing - we will likely need to adjust the lastingEffect/forRemainderOfTurn code)

ripred45676:
Woops good call. Will remove Hideaway Hole because I saw you just but a PR up. That way can just confine this to the drummernaut update

ripred45676:
Took out hideaway hole, so this just fixes Drummernaut according to the errata

PR #64, opened 2019-07-26
Mars Needs Amber, Equalize and minor fixes

LifeIsPizza:


LifeIsPizza:
Edited ReturnAmber to fit Equalize, added a specification in who gets which amber.

Test passing should demonstrate it works properly, even for the only other card that used ReturnAmber (Word of Returning)

PR #69, opened 2019-07-27
Tyxl beambuckler

EmbraCraig:
Consider this a draft as currently not working.

I've attempted to create a MoveToFlank action, based on the play a card action and swap actions.

When attempting to use moveToFlank() in the same way as swap() is used (see core card SanctumGuard for a comparison), then I get an error moveToFlank() is not a function.

There are probably more problems here as I'm attempting to hack this together from what I see in other parts...


cryogen:
I've incorporated this into https://github.com/keyteki/keyteki/commit/e3325e36a6b2ed55db4adf8cb2b2bfe8a1631222

Thanks!

Issue #76, opened 2019-07-27
MoveToFlank action not working correctly


EmbraCraig:
Tyxl Beambuckler and Yxlyx Stimrager both show same symptoms:

Rather than moving the targeted card, it's replacing the card on the active player's chosen flank - target card doesn't move at all

PR #77, opened 2019-07-27
Fixing issues with targetting of MoveToFlankAction; removing unnecess…

EmbraCraig:
…ary if statement (player whose card is chosen as a target for move to flank must have cards in play) and associated code.

Fixes PR #76


server/game/GameActions/MoveToFlankAction.js
```diff
@@ -24,28 +24,24 @@ class MoveToFlankAction extends CardGameAction {
         super.preEventHandler(context);
 
         let card = this.target.length > 0 ? this.target[0] : context.source;
```
jeremylarner: This line needs removing - lint doesn't like unused variables

Issue #154, opened 2019-07-28
Fetchdrones only captures 1 for each discarded logos card


ProntoBollax:

Achocholacek:
pr submitted. will be fixed next code push.

PR #156, opened 2019-07-28
Fetchdrones fix to capture 2A instead of 1A

Achocholacek:
https://github.com/keyteki/keyteki/issues/154


jeremylarner:
Does this work when your opponent has 1 or 2 amber and you discard 2 logos cards?

Achocholacek:
It does, though the interaction is slightly confusing. You still have to select more than one card but it captures 1 or 2A on the first card and nothing on the second (which is how it should behave) I just don't know how to better illustrate it to players.

Achocholacek:
fixes #183

Issue #152, opened 2019-07-28
Forgemaster Og doesn't work as expected


viljums:
Forgemaster Og makes owner lose all of the AEmber if opponent forges a key, instead of opponent losing it, while card text reads "After a player forges a key, they lose all of their remaining AEmber".

Achocholacek:
PR submitted - should be resolved on next code push.

PR #159, opened 2019-07-28
Fix for Forgemaster Og

Achocholacek:
https://github.com/keyteki/keyteki/issues/152


jeremylarner:
You can do this in a single ability by referencing `context.event.player` and `context.event.player.amber`

Issue #141, opened 2019-07-28
Might Makes Right


twilliamson15:
When MMR is played there are two issues:
1) There is no way to "not select" creatures to sacrifice. Card says "may" so I shouldn't have to select.
2) Card appears to allow selection of opponents creatures for sacrifice.

Achocholacek:
Re: 1 - if you click "Cancel prompt" after playing Might Makes Right, that is the equivalent of "not selecting" creatures. I will bring your feedback to the team, though. perhaps there is a way we can make this more clear.

Re: 2 - This has been resolved in a recent PR.

twilliamson15:
Cancel Prompt only shows after entering manual mode.

Achocholacek:
You're 100% correct. My apologies. I have added in the ability to complete the action without selecting anything regardless of game mode.

PR #160, opened 2019-07-28
Added 'self' qualifier on targeting.

Achocholacek:
https://github.com/keyteki/keyteki/issues/141


jeremylarner:
This needs `optional: true` and `controller 'self'`

Achocholacek:
Sorry - must have got my wires crossed when I did the original PRs - didn't realize I took out one when I added the other.

PR #203, opened 2019-07-30
Exterminate

saluk:
Customize prompt and messages for each creature selected and destroyed with exterminate! exterminate!


saluk:
Exterminate will need more work anyway, closing for now.

PR #206, opened 2019-07-30
Changes to Exhaust action

### LINKED ISSUE #188: Containment Field
Containment Field played on opponents Yantzee Gang, Gang action used to Steal 1, not destroyed.

### LINKED ISSUE #198: Boneverot Venom
My opponent used a creature with already 3 damage on it and Boneverot Venom to fight an elusive creature. SO the creature should get another 2 damage but didnt happen.

Achocholacek:
Exhaust Action now returns "onCardExhausted" event by default (aka on use), but can be passed {used: false} to return "onCardExhaustedWithoutUse" event.

This allows efficiency in cards like Containment Field and Bonerot Venom, because we can simply look for the "onCardExhausted" event and know that it has been used.

Cards that manually trigger the exhaust event have been updated to be passed {used: false}.

Test for Bonerot Venom added.
Fixes #198 
Fixes #188

### DIFF
--- a/server/game/GameActions/ExhaustAction.js
+++ b/server/game/GameActions/ExhaustAction.js
@@ -1,6 +1,10 @@
 const CardGameAction = require('./CardGameAction');
 
 class ExhaustAction extends CardGameAction {
+    setDefaultProperties() {
+        this.used = true;
+    }
+
     setup() {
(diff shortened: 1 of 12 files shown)


cryogen:
Superseded

PR #207, opened 2019-07-31
Added chain shedding to Punctuated Equilibrium + test.

### LINKED ISSUE #195: Punctuated Equilibrium
The one who plays Punctuated Equilibrium only refills up to 5 cards, whereas it should be 6 by default. Maybe Punctuated Equilibrium is still counted together with the 5 new cards, making it a false "6-card hand".

Achocholacek:
Fixes #195


cryogen:
This should probably a new type of action/effect that drawing at the end of a turn and this particular  card share, rather than explicitly shedding chains.  That way anything which might interact with it in future will work properly

cryogen:
#394 replaces this

Issue #223, opened 2019-07-31
Foozie reap ability doesn't work if I destroyed a creature in the turn


Pekkhyu:
Foozie reap ability doesn't work if I destroyed a creature in the turn. I only gain 1 aember instead of 2.

Achocholacek:
I cannot reproduce this locally.

Do you remember anything else about board state and the situation?

Additionally, are you sure you didn't receive 2 aember?

The game log doesn't specifically say you gain an aember from reaping, so it reads like this:
MrChocholate uses Foozle to reap with Foozle
MrChocholate uses Foozle to gain 1 Æmber 

but in actuality, you have gained 2. 1 from the reap action and one from his ability.

Thanks for reporting and I look forward to your feedback.

Pekkhyu:
I start my turn with 0 aember. I used First blood with 2 Foozle on the ground. I gain 1 aember from first blood. I target a creature and that creature died. Then I reap with my 2 Foozle on the ground but I only gain the aembers for the reap and not for the effect. I finished the turn with 3 aember and not 5.

maleckai:
@Achocholacek I also experienced a game yesterday where Foozle didn't provide the bonus Aember. Though I thought it related perhaps to Foozle not being played until after the destruction occurred.

My scenario yesterday was:
1. Opponent destroys my last creature with a creature already on the field.
2. Plays Foozle
3. Uses Gauntlet of Command to ready Foozle
4. Reaps and gains only 1
5. Uses Ganger Chieftan to ready Foozle again
6. Reaps and gains only 1

Definitely seems something isn't working quite right with detecting that destruction has occurred.

maleckai:
With my currently limited understanding of the code, I suspect Foozle's destroy tracking doesn't start until it's played to the field. Which would explain my case why it's not working. But doesn't explain why it wouldn't work for @Pekkhyu

Achocholacek:
Yeah, there's definitely an issue. I've been digging into it but I'm going crosseyed and It's time for bed. Will tackle it tomorrow.

cryogen:
I think this was fixed with #226.  Please reopen if not the case.

PR #226, opened 2019-07-31
Foozle Test

Achocholacek:


Achocholacek:
Fixes #223

Achocholacek:
this still doesn't fix the situation where a creature is destroyed before Foozle is in play. will have to work on that.

Achocholacek:
You're 100% right - I haven't had a chance to sit down and try figuring out how to do that, but I figured this at least helps a bit more in the interim.

PR #232, opened 2019-08-01
Fix for Might Makes Right/Test

### LINKED ISSUE #171: might makes right & Archimedes
Good Morning, 
When i play Might makes right to sacrify creatures near Archimedes that creatures are supposed to go in my archive and not in discard pile. i have to report this bug. 

Thanks

Achocholacek:
No longer forges if you select less than 25 power worth of creatures.
Updated test with proper functions for archiving process.
Fixes #171 
Fixes #173

### DIFF
--- a/server/game/cards/02-AoA/MightMakesRight.js
+++ b/server/game/cards/02-AoA/MightMakesRight.js
@@ -3,6 +3,8 @@ const Card = require('../../Card.js');
 class MightMakesRight extends Card {
     setupCardAbilities(ability) {
         this.play({
+            condition: context =>
+                context.player.creaturesInPlay.reduce((total, c) => total + c.power, 0) >= 25,
             target: {
                 optional: true,
                 mode: 'minStat',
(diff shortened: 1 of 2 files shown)


server/game/cards/02-AoA/MightMakesRight.js
```diff
@@ -13,6 +15,7 @@ class MightMakesRight extends Card {
                 gameAction: ability.actions.sacrifice()
             },
             then: {
+                condition: context => context.preThenEvents && context.preThenEvents.filter(event => !event.cancelled).reduce((total, event) => total + event.card.power, 0) > 25 ? 1 : 0,
```
cryogen: I think the right way to do this is "may: 'Forge a key'", see chota for an example

Achocholacek: As per our discussion - MMR does not allow for optional forging. If you choose to sacrifice, it MUST be creatures with a total Power of >= 25, and if you sacrifice, you MUST forge.

cryogen:
Still bothers me we're reaching into the event black box here, but let's fix it later

PR #233, opened 2019-08-01
First Blood - allow damage dealt however player prefers + test

### LINKED ISSUE #219: First Blood doesn't let you freely divide its damage among any number of creatures.
Currently, the way damage is distributed after playing First Blood is the same as after playing Mothership Support. However, the two cards differ subtly in their wording.

First Blood says: "Play: Deal 2D for each friendly Brobnar creature. You may divide this damage among any number of creatures."

Whereas Mothership Support says: Play: For each friendly ready Mars creature, deal 2 to a creature. (You may choose a different creature each time.)

The difference is that with First Blood, you should be able to distribute the damage however you want, whereas for Mothership Support, you have to deal the damage in multiples of 2

Here is an example. Let's say my opponent has 2 3-power creatures in play. I have 3 friendly Brobnar creatures at the start of my turn and play First Blood. That should let me deal 6 damage distributed however I want. I should be able to deal 3 to one of my opponent's creatures and 3 to the other one. On the other hand if my opponent's creatures are the same but I instead have 3 friendly ready Mars creatures and play Mothership Support, I have to deal the damage in multiples of 2, meaning I can't deal 3 to each of my opponent's creatures. Instead, I'd have to deal 4 to one and 2 to the other.

Currently, First Blood is working like Mothership Support, but it should be changed to work the way I described in the example above.

Achocholacek:
Fixes #219

### DIFF
--- a/server/game/cards/01-Core/MothershipSupport.js
+++ b/server/game/cards/01-Core/MothershipSupport.js
@@ -7,6 +7,7 @@ class MothershipSupport extends Card {
             gameAction: ability.actions.sequentialForEach(context => ({
                 num: context.player.cardsInPlay.filter(card => card.type === 'creature' && card.hasHouse('mars') && !card.exhausted).length,
                 action: ability.actions.dealDamage({
+                    noGameStateCheck: true,
                     amount: 2,
                     promptForSelect: {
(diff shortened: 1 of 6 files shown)


Achocholacek:
Fixes #236

Issue #230, opened 2019-08-01
Project ZYX


MythicMikeneto:
Project ZYX wont allow me to choose my opponents card that are in my archive. I'm sure I cant keep them on the field but I would think it would allow me to return them to their hand/discard them if appropriate.

Project ZYX playing a card from the archive when reaping is not optional and it should be.

PR #234, opened 2019-08-01
Make ZYX Optional

Achocholacek:
Partially deals with #230 - still need to address making opponents cards selectable in your archives.


cryogen:
@Achocholacek You need to split this change out (or fix the conflict)

Achocholacek:
This PR is hot trash I should be fired.

PR #254, opened 2019-08-02
Made Project Z.Y.X's reap ability optional.

### LINKED ISSUE #230: Project ZYX

Achocholacek:
Fixes #230


Achocholacek:
Fixes #230

Issue #271, opened 2019-08-04
Neffru granting Aember when it is destroyed.


entropyguardian:
Neffru's constant passive ability is no longer is play when it is destroyed, so it does not grant Aember when it itself is destroyed. 

If it is destroyed simultaneously with a group of creatures, it does no grant Aember for anything. 
![Capture](https://user-images.githubusercontent.com/53681800/62419661-4cdf0f00-b64b-11e9-984e-5309189a0316.JPG)

Same for Krump dying and killing its opponent in a battle, or Tolas dying.

romanogit:
@Achocholacek I think this one is fixed right? Can we close it?

PR #281, opened 2019-08-05
Fixed Neffru awarding opponent an aember if he was attacked and both died. 

Achocholacek:
Added Neffru Test and Krump Test.


Achocholacek:
Fixes #271

PR #295, opened 2019-08-07
Change IG from "steal" to "gain/lose" so it behaves properly when things that affect steal are in play

### LINKED ISSUE #288: Po´s Pixies vs Indermensional Graft
I just played a game where the opponent did not lose his remaining Æmber after Forging.

Po´s+ Pixies protect against Steal and Capture, but the effect from Interdimensional Graft is neither and ths the player should still se his remaining Æmber when it is being granted to me

Achocholacek:
Done as two separate game actions because arraying them in one lastingEffect was throwing errors.

Fixes #288


Achocholacek:
Added new action. Still fixes #288

PR #309, opened 2019-08-09
fix for archived opp creature behaviour

### LINKED ISSUE #308: Tandalin and Sample Collection
I use Sample Collection and put an opposing creature into my archive.  My opponent attacks and discards the creature with Tandalin.  It should go to my opponent's hand - but it goes to their discard pile on here.  Should go to their hand per Sample Collection.

Achocholacek:
If you have an opponents creature in your archive, and it leaves, it will now return to their hand in all cases (including if removed by dysania or tantadlin).
modified destructive analysis to only deal damage based on actual purged creatures (aka your own) rather than *all* creatures, as if it is an opponent's, it doesn't get purged and no damage is dealt.
added test for destructive analysis.
Modified test for sample collection for Tantadlin edge case.

Fixes #308

### DIFF
--- a/test/server/cards/01-Core/SampleCollection.spec.js
+++ b/test/server/cards/01-Core/SampleCollection.spec.js
@@ -9,7 +9,7 @@ describe('Sample Collection', function() {
                         hand: ['sample-collection', 'phase-shift', 'labwork', 'wild-wormhole']
                     },
                     player2: {
-                        inPlay: ['troll', 'bumpsy', 'hebe-the-huge']
+                        inPlay: ['troll', 'bumpsy', 'hebe-the-huge', 'tantadlin']
                     }
                 });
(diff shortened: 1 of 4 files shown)


test/server/cards/01-Core/SampleCollection.spec.js
```diff
@@ -9,7 +9,7 @@ describe('Sample Collection', function() {
                         hand: ['sample-collection', 'phase-shift', 'labwork', 'wild-wormhole']
                     },
                     player2: {
-                        inPlay: ['troll', 'bumpsy', 'hebe-the-huge']
+                        inPlay: ['troll', 'bumpsy', 'hebe-the-huge', 'tantadlin']
```
cryogen: Totally always called this card "tantadin" guess it pays to read..

PR #369, opened 2019-08-22
Add "PreKey" phase to allow for start of turn effects if key phase is suppressed.

### LINKED ISSUE #356: Director Z.Y.X Skipping Archive
Had Director Z.Y.X out for several turns and he archived.  My opponent played Miasma and then Director did not archive.  This actually work that way?  I wouldn't think so.  It appears to me it's not working after Miama is played.

DIRECTOR

Achocholacek:
I'm not sure if this is the best approach to the situation, but it seems not impossible that there will be other effects that we need to worry about that could be adversely affected if the key phase is skipped.

Fixes #356


cryogen:
I think the right thing to do here is to fix all that reference the 'key' phase and make them use the onBeginRound event instead

Issue #350, opened 2019-08-19
Newly added decks don’t go to the top of the list


cryogen:
decks are sorted by most recently added, when a new deck is added it goes to the bottom instead of the top

romanogit:
This one seems a quick fix, so I gave it a try.

romanogit:
Sent a pull request for this one:
https://github.com/keyteki/keyteki/pull/370

romanogit:
@cryogen , 
I believe we can close this one, after pull merge

romanogit:
This one can be close (I now learnt how to do it automatically)

PR #370, opened 2019-08-22
Fixes the order of newly imported deck

romanogit:
Fixes the order of newly imported deck and select the new deck (#350)


romanogit:
Not sure why Github is getting too commits

Issue #342, opened 2019-08-17
UI: Found another deck which large count of Logos which messes with deck list


romanogit:
This one is related to #283 

![image](https://user-images.githubusercontent.com/2104075/63216978-76506e00-c114-11e9-8589-20d198976645.png)

PR #381, opened 2019-08-25
DeckSummary.jsx to use table CSS (#342)

romanogit:
DeckSummary.jsx to use table CSS instead of columns and 'page-break: avoid' (#342)
It handles DIV expansion better. Useful for large card names (when i18n is in place)


cryogen:
@romanogit Can you include a screenshot of how this looks now?  And does it still look the same on the different device breakpoints?

cryogen:
Eg Tablet, mobile phone

Issue #382, opened 2019-08-25
Purge pile does not allow drag and drop in manual mode


romanogit:
Purge pile does not allow drag and drop in manual mode

PR #383, opened 2019-08-25
Purged pile is now a Droppable pile (#382)

romanogit:
Implemented it like other droppable piles.


cryogen:
This changes the purge pile to always display, which I'm not sure we want.  Most of the time it just takes up space and won't get used

PR #408, opened 2019-09-04
Fix for Nerve Blast does not inform which creature took damage (#407)

### LINKED ISSUE #407: Nerve Blast does not inform which creature took a damage. [cosmetic]
It just says it stole 1 amber:

image

Dicky-Boy:
Added message to Nerve Blast to show which creature was targeted when dealing damage.

Fixes #407


cryogen:
Thanks!

PR #418, opened 2019-09-06
Fix for Might Makes Right not considering power modifiers

### LINKED ISSUE #264: Might Makes Right not Counting Power Modifiers
I had panpaca anga to the left and 4 creatures totalling 22 power to the right. Might Makes Right ignored the power added by Anga when I sacrificed them.
MMR TCO

### LINKED ISSUE #332: Might Makes Right doesn't work
When i have creatures with 25 or more power in table and i use might makes right it makes me select creatures and makes me sacrifice them but it doesn't make me forge the key

### LINKED ISSUE #414: Might makes right and pancapa anga
It is not counting the +2 power to creatures to the right of pancapa anga when calculating for might makes right

Dicky-Boy:
Added modifiedPower field to card clone/snapshot, to capture actual effective power of card (including modifiers) at time of snapshot, and updated Might Makes Right to consider this value instead of printed power.
Also updated Might Makes Right test to consider this case (modified power of selected cards > 25, printed power < 25)

Fixes #264 
Fixes #332 
Fixes #414

### DIFF
--- a/server/game/Card.js
+++ b/server/game/Card.js
@@ -418,6 +418,7 @@ class Card extends EffectSource {
         clone.location = this.location;
         clone.parent = this.parent;
         clone.clonedNeighbors = this.neighbors;
+        clone.modifiedPower = this.getPower();
         return clone;
     }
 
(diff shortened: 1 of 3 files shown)


cryogen:
Thanks!

Dicky-Boy:
Cheers :D

PR #429, opened 2019-09-10
A single correction + everything else finished

Blazing-Archon:
"Normal" was listed as "Pequeño" (small) so I changed it to "Normal" (the word is the same. Everything else is translated, including sentences with <> and so on (I have made sure I didn't change any of symbols).


romanogit:
@cryogen , 
what is the issue with the PR?
Do you want me to merge on my local branch and submit a new one?
I am about to submit a PR with French locales as well.

cryogen:
there was an error in the json and then a merge conflict.

should be fixed now

PR #430, opened 2019-09-10
Fix for Redlock 'Check' timing issue

### LINKED ISSUE #428: Redlock and "Check" timing issue
If you get exactly enough aember with Redlock's end of turn effect there is no "Check" notification in chat. With a little testing it seems that the "Check" check fires before Redlock does.

Dicky-Boy:
Changed Redlock effect to trigger at the end of main phase rather than end of round, so that the amber is added before check would be declared

Fixes #428

### DIFF
--- a/server/game/cards/02-AoA/Redlock.js
+++ b/server/game/cards/02-AoA/Redlock.js
@@ -4,7 +4,8 @@ class Redlock extends Card {
     setupCardAbilities(ability) {
         this.interrupt({
             when: {
-                onRoundEnded: (event, context) =>
+                onPhaseEnded: (event, context) =>
+                    event.phase === 'main' &&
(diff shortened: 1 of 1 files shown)


cryogen:
I think onRoundEnded is correct, but we need to fix that event to be generated correctly.

Your change will trigger for both players, and we changed it to avoid using these kind of triggers because it causes other issues

romanogit:
@cryogen : please, not that #433 also fixes the same issue.

romanogit:
@cryogen : please, note that #433 also fixes the same issue.

cryogen:
Thanks for the PR, but the other implementation got merged, sorry!

Dicky-Boy:
No problem!

PR #433, opened 2019-09-10
 Attempt to fix the check statement before collecting "after round" ambers by Redlock

### LINKED ISSUE #428: Redlock and "Check" timing issue

romanogit:
(not sure why the Card.js change is here, though)
Attempt to fixes #428


romanogit:
This is an alternative fix to the one presented in https://github.com/keyteki/keyteki/pull/430

PR #440, opened 2019-09-20
Final changes to fix archived not showing controlled creatures

### LINKED ISSUE #307: Destructive Analysis doesn't allow you to purge opponent's cards from your Archive
The 'choose cards' prompt is displayed, but you can't select your opponent's cards which are in your Archives.

Unfortunately I closed my game before taking screenshots and copying the chat log, but I think this is quite a reproducible bug, as I've had it happen in two separate games.

romanogit:
Fixes #307

### DIFF
--- a/less/cards.less
+++ b/less/cards.less
@@ -1,4 +1,4 @@
-@controlledBox: 0 0 1px 2px @brand-warning;
+@controlledBox: 0 0 1px 2px #ffdf14;
 
 .card-wrapper {
   margin: 0;
--- a/less/cards.less
+++ b/less/cards.less
@@ -1,3 +1,5 @@
+@controlledBox: 0 0 1px 2px @brand-warning;
+
 .card-wrapper {
   margin: 0;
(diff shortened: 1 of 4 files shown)


romanogit:
See my latest comment on #307 for Look&Feel:
https://github.com/keyteki/keyteki/issues/307#issuecomment-533533432

less/cards.less
```diff
@@ -1,3 +1,5 @@
+@controlledBox: 0 0 1px 2px #ffdf14;
```
cryogen: Please make this colour a variable

romanogit:
The full box is a variable? Should I move it to variables.less and extract
the color?

Em sex, 20 de set de 2019 10:07 AM, Stuart Walsh <notifications@github.com>
escreveu:

cryogen:
The colour should go in variables.less

ideally it should be based off one of the theme colours but i think this is special enough to be its own colour

PR #443, opened 2019-09-21
Worlds collide

saluk:
First set of easy worlds collide cards, and then irestaff which needed an implementation of enrage.


server/game/Card.js
```diff
@@ -57,6 +59,28 @@ class Card extends EffectSource {
...
+            location: 'any',
+            printedAbility: false,
+            condition: () => {
+                return this.tokens.enrage >= 1;
+            },
+            match: this,
+            effect: AbilityDsl.effects.mustFightIfAble()
+        });
+
+        this.reaction({
```
cryogen: This probably needs to happen in the engine itself, rather than being a reaction

server/game/cards/03-WC/TheGrimReaper.js
```diff
@@ -0,0 +1,19 @@
...
+                onCardEntersPlay: (event, context) => {
+                    return event.card === context.source && context.player.isHaunted();
+                }
+            },
+            gameAction: ability.actions.ready()
+        });
+    }
+}
+
+TheGrimReaper.id = 'the-grim-reaper'; // This is a guess at what the id might be - please check it!!!
```
cryogen: Please remove the comments here, they're not necessary

server/game/Card.js
```diff
@@ -69,13 +93,15 @@ class Card extends EffectSource {
         this.printedPower = cardData.power;
         this.printedArmor = cardData.armor;
         this.armorUsed = 0;
+        this.enraged = false;
```
cryogen: This doesn't look like it's used

server/game/Card.js
```diff
@@ -457,6 +483,20 @@ class Card extends EffectSource {
         return this.hasToken('amber') ? this.tokens.amber : 0;
     }
 
+    enrage() {
+        this.addToken('enrage', 1);
+        if(this.tokens.enrage > 1) {
```
cryogen: Better to check above if it should be enraged and return early if not

server/game/Card.js
```diff
@@ -457,6 +483,20 @@ class Card extends EffectSource {
...
+            this.tokens.enrage = 1
+        }
+    }
+
+    unenrage() {
+        this.addToken('enrage', -1);
+        if(this.tokens.enrage < 0) {
```
cryogen: Same as above, check first

server/game/GameActions/UnenrageAction.js
```diff
@@ -0,0 +1,22 @@
+const CardGameAction = require('./CardGameAction');
+
+class UnenrageAction extends CardGameAction {
```
cryogen: I'm not sure this needs to be a game action, it should probably just be part of the fight action

server/game/cards/02-AoA/LittleRapscal.js
```diff
@@ -5,10 +5,11 @@ class LittleRapscal extends Card {
         this.persistentEffect({
             targetController: 'any',
             match: card => card.type === 'creature' && ability.actions.fight().canAffect(card, this.game.getFrameworkContext(card.controller)),
-            effect: [
+            effect: ability.effects.mustFightIfAble()
+            /* effect: [
```
cryogen: If you're happy this replaces what was there before please remove the commented code

saluk:
@cryogen Tests failing due to trailing comma in your WC.json

PR #449, opened 2019-09-24
Server side i18n commit which includes:

romanogit:
Server side i18n commit which includes:
- Server side prompt text
- Active player prompt
- locale updates

Relates to (#2)


romanogit:
Somehow, there is something broken with this PR
I am investigating

romanogit:
I noticed it is not a problem with i18n. it is actually a navigation issue. I opened a new ticket to track it: https://github.com/keyteki/keyteki/issues/454

Issue #362, opened 2019-08-21
Selwyn the Fence creates an aditional aember when reaping


YuliYudith:
Although none of the active player's creatures has any captured aember (meaning no aember can be moved to the active player's pool), Selwyn generates 2 aember when reaping.

Dicky-Boy:
Above PR introduces test for Selwyn - under the conditions tested the card appears to work as intended. I know it's been a little while since you reported this issue, but can you remember any other details of the board state which might help us figure out what could be causing the bug?

romanogit:
![image](https://user-images.githubusercontent.com/2104075/73977478-38851300-4909-11ea-92fb-dc2048ced747.png)

Achocholacek:
closing due to inactivity.

PR #450, opened 2019-09-24
Create test spec for Selwyn the Fence

Dicky-Boy:
Created test spec for Selwyn the Fence as part of investigation into Issue #362  - as it stands the tests pass and further info needed to understand root cause of issue.


cryogen:
Tests are good!

PR #453, opened 2019-09-25
Decklist Card

SkyJedi:
added decklist in place of the identity card
added nicer looking archon back for sealed games


server/game/player.js
```diff
@@ -560,6 +560,7 @@ class Player extends GameObject {
             cardPiles: {
                 archives: this.getSummaryForCardList(this.archives, activePlayer, true),
                 cardsInPlay: this.getSummaryForCardList(this.cardsInPlay, activePlayer),
+                deck: this.getSummaryForCardList(this.deck, activePlayer),
```
cryogen: This change will expose both decks to both players, as they're used, I think we need to take a copy of the deck when it's first created, and use that, and I think the sorting needs to be maintained, to preserve some visual notion that it's not the actual deck they're seeing

SkyJedi: made a of deckCards data 

opponents deckCards is only generated in a non sealed game

PR #456, opened 2019-09-25
Fixes Maruck bugs

### LINKED ISSUE #192: Red Hot Armor + Maruck the Marked
When active player plays Red Hot Armor and the opponent has Maruck out, he is capturing aember for the armor which is "removed" by RHA. By the way RHA is worded it seems this isn't as intended but cannot find a rules reference online regarding it. 

Red Hot Armor: "Play: Each enemy creature with armor loses all of its armor until the end of the turn and is dealt 1D for each point of armor it lost this way."

Maruck the Marked: "After Maruck the Marked prevents damage with its armor, capture 1A for each damage just prevented."

### LINKED ISSUE #412: UI: Maruck and Shield of Justice
Maruck is being protected by Shield of Justice, but its armor is being reduced (even though it is not taking the damage). The armor should not be reduced.

Screenshot of Maruck with Protect the Weak, fighting Shooler after Shield of Justice is played:

image

romanogit:
Fixes #192 Redhot armor should not allow Maruck to capture amber
Fixes #412 Dealing damage when it cannot affect a target should not reduce armor

### DIFF
--- a/server/game/GameActions/DealDamageAction.js
+++ b/server/game/GameActions/DealDamageAction.js
@@ -3,8 +3,8 @@ const CardGameAction = require('./CardGameAction');
 class DealDamageAction extends CardGameAction {
     setDefaultProperties() {
         this.amount = null;
-        this.amountForCard = (card, context) => 1; // eslint-disable-line no-unused-vars
-        this.reduceArmorBy = (card, context) => 0; // eslint-disable-line no-unused-vars
+        this.amountForCard = () => 1;
+        this.reduceArmorBy = () => 0;
         this.fightEvent = null;
         this.damageSource = null;
         this.splash = 0;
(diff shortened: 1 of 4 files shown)


server/game/GameActions/DealDamageAction.js
```diff
@@ -4,6 +4,7 @@ class DealDamageAction extends CardGameAction {
     setDefaultProperties() {
         this.amount = null;
         this.amountForCard = (card, context) => 1; // eslint-disable-line no-unused-vars
+        this.reduceArmorBy = (card, context) => 0; // eslint-disable-line no-unused-vars
```
cryogen: the eslint disabling is not needed here

romanogit: When I compiled, I got the context not used error.

cryogen: Ah yes, just remove the variables from both, they're not needed

romanogit: ok, I will make the change

romanogit: I get the same error on card:
C:\workspace_svn\public\keyteki\br-bugs\server\game\GameActions\DealDamageAction.js
  6:31  error  'card' is defined but never used  no-unused-vars
  7:31  error  'card' is defined but never used  no-unused-vars

PR #469, opened 2019-09-27
Using this.props inside Promise

romanogit:
Might be safer to access this inside Promise


romanogit:
@cryogen : this one was implemented because I saw that all props were accessed inside the deck rendering Promise call. I thought that reading props outside could cause some inconsistency by the time the Promise execution was called.

PR #475, opened 2019-09-28
Fix ActivePlayerPrompt.jsx when .text is null

romanogit:


client/Components/GameBoard/ActivePlayerPrompt.jsx
```diff
@@ -131,8 +131,8 @@ class ActivePlayerPrompt extends React.Component {
         }
 
         let timer = null;
-
-        let promptText = this.props.promptText ? (this.props.promptText.text || this.props.promptText) : null;
```
cryogen: this could be:

let promptText = this.props.promptText && this.props.promptText.text ? this.props.promptText.text : this.props.promptText;

romanogit: Make the code more explicit now

PR #476, opened 2019-09-28
Decklist fix

SkyJedi:


server/game/player.js
```diff
@@ -611,9 +623,9 @@ class Player extends GameObject {
                 return 0;
             });
             state.cardPiles.deck = this.getSummaryForCardList(sortedDeck, activePlayer, true);
-            state.deckCards = this.getSummaryForCardList(this.deck, activePlayer);
+            state.deckCards = this.getDeckCards(this.deckData.cards);
```
cryogen: Doing this every state change seems excessive.  We should do this once when the deck is initialised and then always use that array

SkyJedi: Yes agreed, off you figure that out! Still learning this beast

PR #478, opened 2019-09-29
New deck lists

SkyJedi:


client/Components/GameBoard/IdentityCard.jsx
```diff
@@ -7,8 +7,8 @@ import QRCode from 'qrcode';
 import { withTranslation } from 'react-i18next';
 
 class IdentityCard extends React.Component {
-    constructor(props) {
```
cryogen: It is 'proper' to pass the props to the base class, so not sure why you're removing it :)

client/Components/GameBoard/IdentityCard.jsx
```diff
@@ -159,15 +160,29 @@ class IdentityCard extends React.Component {
...
+                ctx.fillText(this.props.t(house).toUpperCase(), houseNames[index].x, houseNames[index].y);
+                ctx.strokeText(this.props.t(house).toUpperCase(), houseNames[index].x, houseNames[index].y);
             });
             this.setState({imageUrl: canvas.toDataURL()});
-
-            this.refs.dummy.parentElement.removeChild(this.refs.dummy);
         });
     }
 
+    findColor(number) {
```
cryogen: ...? :D

client/Components/GameBoard/IdentityCard.jsx
SkyJedi: Googles "props base class"
 realization Dawns
I watch the world burn.

client/Components/GameBoard/IdentityCard.jsx
SkyJedi: How else will the cards truly discover who they are?

PR #486, opened 2019-09-30
Some creature's can be played as upgrades

saluk:


server/game/Card.js
```diff
@@ -516,12 +528,16 @@ class Card extends EffectSource {
         });
     }
 
+    canPlayAsUpgrade() {
+        return this.effects.filter(effect => effect.type === 'canPlayAsUpgrade');
```
jeremylarner: this should be `this.anyEffect('canPlayAsUpgrade')`

server/game/Card.js
```diff
@@ -101,6 +102,17 @@ class Card extends EffectSource {
...
     }
 
+    get type() {
+        if(this.effectiveType !== null) {
+            return this.effectiveType;
+        }
+        return this.printedType;
+    }
+
+    set type(t) {
```
jeremylarner: type should never be set

server/game/GameActions/AttachAction.js
```diff
@@ -48,6 +48,7 @@ class AttachAction extends CardGameAction {
             event.parent.upgrades.push(event.card);
             event.card.parent = event.parent;
             event.card.controller = context.player;
+            event.card.effectiveType = 'upgrade';
```
jeremylarner: this is problematic, because once you turn the card into an upgrade, it never changes back again (so if you go through your deck, the game won't let you play it as a creature).  Ideally, it's state as an upgrade is tracked by an effect (which will be removed when it leaves play). You could hack it some other way too though

server/game/Card.js
saluk: A card is a gameobject and all gameobjects set their type to '' in the constructor. As of this PR the only objects where it makes sense to mask the type is cards, but it won't compile if I don't let the constructor set the type. Should I move the masking into GameObject anyway - or is there a way I can get around the GameObject constructor trying to set the type?

server/game/Card.js
saluk: Hoo boy this was somehow masking some bugs. Fixing...

server/game/GameActions/AttachAction.js
saluk: Nice catch! I added a test case for this scenario and changed my logic (hacked it another way) in the next push.

server/game/Card.js
```diff
@@ -516,12 +527,16 @@ class Card extends EffectSource {
...
+        return this.anyEffect('canPlayAsUpgrade') || this.type === 'upgrade';
+    }
+
     /**
      * Checks whether the passed card meets the upgrade restrictions (e.g.
      * Opponent cards only, specific factions, etc) for this card.
      */
     canAttach(card, context) { // eslint-disable-line no-unused-vars
-        return card && card.getType() === 'creature' && this.getType() === 'upgrade';
+        return card && card.getType() === 'creature';
```
jeremylarner: I don't like that you're removing the safety check here.  Non-upgrades shouldn't be able to be attached to creatures under normal circumstances....

jeremylarner: Maybe you can check for `this.anyEffect('canPlayAsUpgrade')` and `context.ability.isCardPlayed()`?

server/game/Card.js
cryogen: I dunno, this still feels like a smell to me - let me look at this in more detail to see if there's a better way of doing this

server/game/Card.js
saluk: Oops, I think I mistakenly removed it while trying to clean up the rest of the file.

server/game/Card.js
saluk: +1

cryogen: Ok having looked at this further, this.type is only set to '' to keep the "typescript" autocomplete happy, so it seems perverse to introduce a setter just for that.  So let's remove the initialisation and the setter :)

saluk: Done!

PR #488, opened 2019-10-01
Preloaded archons

SkyJedi:


client/Components/Decks/DeckRow.jsx
```diff
@@ -1,10 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment';
-
```
cryogen: I like to have a separation of 'outside' dependencies, vs 'inside' dependencies which is why this blanks space exists :)

client/Components/Decks/DeckSummary.jsx
```diff
@@ -132,37 +132,48 @@ class DeckSummary extends React.Component {
...
                         { this.props.deck.usageLevel > 0 && !this.props.deck.verified ?
-                            <div className='info-row row'><Trans i18nKey='decksummary.toverify'>This deck has been flagged as requiring verification. Please email a photo of the decklist with your username written on a piece of paper to</Trans>: thecrucible.deckcheck@gmail.com
+                            <div className='info-row row'><Trans i18nKey='decksummary.toverify'>This deck has been
+                                flagged as requiring verification. Please email a photo of the decklist with your
+                                username written on a piece of paper to</Trans>: thecrucible.deckcheck@gmail.com
                             </div> : null
                         }
                     </div>
-                    <div className='col-xs-2 col-sm-3 no-x-padding'>{ this.props.deck.agenda && this.props.deck.agenda.code ? <img className='img-responsive' src={ '/img/cards/' + this.props.deck.agenda.code + '.png' } /> : null }</div>
+                    <div
```
cryogen: a better way to break this would be to keep the div attributes on one line and then newline before the {

client/Components/Decks/DeckRow.jsx
SkyJedi: fixed in https://github.com/keyteki/keyteki/pull/488/commits/856c00d0295cf9012fe84bafd8f3ccd160a120df

PR #489, opened 2019-10-01
Portuguese Review

alfoltran:
Just some revisions to improve the Portuguese translation.

I also removed some duplicate keys that existed in the file. By the way, I noticed these same duplications in the other files. Can I remove them?

Finally, the patreon button on the profile is not internationalized. Can I fix this?


cryogen:
Just a very quick instant feedback: 1. Yes please feel free to remove duplicates. 2. Please don’t replace “amber” as the ui automatically converts it to the correct spelling plus an amber icon. Thanks!

alfoltran:
Now I will remove the duplicate keys.

romanogit:
Legal @alfoltran . Obrigado. thanks!

PR #492, opened 2019-10-01
Murkens

### LINKED ISSUE #216: Murkens archive bug
doesn't play random card from archive when the opponents draw pile is at 0

alfoltran:
Fixes #216

Improving testing to detect potential problems. Starting with the most basic tests that are already working.

Next step is to create test scenarios that fail according to reported issues.

### Goals
- [ ] Alpha cards (`log`)
- [x] Cards with **Play** (Persistence Hunting)
- [ ] Archives with multiple cards
- [ ] Archives with cards of multiple owners

### DIFF
--- a/server/game/cards/02-AoA/PersistenceHunting.js
+++ b/server/game/cards/02-AoA/PersistenceHunting.js
@@ -6,7 +6,7 @@ class PersistenceHunting extends Card {
             target: {
                 mode: 'house'
             },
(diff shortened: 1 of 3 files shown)


romanogit:
@cryogen : I believe we can close this PR. It is outdated and the Murkins x alpha is fixed already. 
The pending issue with Murkings is the random + actual draw, which runs random twice, therefore generating an incorrect game message.

PR #494, opened 2019-10-01
Fix RemoveStunAction class name

romanogit:
Found while developing the WC ward feature


cryogen:
It's almost like I copy and paste thing :D

PR #497, opened 2019-10-02
Worlds collide brobnar set1

saluk:


test/server/cards/03-WC/BerserkerSlam.spec.js
```diff
@@ -34,6 +34,7 @@ describe('Berserker Slam', function() {
                 expect(this.forgemasterOg.location).toBe('discard');
                 expect(this.player1.amber).toBe(3);
                 expect(this.player2.amber).toBe(4);
+                expect(this).toHaveRecentChatMessage('player1 uses Berserker Slam to cause player1 to lose 1 aember');
```
cryogen: I’d generally discourage asserting chat messages as the test will break if we ever change the way the log works which will be unrelated to how the card works. I’ll merge this one but bear it in mind

PR #511, opened 2019-10-02
Added basic Ward flags to Card

romanogit:
Added basic Ward flags to Card and enough to allow cards to use Ward actions
Merge after Jeremy's latest PRs should be minimal.


less/cards.less
```diff
@@ -429,6 +429,26 @@
   color: transparent;
 }
 
+.enrage {
+  background-image: url("/img/enrage.png");
+  line-height: 21px;
+}
+
+.enrage.single {
```
cryogen: This is duplicated so should go in a single “.single” class. But, more than that, this whole hiding should be done in the jsx not the css

server/game/cards/03-WC/ForceField.js
```diff
@@ -0,0 +1,15 @@
...
+    setupCardAbilities(ability) {
+        this.whileAttached({
+            effect: ability.effects.gainAbility('reap', {
+                gameAction: ability.actions.ward()
+            })
+        });
+    }
+}
+
+ForceField.id = 'force-field'; // This is a guess at what the id might be - please check it!!!
```
cryogen: Please remove the comment

server/game/cards/03-WC/TricerianLegionary.js
```diff
@@ -0,0 +1,17 @@
...
+        this.play({
+            target: {
+                cardType: 'creature',
+                controller: 'self',
+                gameAction: ability.actions.ward()
+            }
+        });
+    }
+}
+
```
cryogen: Please remove the comment

less/cards.less
romanogit: what do you mean in JSX?
JSX should hold the logic of not showing the number?

romanogit: Fixed

PR #517, opened 2019-10-04
Fixes GameObject.js type

romanogit:


server/game/GameObject.js
```diff
@@ -7,6 +7,7 @@ class GameObject {
         this.game = game;
         this.name = name;
         this.id = this.name;
+        this.objectType = '';
```
cryogen: This is incorrect. This is restoring the behaviour that we removed

server/game/gamesteps/handlermenuprompt.js
```diff
@@ -86,7 +86,7 @@ class HandlerMenuPrompt extends UiPrompt {
                 targets: this.properties.controls.targets.map(target => target.getShortSummary())
             }];
         }
-        if(this.context.source.type === '') {
+        if(!this.context.source.type) {
```
cryogen: This is all that’s needed for the fix

server/game/GameObject.js
romanogit: I really think that having the method getType() returning this.type without having defined the type parameter is wrong.

It is not removing the old behavior because Card#get type is overriden

server/game/gamesteps/handlermenuprompt.js
romanogit: this is needed, regardless, I agree.

server/game/GameObject.js
romanogit: Rolled back this change

PR #527, opened 2019-10-06
Added WC Logos cards/tests - updated test for WC Berserker Slam

Achocholacek:
reflect missing aember pip in original json. 

This is gonna fail SO many tests without the updated WC.json the PR has been submitted for.

### DIFF
--- a/server/game/cards/03-WC/EDAIEdie4x4.js
+++ b/server/game/cards/03-WC/EDAIEdie4x4.js
@@ -1,0 +1,21 @@
+const Card = require('../../Card.js');
+
+class EDAIEdie4x4 extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
+                location: 'hand',
+                controller: 'self',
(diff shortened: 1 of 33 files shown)


server/game/cards/03-WC/EDAIEdie4x4.js
```diff
@@ -0,0 +1,21 @@
...
+        this.play({
+            target: {
+                location: 'hand',
+                controller: 'self',
+                gameAction: ability.actions.archive()
+            }
+        });
+        this.persistentEffect({
+            targetController: 'opponent',
+            effect: ability.effects.modifyKeyCost(() => this.controller.archives.length || 0)
```
cryogen: I'm not sure what you were trying to do with the || 0, but this is redundant

Achocholacek: I copied that from another card, hell if I can remember which one.

PR #533, opened 2019-10-07
added GSA Cards/Tests

Achocholacek:


server/game/cards/03-WC/AEmbertracker.js
```diff
@@ -0,0 +1,19 @@
+const Card = require('../../Card.js');
+
+class AEmbertracker extends Card {
```
cryogen: We usually skip the E in cards that have "amber" in their name

Achocholacek:
this is gonna require the updated wc.json to be merged before it'll work.

server/game/cards/03-WC/AEmbertracker.js
Achocholacek: resolved. will require WC.json update as welll.

cryogen:
the constant is added now

romanogit:
Need a merge of this one :)

cryogen:
It can't be merged until the issues are resolved with it

PR #539, opened 2019-10-08
Implemented ExaltAction

romanogit:
Added two saurian cards to test


server/game/GameActions/RemoveTokenAction.js
```diff
@@ -17,20 +18,20 @@ class RemoveTokenAction extends CardGameAction {
         if(this.amount === 1) {
             token = '+1 power counter';
         }
-        this.effectMsg = 'remove ' + this.amount + ' ' + (this.type === 'power' ? token : this.type) + ' on {0}';
+        this.effectMsg = 'remove ' + (this.all ? 'all' : this.amount) + ' ' + (this.type === 'power' ? token : this.type) + ' from {0}';
```
cryogen: This should be refactored to use `` and the english doesn't make sense

server/game/GameActions/RemoveTokenAction.js
```diff
@@ -17,20 +18,20 @@ class RemoveTokenAction extends CardGameAction {
...
+    }
+
+    computeAmount(card) {
```
cryogen: getAmount is more consistent naming with other things of this nature

server/game/cards/03-WC/CenturionStenopius.js
```diff
@@ -0,0 +1,20 @@
+const Card = require('../../Card.js');
+
+class CenturionStenopius extends Card {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: this,
+            effect: ability.effects.modifyPower(() => 3 * (this.tokens['amber'] || 0))
```
cryogen: tokens.amber

server/game/cards/03-WC/ChantOfHubris.js
```diff
@@ -0,0 +1,29 @@
+const Card = require('../../Card.js');
+
+class ChantOfHubris extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            condition: context => context.game.creaturesInPlay.length > 1 && context.game.creaturesInPlay.some(card => card.hasToken('amber')),
```
cryogen: @jeremylarner can correct me if I'm wrong but I think this is redundant as it's duplicated in the target: below

cryogen:
Also conflicts to resolve

server/game/GameActions/RemoveTokenAction.js
romanogit: Fixed.

server/game/cards/03-WC/ChantOfHubris.js
romanogit: @cryogen: Not really. Since there is an action so select a creature with amber and an action to select another creature, we need a broader condition to validate the whole card has condition to be fulfilled.

cryogen:
conflicts

PR #541, opened 2019-10-08
Fix BreakerHill.js ability

romanogit:


romanogit:
Nevermind this one. Committed too many files

PR #545, opened 2019-10-08
added WC/Untamed cards - added growth counter related action wrappers.

Achocholacek:


server/game/cards/03-WC/DeepwoodDruid.js
```diff
@@ -0,0 +1,21 @@
+const Card = require('../../Card.js');
+
+class DeepwoodDruid extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            reap: true,
+            target: {
+                cardType: 'creature',
+                optional: false,
```
cryogen: Optional: false is the default so this is redunant

server/game/cards/03-WC/SongOfTheWild.js
```diff
@@ -0,0 +1,19 @@
+const Card = require('../../Card.js');
+
+class SongOfTheWild extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            effect: 'give each friendly creature \'Reap: Gain 1A\' for the remainder of the turn',
```
cryogen: A - >amber

Achocholacek: fixed in latest pr.

server/game/cards/03-WC/DeepwoodDruid.js
Achocholacek: fixed in latest pr.

Achocholacek:
should now be in line with expectations

PR #548, opened 2019-10-08
Added Brobnar Cards

Achocholacek:


Achocholacek:
FYI this is failing as the warriors refrain id needs to be changed and a separate PR for that was submitted on the keyteki-json-data repo.

PR #554, opened 2019-10-09
Proposal for Ward x Damage removal (no major refactoring)

romanogit:


romanogit:
I will divide this PR.

PR #559, opened 2019-10-10
Fix archimedes not archiving to owner

### LINKED ISSUE #552: Archimedes vs Collar of Subordination
Set up: Using Collar of Subordination on an enemy creature and then placing Archimedes next to it.

Archimedes gives the stolen creature "Destroyed: Archive this card" but in the rules it says, if a card leaves play, it goes to the rightful owner. So before the card can enter the opponents archive, it should go to the discard pile of its rightful owner.

Right now, this is not the case. Maybe I'm interpreting the rules wrong but as I learned, it was done like this in Call of the Archons.

romanogit:
Fixes #552


server/game/cards/01-Core/MobiusScroll.js
```diff
@@ -9,14 +9,13 @@ class MobiusScroll extends Card {
                 mode: 'upTo',
                 numCards: 2,
                 gameAction: ability.actions.archive(context => ({
-                    target: [context.source].concat(context.target),
-                    owner: true
```
jeremylarner: this is the default, so I don't think you need to make this change

romanogit: I removed `owner: true` because it is the default

Issue #563, opened 2019-10-11
One Stood Against Many issue


PeteQuad:
I used One Stood Against Many on my Staunch Knight.  Opponent had a single creature (Groke).  Groke was killed.  Staunch Knight was not ready after kill.  "Do as much as you can" should have readied him for the next fight even if there was none.

SkyJedi:
#564

cryogen:
Consolidating OSAM issues in #241

PR #564, opened 2019-10-11
Osam fix

SkyJedi:
fix for https://github.com/keyteki/keyteki/issues/563


cryogen:
There should be a better way to do this..

PR #599, opened 2019-10-22
Star Alliance's blaster cards

romanogit:


server/game/cards/03-WC/BlasterCard.js
```diff
@@ -0,0 +1,47 @@
+const Card = require('../../Card.js');
+
+class BlasterCard extends Card {
+    getAttachedAbility(ability, blasterName, officerName) {
+        let choices = {
+            'Deal 2 damages': () => true
```
cryogen: No 's'

server/game/cards/03-WC/BlasterCard.js
```diff
@@ -0,0 +1,47 @@
...
+        };
+
+        choices[`Move ${blasterName}`] = () => true;
```
cryogen: Why not include this in the above initialiser?

server/game/cards/03-WC/BlasterCard.js
```diff
@@ -0,0 +1,47 @@
...
+            optional: true,
+            targets: {
+                action: {
+                    mode: 'select',
+                    choices: choices
+                },
+                creature: {
+                    dependsOn: 'action',
+                    cardType: 'creature',
+                    cardCondition: (card, context) => (context.selects.action.choice === 'Deal 2 damages') ? true : card.name === officerName && !card.upgrades.includes(this),
```
cryogen: No 's' on damage

server/game/cards/03-WC/BlasterCard.js
romanogit: I don't think we can have an expression as a json key. I get an error. That's why I used brackets.

romanogit:
This PR is now complete and merged.

PR #607, opened 2019-10-23
Implement ghosthawk

cryogen:


server/game/cards/03-WC/Malison.js
```diff
@@ -0,0 +1,17 @@
+const Card = require('../../Card.js');
+
+class Malison extends Card {
+    setupCardAbilities(ability) {
+        this.fight({
+            target: context => ({
```
jeremylarner: You cannot pass a function to `target` (and I'm not sure why you are trying to here....)

cryogen: Oops, this wasn't meant to get checked in!

PR #613, opened 2019-10-24
WC anomalies

romanogit:
4 anomalies implemented.
Removed what seems to be some "left-overs" on `MoveCardAction`


server/game/GameActions/MoveCardAction.js
```diff
@@ -22,12 +22,10 @@ class MoveCardAction extends CardGameAction {
...
     getEvent(card, context) {
-        return super.createEvent('onMoveCard', { card: card, context: context }, () => {
-            context.player.moveCard(card, this.destination);
-            if(this.shuffle && (this.target.findIndex(c => c === card) === this.target.length - 1)) {
-                if(this.destination === 'deck') {
-                    context.player.shuffleDeck();
-                }
+        return super.createEvent('onMoveCard', { card: card, context: context }, event => {
+            context.player.moveCard(event.card, this.destination);
+            if(this.shuffle && (this.target.findIndex(c => c === event.card) === this.target.length - 1)) {
```
cryogen: Why've you taken the check of destination?  Won't this always shuffle when any card is moved to anywhere?

romanogit: only if shuffled is set to `true` which should never be the case if you dont want that to happen, right?
This action is not used much.

romanogit: Rollback the MoveCardAction changes

PR #624, opened 2019-10-25
Fix RitualOfBalance condition 

### LINKED ISSUE #617: Cannot Use Ritual of Balance
One should be able to use Ritual of Balance, no matter how much AEmber the opponent has. This is important to trigger cards like Veylan Analyst, or with Poltergeist. This should work the same way that Anomaly Exploiter already works.

romanogit:
Fixes #617


romanogit:
This commit, I believe, is also on #628

PR #634, opened 2019-10-27
Doctor Driscoll

jtrussell:
Would we want to skip the prompt to select a creature to heal when nothing is damaged?

This is how Guardian Demon behaves, but I wasn't sure how to still allow you to use the action in this case. Similar to how you can use Mindwarper's action when your opponent has no aember or no creatures.


cryogen:
I think it's fine as-is.  We seem to be pretty inconsistent with things like that, and the engine should be taking care of whether to prompt or not, not the card itself.

Thanks!

PR #635, opened 2019-10-27
Readme clarifications

avehlies:
I noticed that the JSON file contents in the readme were in the same code block as some bash commands. I split those apart and found other locations where a comment or instruction looked like they were in a block of bash commands.


cryogen:
Thanks!

PR #650, opened 2019-10-29
Fixes Epic Quest

### LINKED ISSUE #584: Cannot use Epic Quest before playing 7 Sanctum cards
Epic Quest has an Omni ability.  You should be able to use+exhaust it even if you haven't played 7 Sanctum cards that turn.  This matters in a deck I have because it has 2 Veylan Analysts that I could use to gain aember by using it without sacrificing it.  Thanks, keep up the good work!

romanogit:
Fixes #584


jeremylarner:
I'm not sure this is the best way to fix it - you're adding an 'If you do,' that isn't on the card

romanogit:
Yes. But how to condition a gameAction?

The global condition prevents the use.

BTW: the key is forged even if artifact is not sacrificed. That was not my expectation. I had to add a condition to the 'then' clause. To me, if I passed an empty target to sacrifice, then should not be called, but I might be misunderstanding

server/game/GameActions.js
```diff
@@ -13,6 +13,7 @@ const Actions = {
     capture: (propertyFactory) => new GameActions.CaptureAction(propertyFactory),
     cardLastingEffect: (propertyFactory) => new GameActions.LastingEffectCardAction(propertyFactory), // duration = 'untilEndOfConflict', effect, targetLocation, condition, until
     clearGrowthTokens: (propertyFactory) => new GameActions.RemoveTokenAction(propertyFactory, 'growth'),
+    conditional: (propertyFactory) => new GameActions.ConditionalAction(propertyFactory),
```
jeremylarner: This shouldn't be with card effects

server/game/GameActions/ConditionalAction.js
```diff
@@ -0,0 +1,42 @@
+const GameAction = require('./GameAction');
+
+class ConditionalAction extends GameAction {
+    setup() {
+        super.setup();
+        this.effectMsg = this.effect || 'conditionally do something';
```
jeremylarner: I don't think you should be using this.effect here

server/game/cards/01-Core/EpicQuest.js
```diff
@@ -10,14 +10,17 @@ class EpicQuest extends Card {
...
-            condition: context => context.game.cardsPlayed.filter(card => card.hasHouse('sanctum')).length > 6,
-            effect: 'sacrifice {0} and forge a key at no cost',
-            gameAction: [
-                ability.actions.sacrifice(),
-                ability.actions.forgeKey(context => ({
-                    modifier: -context.player.getCurrentKeyCost()
-                }))
-            ]
+            gameAction: ability.actions.conditional(({
+                effect: 'sacrifice {1} and forge a key at no cost',
```
jeremylarner: I think this `effect` should be outside the gameAction (as it was before).

test/server/cards/01-Core/EpicQuest.spec.js
```diff
@@ -0,0 +1,50 @@
...
+                this.player1.play(this.protectrix);
+                this.player1.play(this.gormOfOmm);
+                expect(this.player1.amber).toBe(4);
+                this.player1.useAction(this.epicQuest, true);
+                expect(this.epicQuest.location).toBe('play area');
+                expect(this.epicQuest.exhausted).toBe(true);
+                expect(this.player1).toHavePrompt('Choose a card to play, discard or use');
+            });
+
+            it('should not forge a key if played 7', function() {
```
jtrussell: Hey just a heads up - I think you meant `should forge a key` instead of `should not`.

romanogit: I will fix, thanks!

server/game/GameActions.js
romanogit: Fixed.

server/game/cards/01-Core/EpicQuest.js
romanogit: The problem is that `effect` at ability level always prints the message. But, since this is conditional, I would like to have a conditional message as well.
I renamed the variable to `message`. Maybe it makes more sense.

server/game/GameActions/ConditionalAction.js
romanogit: I renamed it to `message`. It is safe to use there as it is set in constructor

server/game/cards/01-Core/EpicQuest.js
jeremylarner: 2 problems with that - it will print a message regardless of whether the condition is true or not.  So you either need two messages, or you need your conditional action to be smarter in how it deals with messaging.  If you're keen to implement this functionality, I'd like to see it included in your tests.

romanogit: @jeremylarner : true/false actions are now objects with an internal message.

romanogit:
@jeremylarner , the engine supports conditional effect messaging for gameActions used in the ability. However, actions like sequential, for example, do not traverse to GameActions to print their messages, therefore, it prints a "do several things" message (check SequentialAction).

I am combining conditional with sequential, which will not present a reasonable message.

Compared to the overall result of the Action, can we consider this to be resolved together with the messaging system?

romanogit:
@cryogen ,
here is the output:

When forging:
```
player1 uses Epic Quest to forge a key if played 6 or more sanctum cards
player1 forges the forgedkeyred, paying 0 amber
```

When not forging:
```
player1 uses Epic Quest to forge a key if played 6 or more sanctum cards
```

strib:
Thanks for fixing this everyone!  The message should read "7 or more sanctum cards" though, right?  (Code looks correct to me.)

PR #656, opened 2019-10-29
WC anomalies: Valoocanth, Infomancer and Memolith

romanogit:


jeremylarner:
Don’t use this and gameAction should be in target

romanogit:
@jeremylarner : fixed your comments above and applied it also to Memolith

server/game/cards/03-WC/Memolith.js
```diff
@@ -0,0 +1,39 @@
...
+                    }
+                },
+                action: {
+                    location: 'any',
+                    controller: 'self',
+                    dependsOn: 'graftOrTrigger',
+                    cardCondition: (card, context) => context.selects.graftOrTrigger &&
+                        (((context.selects.graftOrTrigger.choice === 'Graft an action card') && (card.location === 'hand') && (card.type === 'action'))
+                            || ((context.selects.graftOrTrigger.choice === 'Trigger play effect') && (card.parent === this) && (!card.facedown))),
+                    gameAction: [
```
jeremylarner: this looks like a good place for a conditional action?

server/game/cards/03-WC/Infomancer.js
```diff
@@ -0,0 +1,31 @@
...
+            }
+        });
+
+        this.reap({
+            effect: 'trigger the play effect of {0}',
+            target: {
+                location: 'any',
+                controller: 'self',
+                cardType: 'action',
+                cardCondition: (card, context) => card.parent === context.source && card.facedown === false,
```
cryogen: Why the facedown check?  Is this to target grafted cards?  Surely there's a more explicit way to do that?

server/game/cards/03-WC/Memolith.js
```diff
@@ -0,0 +1,39 @@
...
+
+class Memolith extends Card {
+    setupCardAbilities(ability) {
+        this.action({
+            targets: {
+                graftOrTrigger: {
+                    mode: 'select',
+                    choices: {
+                        'Graft an action card': (context) => context.player.hand.filter(card => card.type === 'action'),
+                        'Trigger play effect': (context) => context.source.childCards.some(card => !card.facedown)
```
cryogen: Same here with the facedown check, I don't like that "faceup === grafted", it needs to be more explicit

server/game/cards/03-WC/Memolith.js
```diff
@@ -0,0 +1,39 @@
...
+                    }
+                },
+                action: {
+                    location: 'any',
+                    controller: 'self',
+                    dependsOn: 'graftOrTrigger',
+                    cardCondition: (card, context) => context.selects.graftOrTrigger &&
+                        (((context.selects.graftOrTrigger.choice === 'Graft an action card') && (card.location === 'hand') && (card.type === 'action'))
+                            || ((context.selects.graftOrTrigger.choice === 'Trigger play effect') && (card.parent === this) && (!card.facedown))),
```
cryogen: and here

server/game/cards/03-WC/Infomancer.js
jeremylarner: I think it's because there's a possibility of using Creed of Nurture to give Jargogle Infomancer's abilities.  We use the same system for both Jargogle's Play/Destroyed effects, and Infomancer's abilities, so we need to avoid the game treating a facedown card as a grafted one

cryogen: I get that, but I think it should be more explicit like some kind of grafted state or flag. That way the next time they had a place under effect we won’t have to rework everything

server/game/cards/03-WC/Memolith.js
romanogit: Fixed it using ConditionalAction

server/game/cards/03-WC/Infomancer.js
romanogit: Using `grafted` location on the card code now

server/game/cards/03-WC/Memolith.js
romanogit: Fixed to use `grafted` location

PR #657, opened 2019-10-30
Implement Saurian 'Favor of Rex'

romanogit:
This is still in draft. There is a difficult scenario and one that I cannot figure out the reason.

1) Capture with opponent's charette is always getting my amber. Should it be capturing from the controller of charette (my opponent)?

2) Not sure what happens when I select a creature without `Play:` It seems to be playing an effect of another card. 

Both tests are commited with `xit`


romanogit:
@cryogen , @jeremylarner : I would appreciate your review. Thanks.

romanogit:
See #664

PR #681, opened 2019-11-03
Implement book of leq

cryogen:


server/game/cards/03-WC/BookOfLeq.js
```diff
@@ -0,0 +1,32 @@
...
+                    trueGameAction: ability.actions.changeActiveHouse(context => ({ house: context.player.deck[0].printedHouse })),
+                    falseGameAction: ability.actions.untilNextTurn({
+                        targetController: 'current',
+                        effect: [
+                            ability.effects.skipStep('ready'),
+                            ability.effects.skipStep('draw')
+                        ]
+                    }),
+                    postHandler: context => {
+                        context.game.endPhaseRightNow = context.player.deck.length > 0 && context.player.deck[0].printedHouse === 'staralliance';
```
jeremylarner: I think you need to check the status of the flag already - if somehow it's already true, you don't want to change it

PR #682, opened 2019-11-03
Added The Colosseum and Gameaction for Glory Counters

Achocholacek:


server/game/cards/03-WC/TheColosseum.js
```diff
@@ -0,0 +1,34 @@
...
+            },
+            effect: 'place a glory counter on itself due to {1} being destroyed',
+            effectArgs: context => [context.event.card],
+            gameAction: ability.actions.addGloryCounter()
+        });
+        this.omni({
+            effect: '{1}',
+            effectArgs: context => [(this.tokens.glory >= 6 && context.player.amber >= context.player.getCurrentKeyCost()) ? 'discard 6 glory counters and forge a key at current cost' : (this.tokens.glory >= 6 && context.player.amber < context.player.getCurrentKeyCost()) ? 'discard 6 glory counters' : 'do nothing'],
+            gameAction: ability.actions.clearGloryCounters(() => {
+                this.preDumpGlory = this.tokens.glory;
```
jeremylarner: Do you need preDumpGlory?  I would expect that if amount is 0, the `then` wouldn't fire....

Achocholacek: apparently not!

PR #685, opened 2019-11-04
final saurian cards

cryogen:


server/game/cards/03-WC/KompsosHaruspex.js
```diff
@@ -0,0 +1,14 @@
+const Card = require('../../Card.js');
+
+class KompsosHaruspex extends Card {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            match: card => card.controller === this.controller && card.type === 'creature',
```
jeremylarner: you don't need to check controller I don't think, the effect engine only affects friendly creatures by default

PR #691, opened 2019-11-05
Tribute should ask to exalt only once

### LINKED ISSUE #667: Murkens and Mimicry
I played Murkens (shadow creature) which allows me to play the top card of my opponent's deck as if it were mine. The card that got played was Mimicry (untamed action) which allows me to copy any action card in my opponent's discard pile. However the game offered up my discard pile to choose a card to copy from, not my opponents. I was able to choose a card to copy but I should have been given my opponent's discard pile.

### LINKED ISSUE #676: Tribute is asking to exalt again after repeating the effect
I play Tribute and decide to exault to repeat the effect. At the end, it asks to exalt again, but did not repeat the effect (maybe because opponent did not have more amber)

romanogit:
Fixes #676

### DIFF
--- a/server/game/cards/03-WC/Tribute.js
+++ b/server/game/cards/03-WC/Tribute.js
@@ -11,17 +11,17 @@ class Tribute extends Card {
                 cardStat: card => card.power,
                 gameAction: ability.actions.capture({ amount: 2 })
             },
-            then: preThenContext => ({
-                condition: context => !preThenContext.secondResolution,
+            then: context => ({
+                condition: () => !context.secondResolution,
                 may: 'exhalt the creature',
-                gameAction: ability.actions.exalt({ target: preThenContext.target }),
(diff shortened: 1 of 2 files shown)


server/game/cards/03-WC/Tribute.js
```diff
@@ -12,10 +12,14 @@ class Tribute extends Card {
...
             },
             then: context => ({
+                condition: () => !context.secondResolution,
                 may: 'exhalt the creature',
                 gameAction: ability.actions.exalt({ target: context.target }),
                 then: {
-                    gameAction: ability.actions.resolveAbility({ ability: context.ability }),
+                    gameAction: ability.actions.resolveAbility({
+                        ability: context.ability,
+                        secondResolution: true
```
cryogen: This will prevent the card from ever triggering again, will it not?

Issue #690, opened 2019-11-05
Chaos Portal + Duskwitch (and possibly any Omega) doesn't End the turn and it should


ira212:
Steps to Repro:
1) Have Chaos Portal in Play
2) Choose Untamed and use Chaos Portal to choose Untamed and reveal Duskwitch on the top of your deck.

Notice that your turn doesn't end.
Expected Behavior: Playing any Omega card (including Duskwitch) should

Here is a deck you can use to test if it's fixed, including a couple different Omega cards (Duskwitch, Gravid Cycle) and Alpha cards too (Glimmer and Binate Rupture) - I'm not sure if Chaos Portal respects Alpha either.

https://www.keyforgegame.com/deck-details/a5e58c6e-f9ed-495d-805a-99eb371a7788

romanogit:
Fixed

PR #700, opened 2019-11-06
Add tests for Chaos Portal

jtrussell:
Tests added to cover the Chaos Portal + Duskwitch (omega cards) interaction mentioned in #690.

It appears that the underlying issue was actually fixed by b393d6c and just hasn't made it to production yet.

### DIFF
--- a/test/server/cards/01-Core/ChaosPortal.spec.js
+++ b/test/server/cards/01-Core/ChaosPortal.spec.js
@@ -1,0 +1,48 @@
+describe('Chaos Portal', function() {
+    integration(function() {
+        describe('Chaos Portal\'s Ability', function() {
+            beforeEach(function() {
+                this.setupTest({
+                    player1: {
+                        house: 'logos',
+                        inPlay: ['chaos-portal'],
+                        discard: ['duskwitch', 'bad-penny']
+                    }
+                });
+            });
+
+            it('should prompt to choose a house', function() {
+                this.player1.useAction(this.chaosPortal);
+                expect(this.player1).toHavePrompt('Chaos Portal');
+                expect(this.player1.currentButtons).toContain('logos');
+                expect(this.player1.currentButtons).toContain('shadows');
+                expect(this.player1.currentButtons).toContain('untamed');
+            });
+
+            it('should play the card if it is of the chosen house', function() {
+                this.player1.moveCard(this.badPenny, 'deck');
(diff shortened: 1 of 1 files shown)


jtrussell:
I wasn't sure if referencing non-cota cards in `tests/../01-Core` was how folks would like to handle this or if we'd rather see this in `tests/.../02-AoA`.

cryogen:
> I wasn't sure if referencing non-cota cards in `tests/../01-Core` was how folks would like to handle this or if we'd rather see this in `tests/.../02-AoA`.

As long as the test is in the folder for the set where the card first appeared, that's fine.  Don't mind which cards various tests refer to.  One trap to avoid falling into is testing unrelated things in your tests.  Eg if you're testing a creature's fight ability fires, you don't want to test that bad penny goes back to hand after she dies, unless that's important to the interaction at hand.  Some of our tests do fall into this trap

test/server/cards/01-Core/ChaosPortal.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                        inPlay: ['chaos-portal'],
+                        discard: ['duskwitch', 'bad-penny']
+                    }
+                });
+            });
+
+            it('should prompt to choose a house', function() {
+                this.player1.useAction(this.chaosPortal);
+                expect(this.player1).toHavePrompt('Chaos Portal');
+                expect(this.player1.currentButtons).toContain('brobnar');
```
cryogen: expect(this.player1).toHavePromptButton('brobnar') is a helper method that exists that would be appropriate here

jtrussell: Sweet - I'll change that up 👍. I have to admit I pretty much copied this straight from `Arise!` and didn't really look at it 😅.

cryogen: The framework has evolved over time, so it's entirely possible that this is a new thing, so don't worry :)

jtrussell:
Looks like the TC build failed. Locally, all (not skipped) tests pass, linting checks out, and the webpack build runs without error.

I don't suppose there's any way I could get at those error messages on TC?

cryogen:
You can see if you click details and log in as a guest. However in this instance the issue is because my deployment server is offline, so nothing for you to do :) it will fix itself shortly!

PR #759, opened 2019-11-08
Fangtooth Cavern fix

Kaosumaru:
Fangtooth Cavern reads "At the end of your turn, destroy the least powerful creature.". Controller should be "any". I've not yet tested it, as I have a while must yet see how to setup keyteki :) But I suspect that this simple change is correct.


cryogen:
Yep that'll do it, thanks!

PR #775, opened 2019-11-09
Fixed SciOfficerMorpheus not triggering Play effect twice

### LINKED ISSUE #714: Play effect on Medic Ingram does not trigger twice when Sci. Officer Morpheus is in play
Apologies, I meant to grab the text log after the game and then forgot.

You can see from the screenshot below that:

- Morpheus gets played,
- Tactical Officer Moon gets played (play effect triggers twice as expected)
- Medic Ingram gets played (play effect triggers once)

image

romanogit:
Fixes #714


cryogen:
copy and paste ftw!

PR #791, opened 2019-11-09
Implement Weasand

romanogit:


jeremylarner:
Why do you need `ignoreWard`?

server/game/TerminalCondition.js
```diff
@@ -26,7 +25,7 @@ class TerminalCondition {
     }
 
     checkCondition() {
-        return this.condition() && (!this.event || this.event.cancelled);
+        return this.condition() && !this.event;
```
jeremylarner: why are you removing this?  Can you test Weasand with an Armageddon Cloak?

server/game/cards/01-Core/TirelessCrocag.js
```diff
@@ -6,19 +6,22 @@ class TirelessCrocag extends Card {
             condition: context => context.player.opponent.creaturesInPlay.length === 0,
```
jeremylarner: If you've got terminal conditions to work, this play ability shouldn't be here

server/game/effectengine.js
```diff
@@ -56,7 +56,8 @@ class EffectEngine {
     checkTerminalConditions() {
         let effectsToTrigger = this.terminalConditions.filter(effect => effect.checkCondition());
         if(effectsToTrigger.length > 0) {
-            this.game.openThenEventWindow(effectsToTrigger.map(effect => effect.getEvent()));
+            let events = effectsToTrigger.reduce((events, effect) => events.concat(effect.getEvent()), []);
```
jeremylarner: why change it from a map to reduce?

server/game/TerminalCondition.js
romanogit: Rolled back TerminalCondition to what it was and added test case to test Armageddon Cloak. It does not work. The creature does not leave play.

server/game/cards/01-Core/TirelessCrocag.js
romanogit: Correct. I will remove it.

server/game/effectengine.js
romanogit: I will rollback this change as well

PR #807, opened 2019-11-10
Fix ReturnToDeckAction to shuffle when card is the last card of the owner

### LINKED ISSUE #805: After play kymoor eclipse, my deck dont get shuffle
It seems it is shuffling only one player deck.
The problem is that the ReturnDeckAction only shuffles at the last card, but since we have mixed controllers, it is causing only one deck to be shuffled.

romanogit:
Fixes #805 

Sample output bellow using console.log to show it shuffled for both players:
```
Started
Shuffling after Troll player1
Shuffling after Streke player2
.
```


romanogit:
@cryogen : any issue with this one? Just wondering....

cryogen:
> @cryogen : any issue with this one? Just wondering....

No just haven't had time to look at it properly

PR #824, opened 2019-11-10
Fixes OrbitalBombardment and other iterative damage cards

### LINKED ISSUE #821: orbital bombardment bypass warding
I tested it yesterday, it seems that Mars card orbital bombardment doesn't remove the wards, and instead directly damages the warded creatures

romanogit:
Fixes #821


server/game/GameActions/AllocateDamageAction.js
```diff
@@ -13,20 +13,23 @@ class AllocateDamageAction extends GameAction {
...
-            numSteps: this.numSteps,
-            selector: this.getSelector(),
-            context: context,
-            onSelect: cardDamage => {
-                for(const uuid of Object.keys(cardDamage)) {
-                    const card = context.game.findAnyCardInPlayByUuid(uuid);
-                    if(card) {
-                        this.events.push(context.game.actions.dealDamage({ amount: cardDamage[uuid] }).getEvent(card, context));
+
+        if(this.numSteps > 0 && this.damageStep > 0) {
```
jeremylarner: maybe this check would be better in `hasLegalTarget`?

romanogit: OB does not work if I add it to `hasLegalTarget`. It tries to target `zero` creatures.

romanogit: @jeremylarner , @cryogen ,
any feedback about this one?

PR #843, opened 2019-11-11
Final forge cost

### LINKED ISSUE #837: The last key forged didn't take Æmber
As title, we notice the game does not take the amber before finishing

SkyJedi:
Fixes #837


server/game/player.js
```diff
@@ -603,6 +603,7 @@ class Player extends GameObject {
                 }
             });
         } else {
+            this.modifyAmber(-modifiedCost);
```
cryogen: There's a lot of duplicated code here, can you refactor this into a function while you're here?

PR #860, opened 2019-11-11
Fixes cost calculation for houseEffects

### LINKED ISSUE #836: Brobnar Ambassador bug: not use, just play
It is not allowing my to use brobnar creatures, just play a new brobnar card. I had bangbang.

Not even in MM

romanogit:
Fixes #836


cryogen:
#866 (i fail at git apparently)

PR #900, opened 2019-11-13
Prompt menu buttons should consider the card house

### LINKED ISSUE #872: Eyegor, two cards, one is maverick.... corner case
Another bug - I played a deck with two copies of Aubade the Grim, one in Sanctum and one in Untamed. When I played Eyegor I didn't know which one it was that was one of the options.

romanogit:
Fixes #872

### DIFF
--- a/server/game/gamesteps/handlermenuprompt.js
+++ b/server/game/gamesteps/handlermenuprompt.js
@@ -40,6 +40,10 @@ class HandlerMenuPrompt extends UiPrompt {
         this.context = properties.context || new AbilityContext({ game: game, player: player, source: this.source });
     }
 
+    id(card) {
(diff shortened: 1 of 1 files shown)


romanogit:
I kept it as a draft because I did not do a visual test. All tests passes, though.

cryogen:
This probably should use the uuid instead of the id  Does it also make it clear on mouse over of the button which is the maverick?

romanogit:
If we use uuid, the problem is that the quantity of the card would be irrelevant. That might be the intention, anyway. As is, the prompt tries to group cards in a single button and I kept the logic. The difference is that I am now considering houses for this particular corner case.

GUI looks like this when we hover over the buttons:
image

versus:

image

Issue #911, opened 2019-11-14
Borr Nit's touch


scyco:
Hi, when I played that card and then I select the one to be purged, it seems that the remaining 4 cards goes into my opponent deck in the same order instead of beeing shuffled into it.

mnapoleon:
Seems to me that the 'Reap' ability on Borr Nit would have the same issue given that the code looks identical in both cards.

mnapoleon:
Is there a simple shuffle deck action?  Seems like that would be useful

romanogit:
I believe this was fixed with #807 . It was merged 6 days ago. Not sure if already in production, though.

mnapoleon:
I don't believe the Borr Nit or Borr Nit's touch actually takes the cards out of the deck so there is no return to deck action, just a purge choice via prompt and it just uses a slice of the card 0-5.  I think #807 modified the Return to Deck action.

romanogit:
Ok, working on this on

scyco:
Thanks @romanogit  !

mnapoleonv:
@romanogit thanks.  Checking out the solution really does make me understand a whole lot more about the way things work and can work.

romanogit:
@mnapoleon : glad you implemented Borr Nit's touch. I did not realize there were two cards almost the same 👍

mnapoleon:
@romanogit no problem. Just a straight copy of your work with a small difference in the tests.  I'm just standing on the shoulder's of giants.

PR #967, opened 2019-11-20
issue 911 updates

mnapoleon:
#911 was resolved but only fixed Borr Nit.  I made same change to Borr Nit's Touch and added tests


cryogen:
Thanks!

PR #982, opened 2019-11-25
Docker-Compose Updates

jtrussell:
This PR includes a few updates - I've left the individual commits for each un-squashed in case we didn't want keep some of those changes. There are more details with each commit, but briefly:

- We pin the node major version to 10. This avoids an incompatibility between ZMQ and node@13. We also get to match up versions with production.
- Minor change to tag the lobby and game nodes with different image names (rather than distinguishing with versions). It looks like the images being built are actually identical (just different start-up commands and ports exposed) - if there's interest we could speed up build times considerably by having a single image that gets referenced by both services (i.e. still two containers, but just one image). This might be nice as the images are fairly large, ~1.7 GB, and nobody likes waiting for node modules to install twice :).
- I exposed the mongodb port so folks could connect with their own clients running outside the container to look at the keyteki data.

I didn't want to muddy the waters too much. I'll have a separate PR for mounting source files as volumes so folks can edit source files outside the containers and see their changes without having to rebuild everything.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM node:latest
+FROM node:10
 RUN mkdir -p /usr/src/lobby
 WORKDIR /usr/src/lobby
 COPY package.json /usr/src/lobby/
 COPY package-lock.json /usr/src/lobby/
-RUN npm install
+RUN npm install --no-optional
 COPY . /usr/src/lobby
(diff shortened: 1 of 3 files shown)


cryogen:
Thanks!  I've changed the external mongo port to end in 27 to remove the conflict if you are running mongo locally

PR #989, opened 2019-11-27
Use single Dockerfile, mount local src directories

jtrussell:
The commit does two things (sorry - should have made two separate
commits).

First, we remove the gamenode Dockerfile. I.e. the one that was
previously being used to build the `node` service. The `lobby` and
`node` Dockerfiles were close to identical - the primary differences
were the name of the workspace directory (doesn't really matter) and the
ports that were being exposed (can be done when the service is created).
By having one Dockerfile/image rather than two (or, I suppose we could
have two identical Dockerfiles), we get much faster builds. The first
image build runs as normal, the second will run as well but gets to pull
everything from cache.

The second change is to mount local source directories into the service
containers. This lets us push changes to the running app without
rebuilding. HMR updates will flow through where they can, and other
updates you can get with a `docker-compose restart` rather than having
to rebuild the underlying image. I could definitely use some help to
make sure the right folders are getting mounted here.


docker-compose.yml
```diff
@@ -2,12 +2,16 @@ version: "3.3"
...
     restart: always
-    build:
-      context: .
-    image: keyteki-lobby
+    build: .
+    image: keyteki
     ports:
       - "4000:4000"
       - "9229:9229"
+    volumes:
```
jtrussell: Just doing:

```
volumes:
  - ".:/usr/src/app"
```

Didn't seem to work, I didn't dig through it too much but in retrospect it may have had something to do with trying to use my local node modules within the container.

Dockerfile
```diff
@@ -1,10 +1,7 @@
 FROM node:10
-RUN mkdir -p /usr/src/lobby
-WORKDIR /usr/src/lobby
-COPY package.json /usr/src/lobby/
-COPY package-lock.json /usr/src/lobby/
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
```
jtrussell: Changed the `WORKDIR` to something more general since the resulting image would be used by both the `node` and `lobby` images.

Dockerfile
```diff
@@ -1,10 +1,7 @@
...
+COPY package.json /usr/src/app
+COPY package-lock.json /usr/src/app
 RUN npm install --no-optional
-COPY . /usr/src/lobby
-EXPOSE 4000
```
jtrussell: Moved all port handling to the docker-compose script since which port gets exposed will depend on which service the image is used for.

PR #1003, opened 2019-12-02
Add focus on deck list and import UI

### LINKED ISSUE #1002: Deck List and Import Deck screens should have default UI focus
While importing 20+ decks today, I thought it'd be a nice touch for the Import Deck URL input box to gain focus when entering the screen.

Part 1:
**Steps to Reproduce**
1. Click on "Decks"

**Expected Result**
The focus to be on the filter input

**Actual Result**
There is no focus on any element

Part 2:
**Steps to Reproduce**
1. Click on "Decks"
1. Click on "Import Deck" button

**Expected Result**
The focus to be on the URL input

**Actual Result**
There is no focus on any element

Merge request incoming.

avehlies:
This is my first React code so please let me know if there's an issue with using `React.createRef()`

Fixes #1002

### DIFF
--- a/client/Components/Decks/DeckList.jsx
+++ b/client/Components/Decks/DeckList.jsx
@@ -98,7 +98,7 @@ class DeckList extends React.Component {
                 <form className='form'>
                     <div className='col-md-8'>
                         <div className='form-group'>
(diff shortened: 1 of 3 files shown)


avehlies:
After doing some more reading, it looks like using `React.createRef()` isn't great. I decided to just add `autoFocus` as a property to `Input` and pass it down to the `<input/>` tag.

cryogen:
Thanks!

PR #1023, opened 2019-12-07
Fix prompt for order inconsistency.

### LINKED ISSUE #945: Assault and Hazardous does not prompt for ordering
They happen in the same place on the timing chart, therefor as the attacker I should be able to choose the order they happen in. In this game I attacked a xenos with my moon and the moon just died without dealing the assault damage https://pastebin.com/Yh8VZtTu
Sorry the screenshot is right before this happened, I encountered multiple bugs this game and lost the other screenshot of after 
image

### LINKED ISSUE #1022: "Prompt to order simultaneous abilities" behavior is inconsistent
Some behaviors, like Assault vs. Hazardous does not prompt for order even if "Prompt to order simultaneous abilities" checkbox is clicked during game settings. Other abilities ask for prompt, though, for example, assault vs. a before fight.

romanogit:
Fixes #1022
Check SimultaneousEffectWindow.js for reference
Also fixes #945

### DIFF
--- a/server/game/gamesteps/forcedtriggeredabilitywindow.js
+++ b/server/game/gamesteps/forcedtriggeredabilitywindow.js
@@ -48,7 +48,7 @@ class ForcedTriggeredAbilityWindow extends BaseStep {
         }
 
         this.noOptionalChoices = this.choices.every(context => !context.ability.optional);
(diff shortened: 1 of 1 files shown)


romanogit:
I broke a lot of tests :(

PR #1034, opened 2019-12-12
Added number of keys to the end of turn message (Issue#608)

admiraldeathrain:
I just realized the original issue requested this information to be displayed at the end of a turn rather than in the key-step, so this needs some discussion on what is actually the desireable behaviour.


admiraldeathrain:
grafik
The key step message

admiraldeathrain:
Moved the message to the end of round instead, it just makes more sense for providing an overview.

admiraldeathrain:
#1088 Does is complete and does it better.

PR #1041, opened 2019-12-17
Fix order trigger setting bug

jeremylarner:


romanogit:
Is this bug fix similar to this one:
https://github.com/keyteki/keyteki/pull/1023

PR #1045, opened 2019-12-19
ImperialTraitor is optional and should show hand without Sanctum card

### LINKED ISSUE #1038: Imperial Traitor does not allow you to see the opponents hand if they don't have any Sanctum cards
Imperial traitor says "Look at opponent's hand. You may choose and purge a sanctum card." When playing Worlds Collide decks if you play the card, it does not allow you to see their hand. This should happen regardless of whether they have Sanctum or not.

romanogit:
Fixes #1038


romanogit:
This one Fixes #1038

PR #1061, opened 2019-12-25
Deck search applies to all search pages

avehlies:
* Fixes the search functionality of decks. Searches through all decks and not just the ones that appear on the current page of pagination.
* Re-paginates results after searching


cryogen:
Thanks!

Issue #677, opened 2019-11-01
City-State Interest: should prompt for creatures if opponent's amber less than friendly creatures in play


romanogit:
City-State Interest: should prompt for creatures if opponent's amber less than friendly creatures in play

At the moment, it is choosing the creatures in order of play or battleline (the left most captured the amber)

Other WC card that might have the same problem:
- [x] Xenotraining

Other non-WC cards that might have the same problem:
- [x] Yxili Marauder
- [x] Shard of Hope
- [x] Unguarded Camp
- [x] Pandemonium
- [x] Mars Needs Æmber
- [x] Honorable Claim

New MM
- [x] Spoils of Battle

Those are creatures that capture for each something.

bensissel:
this is still happening for me. Any idea what can be done to fix this?

cryogen:
It's still happening because the issue is not fixed, this issue will be closed when it is fixed

romanogit:
Jadiel fixed this with the CaptureAction refactor. I will go through the cards, simplify them if needed and add test cases.

romanogit:
I checked all cards above and all of them are implemented ok and also using the CaptureAction refactor.

PR #1063, opened 2019-12-26
Fix for City State Interest

### LINKED ISSUE #1037: City-State Interest doesn't prompt which creatures you want to capture the aember on
City-State Interest works fine if opponent has enough aember for each creature to capture one.
But when I have 6 creatures and opponent has let's say 3 aember it will just capture from left to right and does not let you chose which creature you want to put it on.

LifeIsPizza:
Fixes #1037

### DIFF
--- a/server/game/cards/03-WC/CityStateInterest.js
+++ b/server/game/cards/03-WC/CityStateInterest.js
@@ -3,10 +3,21 @@ const Card = require('../../Card.js');
 class CityStateInterest extends Card {
     setupCardAbilities(ability) {
         this.play({
-            gameAction: ability.actions.capture(context => ({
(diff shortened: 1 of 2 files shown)


romanogit:
I think this should be a more generic solution to capture. Other cards could benefit.

LifeIsPizza:
> 
> 
> I think this should be a more generic solution to capture. Other cards could benefit.

Yeah it's basically copied from Pandemonium. Not sure if there's other similar cards that could use this kind of effect.

romanogit:
This issue https://github.com/keyteki/keyteki/issues/677 shows cards that also (might) have the same bug

Issue #608, opened 2019-10-23
Recap (in brackets) the number of keys - Save on clipboard the entire log


meta-lupo:
Mr./Mrs. @cryogen is it possible to recap (in brackets) the number of the players keys near the total æmbers at the end of every turn? Helps a lot when People want to read the saved log after the game, to recap errors and so. And is it possible to add a button that saves on clipboard the entire log or select all the log for copying to facilitate archiving games?

Thank you a lot.

(you asked me to put this request as an issue in thecrucible.online chat yesterday; see you again and thank you)

donders:
Copying text that is not in an input field is only possible in Chrome atm AFAIK.

PR #1088, opened 2020-01-14
number of keys in brackets and saving to clipboard

donders:
Solves #608

### DIFF
--- a/client/Components/GameBoard/PlayerStats.jsx
+++ b/client/Components/GameBoard/PlayerStats.jsx
@@ -57,6 +57,16 @@ export class PlayerStats extends React.Component {
         );
     }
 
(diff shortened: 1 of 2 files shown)


client/Components/GameBoard/PlayerStats.jsx
```diff
@@ -56,6 +57,16 @@ export class PlayerStats extends React.Component {
         );
     }
 
+    writeChatToClipboard(event) {
+        event.preventDefault();
+        let messagePanel = document.getElementsByClassName('messages panel')[0];
```
cryogen: This will work, but it’s not very “react”. It should use refs to get the handle to the Dom element

Issue #577, opened 2019-10-15
Maverick symbol on the first card in your hand is covered up by the hand size indicator


saluk:
It's difficult to see that a card in the first slot in your hand is a maverick.
https://i.imgur.com/jf2N8H9.png

dmikalova:
New UI fixes this

PR #1096, opened 2020-01-17
Moved label position outside of card area

admiraldeathrain:
Possible fix for #577


cryogen:
Thanks for the PR! Do you have a screenshot of how this looks?

admiraldeathrain:
grafik
It kind of is the minimal possible change to adress the issue. Possibly creates issues with card text legibility, but on most screens you need to hover over for that anyways. Does scale with the different card heights.

cryogen:
Hmm, I think both are not all that great, my prefered solution to this would probably be transpararentish tabs with the header info on it, I think that might work better.

admiraldeathrain:
@cryogen Overlapping a the card are or not? I think overlapping would look better and more compact, but probably less readable.

cryogen:
Preferably not, just a little tab I think would work quite well?

admiraldeathrain:
grafik
I introduced a margin for the card-areas and moved the label outside of the card area, this is how it looks now.

romanogit:
My personal opinion : I liked the label, but I think opponent's label could be bellow card area.

drallieiv:
I forgot to bookmark the deck it was from, but when a Gigantic creature gets an upgrade, the additional icon pips are shown on the real card in the top left corner when looking at the card in landscape mode, so in the top right corner when fanning the cards in portrait mode.

I don't know if we should respect that display in TCO, or if we do show the bonus pips in the same corner then a classic card.

cryogen:
Sorry for not responding sooner. But I was never really happy with how this looked. It's definitely still a problem, but one for which we need a different solution I think.

PR #1098, opened 2020-01-19
Fix to Thorium Plasmate

### LINKED ISSUE #939: Thorium plasmate does not consistently relocate creatures to where they are selected to move
I don't have screenshots or game logs, but there really isn't anything you could gain from them since there is no indication in the logs of where I clicked to move a creature.That said this occurs in nearly every single game I play with my double plasmate deck so it shouldn't be super hard to reproduce (I'll come back with screenshots when I get a chance). Essentially instead of moving the creature to the right of the creature selected or left of creature selected it moves the creature several spaces further to the selected direction. This happens roughly 70% of the time. I cannot fix this in manual mode either because I cannot figure out how to move creatures in manual mode. All other battleline reordering cards seem to work as intended. 
EDIT: I did more testing and it "further in that direction" seems to not be the case, it just randomly moves things around

jtrussell:
MoveOnBattlelineAction was determining a new index for moved cards based on a battleline with the card being moved still in it.

The "move" happens by first splicing out the card to be moved then splicing it back it. This means we should calculate our "slice back in" index assuming the moved card is not in the battleline.

Fixes #939

### DIFF
--- a/server/game/GameActions/MoveOnBattlelineAction.js
+++ b/server/game/GameActions/MoveOnBattlelineAction.js
@@ -59,9 +59,10 @@ class MoveOnBattlelineAction extends CardGameAction {
         return super.createEvent('onCardMovedInBattleline', { card: card, context: context }, () => {
             let player = card.controller;
             let cardIndex = player.cardsInPlay.indexOf(card);
+            let cardInsertionIndex = this.moveIndex > cardIndex ? this.moveIndex - 1 : this.moveIndex;
 
             player.cardsInPlay.splice(cardIndex, 1);
-            player.cardsInPlay.splice(this.moveIndex, 0, card);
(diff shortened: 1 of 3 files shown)


jtrussell:
Side note - I tired (not exhaustively) to play around with using `creaturesInPlay` instead of `cardsInPlay` in the move action but couldn't get a new test to fail.

jtrussell:
Regarding the fixed Weasand test - the goal of that test was to use Plasmate to get Weasand on a flank and make sure it dies. I think we just had left/right swapped (likely fudged to accommodate  Plasmate misbehaving). That aside, I'm curious on the ruling for how much damage Dodger should get in that test. Should Weasand get a chance to die before Plasmate checks for neighbors?

Issue #1055, opened 2019-12-21
Crassosaurus not purging 


Bigstubbo:
I have just played Crassosaurus and only put 3 ember on it, when I clicked done I expected it to be purged,  it was not it just sat on my side like any other creature.

romanogit:
![image](https://user-images.githubusercontent.com/2104075/71681253-f21d2100-2d6a-11ea-9287-dfd483331118.png)

Bigstubbo:
Yes that is the card, should there be a comment?

romanogit:
@Bigstubbo : just added the card here for quick reference

PR #1102, opened 2020-01-20
Crassosaurus should purge after capturing 0 from self

jtrussell:


jtrussell:
There are still things to fix with Crassy. You should be *forced* to capture at least enough from your opponent to get to 10 total if it's possible to do so, but this at least gets us out of "super Charette" territory 😂.

jtrussell:
Fixes #1055.

PR #1135, opened 2020-02-13
Fix Into the Night

### LINKED ISSUE #882: Into the Night's effect does not work if you play Relentless Assault later
I play Into the Night, so non-Shadows creatures cannot fight until my next turn.
My opponent plays Relentless Assault and fights with his Urchin (Shadows, this is OK), but then he also fights with Troll, which is Brobnar (this is not OK).
I think this is a bug that basically let's you fight with any creatures as long as you use a Shadows one first to fight.

Correct expected behaviour in my case would be that only Urchin would be allowed to fight.

### LINKED ISSUE #1062: Golden Spiral + Spartasaur overcame Into the Night
In [this game](https://www.thecrucibletracker.com/games/efa833b0-26b7-11ea-b905-f7aec34d7932), nedmike plays [Into the Night](https://archonarcana.com/Into_the_Night) on turn 8, which should prevent non-Shadows creatures from fighting during the next turn.

During turn 9, anonymous plays [Spartasaur](https://archonarcana.com/Spartasaur) and uses their existing 2 [Golden Spirals](https://archonarcana.com/The_Golden_Spiral) to exalt and fight with Spartasaur twice.  I don't think it should have been able to fight at all because of the Into the Night.

(During the game we used manual mode after turn 9 to correct anonymous's Aember count, as if they had reaped twice instead.)

image
image
image

corben2:
Fixes #882 and Fixes #1062

### DIFF
--- a/server/game/cards/03-WC/IntoTheNight.js
+++ b/server/game/cards/03-WC/IntoTheNight.js
@@ -4,10 +4,9 @@ class IntoTheNight extends Card {
     setupCardAbilities(ability) {
         this.play({
             effect: 'stop all non-Shadows creatures from fighting until the start of their next turn',
-            gameAction: ability.actions.cardLastingEffect(context => ({
-                duration: 'untilNextTurn',
-                target: context.game.creaturesInPlay.filter(card => !card.hasHouse('shadows')),
-                effect: ability.effects.cardCannot('fight')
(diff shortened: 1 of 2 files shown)


cryogen:
Thanks for the fix and sorry for the extended time taken to review it!

Issue #375, opened 2019-08-24
Master the Theory


CalumSLZ:
Master theory says 'may' archive, so i should be able to choose to archive fewer cards than my opponent has creatures even if the board is empty. However it forces you to archive that number, so leaving no choice.

cryogen:
I suppose that depends on how you read the card. That way I read it (and there being no “up to” involved) I’d say the way we have it is right. But if you can point me at a ruling that says otherwise we can change it

4ugeistr:
From Discord #rules-questions:
![image](https://user-images.githubusercontent.com/23107141/63843172-9c84d200-c98e-11e9-9479-73e2aa27a94f.png)

ryanch:
I'll see if I can get a pull request up for this.

PR #1159, opened 2020-03-15
fixing master the theory

ryanch:
This is a fix for the issue:
https://github.com/keyteki/keyteki/issues/375 

To make this fix, I had to update the ExactXSelector to support an optional number of cards.  This was needed because the UpToSelector does not support a function for numCards. This seemed approach seemed safer, because otherwise all of the cards that use upTo would need to be updated.

I tested this change by:
1) Creating a test game where I confirmed the new behavior for Master The Theory Works
2) I tested Twin Bolt Emission (since it uses 'exact' ) I confirmed it is not broken, and still requires 2 targets, but does fail  if there is only 1.
3) I added a test class for Master The Theory.

(Note I posted this twice, because the first time I accidentally included my other changes into this fix.)

### DIFF
--- a/server/game/CardSelectors/ExactlyXCardSelector.js
+++ b/server/game/CardSelectors/ExactlyXCardSelector.js
@@ -25,7 +25,7 @@ class ExactlyXCardSelector extends BaseCardSelector {
     }
 
     hasEnoughSelected(selectedCards, context) {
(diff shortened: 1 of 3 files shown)


ryanch:
Per Romano@ I will change this so the UpToSelector supports a function.

PR #1161, opened 2020-03-15
Master The Theory Fix to support 'may'

ryanch:
Alright, 3rd attempt.

Updated the Master The Theory per the issue here:
https://github.com/keyteki/keyteki/issues/375

In this update I have:
* Updated the UpToCardSelector to take a function for num cards
* Updated Master The Theory to use upTo
* Added a test case for Master The Theory
* Added a test case for Natures Call

Testing I have done:
* I confirmed Natures Call works correctly (since it uses upTo)
* Confirmed Master The Theory works correctly
* Lint & Test cases pass

### DIFF
--- a/server/game/CardSelectors/ExactlyXCardSelector.js
+++ b/server/game/CardSelectors/ExactlyXCardSelector.js
@@ -25,7 +25,7 @@ class ExactlyXCardSelector extends BaseCardSelector {
     }
 
     hasEnoughSelected(selectedCards, context) {
-        return selectedCards.length === this.getNumCards(context) || this.optional;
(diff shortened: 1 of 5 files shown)


romanogit:
@cryogen , I have the impression that `optional` will be either none or all.

ryanch:
@cryogen romanogit is right. If I use "may:" then it solves part of the problem but not all of it.

If the game state is:
* No local creatures
* 3 opponents creatures

I should be able to Archive 0, 1, 2 or 3 cards.

With "may:" I would only be able to archive 0 or 3 cards.

PR #1166, opened 2020-03-16
Maintenance/ui fixes and name changes

QuantariusRay:
Set player health to 25
Remove instances of crucible and keyforge to Manacrest
Remove images/keys/amber
Remove some phases that aren't necessary
Added necessary phases


cryogen:
Hi I think you've raised this request against the wrong repository!

PR #1180, opened 2020-03-20
feat: implement diametric charge

granttitus:
image


granttitus:
Does the TCO team write tests for all cards or just complex ones?

cryogen:
Tests are preferred for all cards, but we'll take what we can get.  WC has pretty good test coverage, earlier packs less so

PR #1181, opened 2020-03-20
feat: implement gizelharts wrath

granttitus:
image

### DIFF
--- a/server/game/cards/04-MM/GizelhartsWrath.js
+++ b/server/game/cards/04-MM/GizelhartsWrath.js
@@ -1,0 +1,15 @@
+const Card = require('../../Card.js');
+
+class GizelhartsWrath extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            gameAction: ability.actions.destroy(context => ({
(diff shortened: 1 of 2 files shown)


cryogen:
This one has a genuine lint error, and isn't caused by the build issues

romanogit:
I think you should not use *target* here. 

```javascript
this.play({
            effect: 'destroy each undamaged creature',
            gameAction: ability.actions.destroy(context => ({
                    target: context.game.creaturesInPlay.filter(card => !card.hasTrait('mutant'))
                }))
        });
```

granttitus:
> I think you should not use _target_ here.
> 
> ```js
> this.play({
>             effect: 'destroy each undamaged creature',
>             gameAction: ability.actions.destroy(context => ({
>                     target: context.game.creaturesInPlay.filter(card => !card.hasTrait('mutant'))
>                 }))
>         });
> ```

Yes, you are correct. I would have caught that if I had written tests!

granttitus:
I fixed the implementation and added tests

PR #1187, opened 2020-03-20
feat: implement bonesaw

granttitus:
image

I mimicked the approach used for Alaka and added a `creatureDestroyed` attribute to players.

image


cryogen:
Leaving this one for the moment as both of these cards should really use an event registrar rather than setting properties on the player

cryogen:
Merge now, fix later

PR #1189, opened 2020-03-21
feat: implement stirring grave

granttitus:
image


server/game/cards/04-MM/StirringGrave.js
```diff
@@ -0,0 +1,18 @@
+const Card = require('../../Card.js');
+
+class StirringGrave extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
+                cardType: ['creature'],
```
cryogen: FYI arrays are optional if there's only one element

Issue #1172, opened 2020-03-17
StarAlliance House - Mass Mutation


romanogit:
Star Alliance 60 / 40 New 67%
---
- [x] 305 - [Crewman Jörg](https://cdn.keyforgegame.com/media/card_front/en/479_305_4QW44X7Q58QV_en.png)
- [x] 306 - [Dæmo-Alien](https://cdn.keyforgegame.com/media/card_front/en/479_306_8V3P9R7QCHVV_en.png)
- [x] 308 - [Hadron Collision](https://cdn.keyforgegame.com/media/card_front/en/479_308_55R349MHQ2P7_en.png)
- [x] 309 - [Observ-u-Max](https://cdn.keyforgegame.com/media/card_front/en/479_309_58M3CG48V3QC_en.png)
- [x] 321 - [Detention Coil](https://cdn.keyforgegame.com/media/card_front/en/479_321_VFQGV2CCWH22_en.png)
- [x] 324 - [LCdr. Trigon](https://cdn.keyforgegame.com/media/card_front/en/479_324_R242QFCM77C_en.png)
- [x] 325 - [Mutagenesis Researcher](https://cdn.keyforgegame.com/media/card_front/en/479_325_W237RJ5H5V4R_en.png)
- [x] 326 - [New Frontiers](https://cdn.keyforgegame.com/media/card_front/en/479_326_5W8C7PWFJW8_en.png)
- [x] 329 - [Stealthster](https://cdn.keyforgegame.com/media/card_front/en/479_329_Q48JP863C362_en.png)
- [x] 335 - [Ambassador Liu](https://cdn.keyforgegame.com/media/card_front/en/479_335_582P839P6HF9_en.png)
- [x] 337 - [Commander Dhrxgar](https://cdn.keyforgegame.com/media/card_front/en/479_337_6FHWP43QGPJW_en.png)
- [x] 345 - [Hapless Cadet](https://cdn.keyforgegame.com/media/card_front/en/479_345_3JPX2CHRGR6J_en.png)
- [x] 346 - [Hold the Line](https://cdn.keyforgegame.com/media/card_front/en/479_346_G62M782WF22P_en.png)
- [x] 347 - [J.O.N. Cargo](https://cdn.keyforgegame.com/media/card_front/en/479_347_R4XHX8JQ3RJ_en.png)
- [x] 348 - [Kartanoo](https://cdn.keyforgegame.com/media/card_front/en/479_348_4WHJ78MJRP7M_en.png)
[...]

PR #1195, opened 2020-03-22
adding access denied card

ryanch:
Adding Access Denied Card.
* Added Card
* Added Test
* lint and tests pass


ryanch:
This is part of this issue: https://github.com/keyteki/keyteki/issues/1172

Issue #1168, opened 2020-03-17
Logos House - Mass Mutation


romanogit:
Logos 60 Cards / 37 New 62%
---
- [x] 099 - [Adaptoid](https://cdn.keyforgegame.com/media/card_front/en/479_099_8JCF2WRJH9WC_en.png)
- [x] 101 - [Auto-Vac 5150](https://cdn.keyforgegame.com/media/card_front/en/479_101_6R4HGCMP84FP_en.png)
- [x] 113 - [The Archivist](https://cdn.keyforgegame.com/media/card_front/en/479_113_W63QXFW9GX95_en.png)

---

- [x] 066 - [Auto-Encoder](https://cdn.keyforgegame.com/media/card_front/en/479_066_CPHXV4PF7PC7_en.png)
- [x] 073 - [Even Ivan](https://cdn.keyforgegame.com/media/card_front/en/479_073_HP57J82528CR_en.png)
- [x] 074 - [Infomorph](https://cdn.keyforgegame.com/media/card_front/en/479_074_WX8FX4JR7R5R_en.png)
- [x] 081 - [Titan Engineer](https://cdn.keyforgegame.com/media/card_front/en/479_081_46X98FVV69Q3_en.png)
- [x] 083 - [Causal Loop](https://cdn.keyforgegame.com/media/card_front/en/479_083_XWV688QRWQH6_en.png)
- [x] 087 - [Fission Bloom](https://cdn.keyforgegame.com/media/card_front/en/479_087_5HW74C3JM99M_en.png)
- [x] 088 - [Forge Compiler](https://cdn.keyforgegame.com/media/card_front/en/479_088_9HC4QFXWCVRP_en.png)
- [x] 090 - [Krrrzzzaaap!!!](https://cdn.keyforgegame.com/media/card_front/en/479_090_F6CQJH6R6GHM_en.png)
- [x] 095 - [Professor Terato](https://cdn.keyforgegame.com/media/card_front/en/479_095_WGJHF2XQ329X_en.png)
- [x] 100 - [Animator](https://cdn.keyforgegame.com/media/card_front/en/479_100_8RCWGMCG7FCV_en.png)
- [x] 105 - [Library Card](https://cdn.keyforgegame.com/media/card_front/en/479_105_JMWVQGCR4P37_en.png)
- [x] 121 - [Odd Clawde](https://cdn.keyforgegame.com/media/card_front/en/479_121_46F9MVPJ46R4_en.png)
[...]

PR #1197, opened 2020-03-23
MM: Logos Lethologica, Mind over Matter, Q-Mechs

SkyJedi:
https://github.com/keyteki/keyteki/issues/1168


romanogit:
I believe DiametricCharge was commited here:
https://github.com/keyteki/keyteki/pull/1180

PR #1200, opened 2020-03-23
MM: Saurian Consul Primus

romanogit:
image

### DIFF
--- a/server/game/cards/04-MM/ConsulPrimus.js
+++ b/server/game/cards/04-MM/ConsulPrimus.js
@@ -17,7 +17,7 @@ class ConsulPrimus extends Card {
                         cardType: 'creature',
                         activePromptTitle: 'Choose another creature',
                         cardCondition: card => card !== preContext.target
-            }
+                    }
                 })
             })
         });
--- a/server/game/cards/04-MM/ConsulPrimus.js
+++ b/server/game/cards/04-MM/ConsulPrimus.js
@@ -3,7 +3,7 @@ const Card = require('../../Card.js');
 class ConsulPrimus extends Card {
     setupCardAbilities(ability) {
         this.reap({
-            condition: context => context.game.creaturesInPlay.length > 1 && context.game.creaturesInPlay.some(card => card.hasToken('amber')),
+            condition: context => context.game.creaturesInPlay.length > 1,
             target: {
                 cardType: 'creature',
                 cardCondition: card => card.hasToken('amber'),
@@ -17,7 +17,7 @@ class ConsulPrimus extends Card {
                         cardType: 'creature',
                         activePromptTitle: 'Choose another creature',
                         cardCondition: card => card !== preContext.target
-                    }
+            }
                 })
             })
         });
--- a/server/game/cards/04-MM/ConsulPrimus.js
+++ b/server/game/cards/04-MM/ConsulPrimus.js
@@ -1,0 +1,29 @@
+const Card = require('../../Card.js');
+
+class ConsulPrimus extends Card {
+    setupCardAbilities(ability) {
+        this.reap({
+            condition: context => context.game.creaturesInPlay.length > 1 && context.game.creaturesInPlay.some(card => card.hasToken('amber')),
+            target: {
+                cardType: 'creature',
+                cardCondition: card => card.hasToken('amber'),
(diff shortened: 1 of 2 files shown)


server/game/cards/04-MM/ConsulPrimus.js
```diff
@@ -0,0 +1,29 @@
+const Card = require('../../Card.js');
+
+class ConsulPrimus extends Card {
+    setupCardAbilities(ability) {
+        this.reap({
+            condition: context => context.game.creaturesInPlay.length > 1 && context.game.creaturesInPlay.some(card => card.hasToken('amber')),
```
cryogen: Is this not redundant?

server/game/cards/04-MM/ConsulPrimus.js
```diff
@@ -0,0 +1,29 @@
...
+            target: {
+                cardType: 'creature',
+                cardCondition: card => card.hasToken('amber'),
+                gameAction: ability.actions.removeAmber()
```
cryogen: This feels like it should be a "moveAmber" action and done all in one

server/game/cards/04-MM/ConsulPrimus.js
romanogit: Not really, we need at least 2 creatures, one of which must have amber on it.

cryogen: But does the target condition not already cover this?

romanogit: No, it does not.
We had the same discussion by the time when ChantOfHubris was implemented 😄 
I cannot have just 1 creature with amber. Target looks for a creature with amber, but I need to have at least another creature, that's why I added the condition.

server/game/cards/04-MM/ConsulPrimus.js
romanogit: I do not think we have a moveAmber action at the moment. I think all cards that move amber does remove + place.

server/game/cards/04-MM/ConsulPrimus.js
cryogen: Then it should just be the first half, as the second half is covered by the target

server/game/cards/04-MM/ConsulPrimus.js
cryogen: Fancy adding one?

server/game/cards/04-MM/ConsulPrimus.js
romanogit: Ok, I will try it

server/game/cards/04-MM/ConsulPrimus.js
romanogit: It would require a double target :) I can try as well

server/game/cards/04-MM/ConsulPrimus.js
romanogit: I removed the extra condition check

server/game/cards/04-MM/ConsulPrimus.js
romanogit: I did not implement the moveAmber. It has many implications which the engine cannot handle. I could implement a naive solution, but I could foresee cases where it would be difficult to adjust in the future. 

First is the problem of duplicating the same event type in multiple places. We have an addToken which could be modified to consider future card effects, which would have to be duplicated in MoveAmber.

A pure promptForSelect would not be able to apply all the canAffect/legalTarget validation which is part of the GameAction implementation. For example, if in the future we have a card that says "this creature cannot hold any amber", we could easily add an effect check for AddToken. We would have to do the same in MoveAmber pfs.

Ideally, the best thing to do would be some sort of composing gameactions allowing a multiple target as input.

Something like this:
```javascript
    condition: context => context.game.creaturesInPlay.length > 1,
            target: 
                source: {
                    cardType: 'creature',
                    cardCondition: card => card.hasToken('amber')                    
                },
                destination: {
                    //dependsOn: 'source'; implicit
                    cardType: 'creature'
                    // implicit cardCondition: card => card !== preContext.target 
                }                
            }),
            gameAction: ..moveAmber()
```

But I do not know how to do this double targe moveAmber.

PR #1201, opened 2020-03-23
MM: Logos Munchling

SkyJedi:
https://github.com/keyteki/keyteki/issues/1168

### DIFF
--- a/test/server/cards/04-MM/Munchling.spec.js
+++ b/test/server/cards/04-MM/Munchling.spec.js
@@ -18,7 +18,7 @@ describe('munchling', function () {
             });
 
             it('give prompt', function () {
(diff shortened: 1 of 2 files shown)


test/server/cards/04-MM/Munchling.spec.js
```diff
@@ -0,0 +1,70 @@
...
+                    },
+                    player2: {
+                        inPlay: ['helper-bot']
+                    }
+                });
+                this.player1.fightWith(this.munchling, this.helperBot);
+            });
+
+            it('give prompt', function () {
+                expect(this.player1).toHavePrompt('Any reactions to Helper Bot leaving play?');
```
cryogen: Generally you shouldn't test for this specific test, better to test for a card being clickable or something like that

PR #1231, opened 2020-03-29
MM: Star Alliance - Matter Maker

ryanch:
Added Matter Maker card per issue:
https://github.com/keyteki/keyteki/issues/1172

Validation done:
* New test case passes
* Lint passes

Note: I could not find documentation that "*" would work for 'house:', but it does. 

I also tried:
* leaving it out
* the value ''
* providing a function doing something like done in Val Jerico - like: `(card, context) => context.player.activeHouse`. But it seems that "house" does not accept a function. 

However, using "*" does pass the test for both Star Alliance and Non-Star Alliance turns.

### DIFF
--- a/server/game/cards/04-MM/MatterMaker.js
+++ b/server/game/cards/04-MM/MatterMaker.js
@@ -4,10 +4,7 @@ class MatterMaker extends Card {
     setupCardAbilities(ability) {
         this.persistentEffect({
             targetController: 'current',
-            effect: ability.effects.canPlayNonHouse(({
-                house: '*',
-                condition: card => card.type === 'upgrade'
-            }))
+            effect: ability.effects.canPlay( card => card.type === 'upgrade' )
         });
     }
 }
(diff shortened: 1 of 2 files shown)


server/game/cards/04-MM/MatterMaker.js
```diff
@@ -0,0 +1,17 @@
+const Card = require('../../Card.js');
+
+class MatterMaker extends Card {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'current',
+            effect: ability.effects.canPlayNonHouse(({
```
cryogen, line 7: I think you just want canPlay here

ryanch: Nice catch. There was actually a bigger problem here. I think the old code was letting ANY card be played out of house. I've updated to use ```canPlay``` and added a test case to check for not being able to play creatures and actions. Fixed in update.

PR #1233, opened 2020-03-29
MM: Star Alliance - Ant1-10ny

SkyJedi:
https://github.com/keyteki/keyteki/issues/1172

### DIFF
--- a/server/game/cards/04-MM/Ant110ny.js
+++ b/server/game/cards/04-MM/Ant110ny.js
@@ -1,0 +1,26 @@
+const Card = require('../../Card.js');
+
+class Ant110ny extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            gameAction: ability.actions.capture(context => ({
+                amount: context.player.opponent.amber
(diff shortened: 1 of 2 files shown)


server/game/cards/04-MM/Ant110ny.js
```diff
@@ -0,0 +1,26 @@
...
+        });
+        this.interrupt({
+            when: {
+                onRoundEnded: (event, context) =>
+                    context.player === this.game.activePlayer
+            },
+            condition: context => context.source.hasToken('amber'),
+            gameAction: [
+                ability.actions.removeAmber({ amount: 1 }),
+                ability.actions.gainAmber(context => ({ target: context.player.opponent }))
```
romanogit: You might want to add a test in conjunction with Exile, but I think considering other cards like Shaffle and Scowly Capper, it will sorry!

SkyJedi: added test and added what you told me to add master

PR #1244, opened 2020-04-01
GameMode: Reckoning - Adaptive bo1

SkyJedi:
I reckonin`


client/Components/Games/NewGame.jsx
```diff
@@ -36,6 +36,7 @@ class NewGame extends React.Component {
             password: '',
             useGameTimeLimit: false,
             hideDecklists: false,
+            adaptive: { chains: 0, selection: [], biddingWinner: '', match1Winner: '', match2Winner: '', match3Winner: '' },
```
cryogen, line 39: As per the other PR, default this on the server, don't send it across from the client

server/game/game.js
```diff
@@ -623,6 +623,35 @@ class Game extends EventEmitter {
         this.continue();
     }
 
+    reInitialisePlayers(swap) {
+        let players = {};
+
+        _.each(this.playersAndSpectators, player => {
```
cryogen, line 629: underscore usage is deprecated, please convert this to use a for(foo of bar) loop

Issue #1170, opened 2020-03-17
Saurian House - Mass Mutation


romanogit:
Saurian 62 (61 Deus) / 43 New 69%
---
- [x] 222 - [High Priest Torvus](https://cdn.keyforgegame.com/media/card_front/en/479_222_CX663X678C56_en.png)
- [x] 226 - [Optio Gorkus](https://cdn.keyforgegame.com/media/card_front/en/479_226_6V5WH64JR74P_en.png)
- [x] 230 - [Scylla](https://cdn.keyforgegame.com/media/card_front/en/479_230_VHG8J2GPHF29_en.png)

---

- [x] 186 - [Citizen Shrix](https://cdn.keyforgegame.com/media/card_front/en/479_186_RGCC6M9G2QRX_en.png)
- [x] 188 - [Cornicen Octavia](https://cdn.keyforgegame.com/media/card_front/en/479_188_PPPRWFXCVWWG_en.png)
- [x] 189 - [Curse of Vanity](https://cdn.keyforgegame.com/media/card_front/en/479_189_66RHWHRQ8863_en.png)
- [x] 191 - [Defense Initiative](https://cdn.keyforgegame.com/media/card_front/en/479_191_93PJW4C6WP5F_en.png)
- [x] 194 - [Monument to Ludo](https://cdn.keyforgegame.com/media/card_front/en/479_194_GVVH5C3X2PJV_en.png)
- [x] 196 - [Sagittarii's Gaze](https://cdn.keyforgegame.com/media/card_front/en/479_196_86HWRJ97W8VG_en.png)
- [x] 200 - [Blast from the Past](https://cdn.keyforgegame.com/media/card_front/en/479_200_7PW589CCHXV3_en.png)
- [x] 202 - [Curia Saurus](https://cdn.keyforgegame.com/media/card_front/en/479_202_Q556C9VQGR8P_en.png)
- [x] 204 - [Dreadbone Decimus](https://cdn.keyforgegame.com/media/card_front/en/479_204_99C6R8FF22C2_en.png)
- [x] 206 - [Gladiodontus](https://cdn.keyforgegame.com/media/card_front/en/479_206_MHPJPV8W43Q5_en.png)
- [x] 211 - [Senator Quintina](https://cdn.keyforgegame.com/media/card_front/en/479_211_V43QR4XQXW2J_en.png)
[...]

PR #1245, opened 2020-04-03
MM: Saurian - Saurian Egg

ryanch:
This implements the Saurian Egg from this issue:
https://github.com/keyteki/keyteki/issues/1170

Includes:
* Test for Saurian Egg
* Implementation for Egg

Validation:
* Tests pass
* Lint passes

### DIFF
--- a/server/game/cards/04-MM/SaurianEgg.js
+++ b/server/game/cards/04-MM/SaurianEgg.js
@@ -1,11 +1,7 @@
 const Card = require('../../Card.js');
 
-
-
 class SaurianEgg extends Card {
-	
     setupCardAbilities(ability) {
-		
         this.persistentEffect({
             match: this,
             effect: ability.effects.cardCannot('reap')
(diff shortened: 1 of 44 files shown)


server/game/cards/04-MM/SaurianEgg.js
```diff
@@ -0,0 +1,54 @@
...
+                    action: ability.actions.sequential([
+                        ability.actions.putIntoPlay(),
+                        ability.actions.ready(),
+                        ability.actions.addPowerCounter({ amount: 3 })
+
+                    ])
+                })),
+
+                then: {
+                    gameAction: ability.actions.destroy(context => ({
```
jeremylarner, line 44: this should have `alwaysTriggers: true` no?

ryanch: No, the egg should only be destroyed if there are creatures put into play. 

**The card is here:**
https://decksofkeyforge.com/spoilers/1721066

**The text says:**
The Saurian Egg cannot fight or reap.
Omni: Discard the top 2 cards of your deck. **If you discard any Saurian creatures this way**, put them into play ready, give them three +1 power counters, and destroy Saurian Egg.

If I add "alwaysTrue" then the egg will get destroyed, and test cases fail:
* saurian-egg integration Saurian Egg's ability should discard 2 cards that are not saurian creatures and not be destroyed
* saurian-egg integration Saurian Egg's ability should discard only 1 cards that and not be destroyed, when only 1 non-saurian card in the deck.

Let me know if I am miss-understanding your comment, but I think no change is needed, unless:
A) I miss-understand the card.
B) I miss-understood your feedback.

jeremylarner: I agree that if the egg doesn't discard a saurian creature, the rest of the effect shouldn't trigger (so no `alwaysTriggers: true` in the first `then:`)
However, I'm concerned about the case where the egg discards a saurian creature (e.g. half of Deusillus) and that creature doesn't enter play.  In that case the egg should be destroyed, and I don't think your code will do that

romanogit: Or if it discards a Saurian alpha. Or if you use it and opponent has Ember Imp.

ryanch: Good catch guys. I will update, and add another test.

jeremylarner: Ember Imp is a red herring - it won't prevent creatures being put into play

ryanch: After chasing my tail for a bit I tried to make a test - but could not find a situation that would make the creatures not playable. So I can't test it, but I have updated the code to support this possibility.

cryogen:
@jeremylarner are your concerns on this addressed?

romanogit:
@cryogen : This PR may be closed. 
@ryanch : I used most of your code, just simplified the for loop for events which I found was not necessary.
And also, I added a complex logic for giants.

PR #1246, opened 2020-04-03
Docs: added saurian egg explanation

ryanch:
No code change, just documentation on how the saurian egg works.


docs/saurian_egg_explained.md
```diff
@@ -0,0 +1,273 @@
...
+If we wanted modify each element so that it is uppercased we could do:
+```
+newlist = [];
+for each str in list
+    add toUpperCase(str) to newList
+return newlist
+```
+A map allows you to provide a function to encapsulate the ```toUpperCase(str)``` and remove the other code. For example:
+```
+newList = list.mapp(toUpperCase(str) )
```
jeremylarner, line 194: map, not mapp

ryanch: Fixed in update.

drallieiv:
I think this PR should be rebased

Issue #1257, opened 2020-04-04
Sequential Actions Rules Clarifications


SkyJedi:
https://youtu.be/0NbXUv8TSr4?t=635

> You resolve the sentences of a card's text one at a time.

`ability.actions.sequential() ` is going to be a workhorse for the majority of these.

- [x] [Positron Bolt](https://decksofkeyforge.com/cards/positron-bolt)
- [x] [Word of Returning](https://decksofkeyforge.com/cards/word-of-returning)
- [x] [Hit and Run](https://decksofkeyforge.com/cards/hit-and-run)
- [x] [Destructive Analysis](https://decksofkeyforge.com/cards/destructive-analysis)
- [x] [Champions Challenge](https://decksofkeyforge.com/cards/champions-challenge)
- [x] [Tendrils of Pain](https://decksofkeyforge.com/cards/tendrils-of-pain)
- [x] [Key Hammer](https://decksofkeyforge.com/cards/key-hammer)
- [x] [Eureka!](https://decksofkeyforge.com/cards/eureka)
- [x] [Sloppy Labwork](https://decksofkeyforge.com/cards/sloppy-labwork)
- [x] [Tautau Vapors](https://decksofkeyforge.com/cards/tautau-vapors)
- [x] [Thorium Plasmate](https://decksofkeyforge.com/cards/thorium-plasmate)
- [x] [Timequake](https://decksofkeyforge.com/cards/timequake)
- [x] [Total Recall](https://decksofkeyforge.com/cards/total-recall)
- [x] [Cleansing Wave](https://decksofkeyforge.com/cards/cleansing-wave)
- [x] [Healing Blast](https://decksofkeyforge.com/cards/healing-blast)
- [x] [Smite](https://decksofkeyforge.com/cards/smite)
- [x] [Martyr's End](https://decksofkeyforge.com/cards/martyrs-end)
- [x] [Oath of Poverty](https://decksofkeyforge.com/cards/oath-of-poverty)
- [x] [Saury about that](https://decksofkeyforge.com/cards/saury-about-that)
- [x] [Mug](https://decksofkeyforge.com/cards/mug)
- [x] [One Last Job](https://decksofkeyforge.com/cards/one-last-job)
- [x] [They're Everywhere](https://decksofkeyforge.com/cards/theyre-everywhere)

SkyJedi:
All Done!

PR #1261, opened 2020-04-04
Sequential: Hit and Run

SkyJedi:
https://github.com/keyteki/keyteki/issues/1257
image


server/game/cards/03-WC/HitAndRun.js
```diff
@@ -8,17 +8,23 @@ class HitAndRun extends Card {
...
-                },
-                returnCreature: {
-                    cardType: 'creature',
-                    controller: 'self',
-                    cardCondition: () => true,
-                    gameAction: ability.actions.returnToHand()
                 }
-            }
+            },
+            then: ()=> ({
```
jeremylarner, line 13: Doesn't need to be an arrow function

PR #1271, opened 2020-04-05
Quintrino Flux bug fix

### LINKED ISSUE #1032: Quintrino Flux destroys warded creatures when chosen powers are equal
In [this log](https://www.thecrucibletracker.com/games/cea9f930-1c87-11ea-977d-3b5a45bd2ef0), on nedmike's turn 2, they play Quintrino Flux and choose a warded, exalted Centurion Stenopius (current power 6) on their side, and a Champion Anaphiel (power 6) on the other side.  The Centurion was destroyed, even though it was warded.  I'm pretty sure the destruction only happens once, and so the ward should be removed but the creature not destroyed.

image

Thanks!

jeremylarner:
Fixes #1032


cryogen:
You've got a test failure

PR #1274, opened 2020-04-06
MM: Saurian - Spoils of Battle

ryanch:
Implementation and tests for Spoils of Battle, per issue:
https://github.com/keyteki/keyteki/issues/1170

Spoils of Battle

Includes:
* New card Spoils of Battle
* New test case

Validation done:
* Test cases pass
* Lint passes

Notes:

This implementation is mostly a copy with small updates to [Pandemonium](https://github.com/keyteki/keyteki/blob/master/server/game/cards/01-Core/Pandemonium.js)


ryanch:
Updated to use the simpler capture impl that is in mainline, but not in mm. For now the tests in Spoils of Battle have been disabled.

Issue #1173, opened 2020-03-17
Untamed House - Mass Mutation


romanogit:
Untamed 61 (60 inc Niff) / 35 New 57%
---
- [x] 362 - [Cephaloist](https://cdn.keyforgegame.com/media/card_front/en/479_362_PXR679JWGGCR_en.png)
- [x] 364 - [Dark Faerie](https://cdn.keyforgegame.com/media/card_front/en/479_364_GR696M75FX2F_en.png)
- [x] 365 - [Fandangle](https://cdn.keyforgegame.com/media/card_front/en/479_365_4MGG3MJ4GVJ6_en.png)
- [x] 368 - [Gloriana's Attendant](https://cdn.keyforgegame.com/media/card_front/en/479_368_4R7RMFG3HCPP_en.png)
- [x] 372 - [Pismire](https://cdn.keyforgegame.com/media/card_front/en/479_372_RFRFQ3J846JX_en.png)
- [x] 373 - [Rapid Evolution](https://cdn.keyforgegame.com/media/card_front/en/479_373_HCPV29XGH25X_en.png)
- [x] 380 - [Commune](https://cdn.keyforgegame.com/media/card_front/en/479_380_VRV63277QR2W_en.png)
- [x] 381 - [Dark Harbinger](https://cdn.keyforgegame.com/media/card_front/en/479_381_FGHP825MPQG6_en.png)
- [x] 383 - [Growth Surge](https://cdn.keyforgegame.com/media/card_front/en/479_383_89G8P7PMGG8_en.png)
- [x] 385 - [Lumilu](https://cdn.keyforgegame.com/media/card_front/en/479_385_MV8M8QM2PRH2_en.png)
- [x] 389 - [Trimble](https://cdn.keyforgegame.com/media/card_front/en/479_389_62HXJ98MX482_en.png)
- [x] 391 - [Vault’s Blessing](https://cdn.keyforgegame.com/media/card_front/en/479_391_66QQ7HV563PV_en.png)
- [x] 394 - [Æmberlution](https://cdn.keyforgegame.com/media/card_front/en/479_394_85J25GC24WC2_en.png)
- [x] 395 - [Blossom Drake](https://cdn.keyforgegame.com/media/card_front/en/479_395_FFJ4XQRPR824_en.png)
- [x] 396 - [Chonkers](https://cdn.keyforgegame.com/media/card_front/en/479_396_86W72JX3Q4WH_en.png)
[...]

PR #1278, opened 2020-04-08
MM: Untamed 

SkyJedi:
https://decksofkeyforge.com/spoilers/1806958
https://decksofkeyforge.com/spoilers/1806975
https://decksofkeyforge.com/spoilers/1806976
https://decksofkeyforge.com/spoilers/1757707

https://github.com/keyteki/keyteki/issues/1173

### DIFF
--- a/server/game/cards/04-MM/DinoBeast.js
+++ b/server/game/cards/04-MM/DinoBeast.js
@@ -4,9 +4,7 @@ class DinoBeast extends Card {
     setupCardAbilities(ability) {
         this.play({
             optional: true,
-            gameAction: ability.actions.exalt(context => ({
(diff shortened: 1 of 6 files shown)


server/game/cards/04-MM/DinoBeast.js
```diff
@@ -0,0 +1,22 @@
+const Card = require('../../Card.js');
+
+class DinoBeast extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            optional: true,
+            gameAction: ability.actions.exalt(context => ({
```
romanogit, line 7: You can just use exalt() here as source is the default target.

PR #1286, opened 2020-04-09
Adaptive deckname buttons

SkyJedi:
props to bhawk for the ellipse hack, which i now have to figure out how to make client side
image


cryogen:
As discussed, I ported this to the client

PR #1288, opened 2020-04-10
Trim buttons on prompts so they don't overflow

cryogen:


SkyJedi:
https://github.com/keyteki/keyteki/pull/1286 

BHawk  gets props for this idea

PR #1310, opened 2020-04-18
Mm break key

romanogit:
Along with a rename of keyForged in player to keysForgedThisRound

image

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -629,6 +629,7 @@ class Player extends GameObject {
                         if(forgedKeyIndex !== -1) {
                             this.keysForgedThisRound.splice(forgedKeyIndex, 1);
                         }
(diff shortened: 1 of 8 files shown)


server/game/player.js
```diff
@@ -612,28 +612,36 @@ class Player extends GameObject {
         }
 
         this.keys[key] = true;
-        this.keyForged.push(key);
+        this.keysForgedThisRound.push(key);
         this.game.addMessage('{0} forges the {1}, paying {2} amber', this.game.activePlayer, `forgedkey${key}`, modifiedCost);
     }
 
     unforgeKey(choices) {
```
cryogen, line 619: This whole method is still confusing and wrong, but we'll fix it later

PR #1328, opened 2020-04-21
MM Deusillus, Ultra Graviton, Niffle Kong and It´s Coming

romanogit:
image

image

image

(Ultra Graviton is pending to resolve the bonus icons)

image


server/game/CompositeCard.js
```diff
@@ -0,0 +1,38 @@
...
+    moveTo(targetLocation) {
+        super.moveTo(targetLocation);
+
+        if(this.location === 'play area') {
+            this.compositeParts.forEach(id => {
+                // Always remove the second part from hand
+                let part = this.controller.getSourceList('hand').find(card => id === card.id);
+                this.controller.removeCardFromPile(part);
+                this.playedParts.push(part);
+            });
```
jeremylarner, line 22: I think this code should be in PutIntoPlayAction

server/game/CompositeCard.js
```diff
@@ -0,0 +1,38 @@
...
+
+            this.image = this.compositeImageId || this.id;
+        } else {
+            let cardPile = this.controller.getSourceList(this.location);
+            let cardIndex = cardPile.indexOf(this);
+            this.playedParts.forEach(part => {
+                part.location = this.location;
+                cardPile.splice(cardIndex, 0, part);
```
jeremylarner, line 30: Have you tested this?  As far as I can see, the card isn't in a pile when moveTo is called.
I can't see how you're removing the composite card from the pile either....

romanogit: Put this code in player

romanogit:
See #1361

PR #1329, opened 2020-04-21
moved the prompt over to the left side

SkyJedi:
image
moved the prompt over to the left side.  i dunno if i like it though.  it pushes the the play zone around every time the prompt changes.

Another option is this, which makes a gap on the left for the player row and battle lines
image


SkyJedi:
moving all the cards over to the right is wasting a lot of space.
allowing prompt to go over cards will block cards from users.

tabling this for now

PR #1331, opened 2020-04-21
fix for game-links and passwords

SkyJedi:
you can send game-link for passworded games now

### DIFF
--- a/client/Components/Games/GameLobby.jsx
+++ b/client/Components/Games/GameLobby.jsx
@@ -102,6 +102,19 @@ class GameLobby extends React.Component {
             $('#pendingGameModal').modal('show');
             this.setState({ gameState: GameState.PendingGame });
         }
+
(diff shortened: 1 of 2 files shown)


client/Components/Games/GameLobby.jsx
```diff
@@ -102,6 +102,19 @@ class GameLobby extends React.Component {
             $('#pendingGameModal').modal('show');
             this.setState({ gameState: GameState.PendingGame });
         }
+
+        if(!props.currentGame && this.props.gameId) {
+            const game = props.games.find(x => x.id === this.props.gameId);
```
cryogen, line 107: this should just be props.currentGame

SkyJedi: wouldn't they already be in a game and not need this then?

cryogen: Ah this is one level higher than i thought it was

PR #1334, opened 2020-04-23
feat: implement citizen shrix

granttitus:
```
Play/Reap: Exalt Citizen Shrix. Steal 1A.
```

image

Tests _should_ pass after keyteki-json-data is updated with this card.


romanogit:
Merged manually in https://github.com/keyteki/keyteki/pull/1484
Thanks for the cards implementation!

PR #1335, opened 2020-04-23
feat: implement cornist octavia

granttitus:
```
Action: Capture 2A.
```

image

Tests _should_ pass after keyteki-json-data is updated with this card.


romanogit:
Merged manually in https://github.com/keyteki/keyteki/pull/1484
Can be closed

PR #1350, opened 2020-04-24
Decklist zoom

SkyJedi:
image


cryogen:
My main concern with this is that it makes the already-ugly profile screen even more hideous.

SkyJedi:
This one one is just increasing only the deck in game.

cryogen:
Ahha! my bad

PR #1379, opened 2020-05-01
MM Saurian egg in master (merged and simplified a bit)

romanogit:
Supersedes #1245 

Added a gigantic logic


cryogen:
I don't like that this has so much gigantic specific code when it doesn't refer to gigantic on the card at all.  But it will do for now.

PR #1382, opened 2020-05-02
Missing play effect on Sacro Saurus

drallieiv:
Same Saurus effect then LycoSaurus


server/game/cards/04-MM/SacroSaurus.js
```diff
@@ -16,6 +16,16 @@ class SacroSaurus extends Card {
                 }
             }
         });
+        this.play({
```
romanogit, line 19: You have to remove the fight effect.
image

drallieiv: removed

drallieiv:
see #1382

PR #1437, opened 2020-05-11
Fix creature played as upgrade with "taunt" keyword

### LINKED ISSUE #1371: Mass Mutation - Securi-Droid graphical error
My opponent attached Securi-Druid as an upgraded to their Logo creature to give it taunt. The creature received taunt, yet graphically the card was shown attached to one of my creatures and one of my artifacts.
SecuriDroid

romanogit:
Fixes #1371

### DIFF
--- a/client/Components/GameBoard/Card.jsx
+++ b/client/Components/GameBoard/Card.jsx
@@ -258,7 +258,7 @@ class InnerCard extends React.Component {
             'unselectable': this.props.card.unselectable,
             'dragging': this.props.isDragging,
             'controlled': this.props.card.controlled,
-            'taunt': this.props.card.type === 'creature' && this.props.card.taunt && this.props.source === 'play area'
+            'taunt': this.props.card.taunt && this.props.source === 'play area'
         });
         let imageClass = classNames('card-image vertical', this.sizeClass, {
             'exhausted': (this.props.orientation === 'exhausted' || this.props.card.exhausted || this.props.orientation === 'horizontal')
--- a/client/Components/GameBoard/Card.jsx
+++ b/client/Components/GameBoard/Card.jsx
@@ -258,7 +258,7 @@ class InnerCard extends React.Component {
             'unselectable': this.props.card.unselectable,
             'dragging': this.props.isDragging,
             'controlled': this.props.card.controlled,
-            'taunt': this.props.card.taunt && this.props.source === 'play area'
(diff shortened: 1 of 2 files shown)


client/Components/GameBoard/Card.jsx
```diff
@@ -258,7 +258,7 @@ class InnerCard extends React.Component {
             'unselectable': this.props.card.unselectable,
             'dragging': this.props.isDragging,
             'controlled': this.props.card.controlled,
-            'taunt': this.props.card.taunt && this.props.source === 'play area'
+            'taunt': this.props.card.type === 'creature' && this.props.card.taunt && this.props.source === 'play area'
```
cryogen, line 261: This is adding game logic to the client, which we shouldn't do. Why is the engine applying the taunt property to an artefact? Am I missing something?

romanogit: @cryogen : it was not applying taunt to an artifact. It was applying to a creature played as an upgrade. So, the creature was played as an upgrade.

A good point though, is that I did not check if "type" of a creature played as upgrade in the GUI changed to "upgrade".

cryogen: Ok, so that still seems like an engine bug to me, why does it have the taunt flag set on the state message if it can't have taunt applied to it from a game perspective?

romanogit:
I think the fact that the creature is played as upgrade does not nullify
the creature only keywords, such as taunt, elusive, etc.

client/Components/GameBoard/Card.jsx
jeremylarner: It's because the card does have the taunt keyword.  I can see why you would say that the client shouldn't have game logic, but I don't think this case is actually clear cut.  The game state being sent to the client is correct (the card does have the taunt keyword), and the desired outcome is that the client doesn't offset cards with taunt that aren't creatures.  Should that be picked up by the game engine (non-creatures with taunt shouldn't be marked as having it in the game state), or by the client (only offset creatures with taunt, not all cards with it)?

cryogen: Definitely the former, it's not valid game state. It doesn't even have to be an engine change, it's a change to the getSummary function at most

romanogit: Update `getSummary`

PR #1440, opened 2020-05-12
Misc cleanups/fixes

cryogen:
Remove outdated community page
Update how to play page
Change to new captcha key
add debug logging for reset password
add circular ref detection to node messages to find source of node crashing bug


romanogit:
Probably, we will need to fix the localization files as well. I think that even though the text is in the page, it loads from localization files.

PR #1465, opened 2020-05-14
Minimize issue with Overrun and Lifeweb.

### LINKED ISSUE #1443: Overrun: Cannot read property 'length' of undefined
Sentry Issue: [KEYTEKI-16J](https://sentry.io/organizations/throneteki/issues/1661812616/?referrer=github_integration)

```
TypeError: Cannot read property 'length' of undefined
  File "/Users/keyteki/keyteki-node1/server/game/cards/03-WC/Overrun.js", line 11, in Object.condition
    condition: context => context.player.opponent && this.creaturesDestroyed[context.player.opponent.uuid].length >= 3,
  File "/Users/keyteki/keyteki-node1/server/game/triggeredability.js", line 83, in TriggeredAbility.isTriggeredByEvent
    if(this.properties.condition && !this.properties.condition(context)) {
  File "/Users/keyteki/keyteki-node1/server/game/triggeredability.js", line 66, in TriggeredAbility.eventHandler
    if(this.isTriggeredByEvent(event, context) && this.meetsRequirements(context) === '') {
  File "/Users/keyteki/keyteki-node1/server/game/triggeredability.js", line 125, in Game.handler
    handler: (event, window) => this.eventHandler(event, window)
  File "/Users/keyteki/keyteki-node1/server/game/gamesteps/forcedtriggeredabilitywindow.js", line 239, in null.<anonymous>
    this.game.emit(event.name + ':' + this.abilityType, event, this);
...
(5 additional frame(s) were not displayed)
```

romanogit:
If array is undefined, length would cause an exception. Storing a number allow, at least, to return false in the inequality
Fixes #1443
Fixes KEYTEKI-16J
Fixes KEYTEKI-HK
Fixes KEYTEKI-FN
Fixes KEYTEKI-SH


cryogen:
The reason I haven't merged this yet (and sorry I should have commented earlier) is that I'm not convinced this just doesn't hide the problem. Having just relooked at the sentry issue, I see the problem. If they are the first card played in the game, they will raise this error. The proper fix for this is to initialise the object for both players at the same time

romanogit:
> I see the problem. If they are the first card played in the game, they will raise this error

I added a test case specific for that and it was not the reason. 

>  The proper fix for this is to initialise the object for both players at the same time

The objects are indeed initialized for both players. onPhaseStarted event is raised for both players, that's what i saw in my tests, but i will recheck.

romanogit:
Hummm, I think our test cases are a bit deceiving because it does a full round of player1 and player2 before actually starting the test.

romanogit:
I figured out how to run the "setup" phase test and i was able to recreate the issue and my PR does fix it :)

cryogen:
I know it “fixes” it, that’s not at issue. It’s just the wrong fix. It hides the real issue and is just a workaround. I’ve already given you the correct fix, which your new test should confirm

romanogit:
Initialized both player and opponent

PR #1468, opened 2020-05-15
removed key cost from forge key action to simplify log

### LINKED ISSUE #1467: Imperial forge reporting negative amber cost in log
image

SkyJedi:
Fixes #1467


romanogit:
I like the cost in the log, to be honest

cryogen:
Agreed, can we keep the cost and fix the negative cost display issue?

PR #1498, opened 2020-05-20
Fix MM card bugs

jeremylarner:


server/game/cards/04-MM/MadProphetGizelhart.js
```diff
@@ -8,7 +8,7 @@ class MadProphetGizelhart extends Card {
             effect: ability.effects.gainAbility('action', {
                 effect: 'heal 1 damage from all creatures',
```
romanogit, line 9: Should fix this `heal 1 damage` here.

PR #1521, opened 2020-05-27
Auto-leg fix for applying +5 power every action

### LINKED ISSUE #1519: Using Auto-Legionary's Action while a creature increases it's power +5
When [Auto-Legionary] is a creature, you can use it's action to increase it's power by 5.  This is incorrect; it should reset the power to exactly 5 if it has been changed somehow.

SkyJedi:
Fixes #1519 

we should get a future effect that sets a cards power to X and use it here


cryogen:
Yes we really need a SetPowerAction for this.  This will do for now

PR #1559, opened 2020-06-05
hiding the api key

### LINKED ISSUE #1558: make Challonge Api key a password field
Streamers be blasting api keys out by accident

SkyJedi:
Fixes #1558


cryogen:
Think I preferred the first way you did it to be honest!

SkyJedi:
It was easier, I think it would send password managers into a hissy fit though

cryogen:
I've taken care of this in the client refactor

PR #1572, opened 2020-06-11
MM last two cards Commander Dhrxgar and J.O.N. Cargo

romanogit:
image

image


drallieiv:
Cargo seems made to handle weird "multiple houses" cases that could happen, along with all empty decks / empty hands.

For commander, we may want to check interaction with collar of subordination. Either when the commander gets stolen from the opponent, or when an opponent creature is stolen by collar, and becomes neighbor of commander.

Collar

romanogit:
@drallieiv 

I think while cards are in hand or deck, they will probably never have multiple houses. Nevertheless, Card's hasHouse method is ready for that.

Regarding Commander, good point about Collar. I do not know how it resolves.

PR #1574, opened 2020-06-11
Mm Untamed "Mutation *" and Vault's Blessing

romanogit:
image

image

image

image


drallieiv:
Does "untilNextTurn" handles the case where the creature switch controls ?

Player A plays Mutation on creature
Player B steals the creature
Player A starts his next turn < Effect should drop.

cryogen:
@drallieiv the engine handles all of that, you don't need to worry about it in the card itself

PR #1575, opened 2020-06-12
MM Mindfire and Lady Loreena

romanogit:
image

image

Fixed also a glitch with Nogi Smartfist

### DIFF
--- a/server/game/GameActions/RandomDiscardAction.js
+++ b/server/game/GameActions/RandomDiscardAction.js
@@ -23,19 +23,19 @@ class RandomDiscardAction extends PlayerAction {
     }
 
     getEvent(player, context) {
-        let amount = Math.min(this.amount, player.hand.length);
-        if (this.location === 'archives') {
-            amount = Math.min(this.amount, player.archives.length);
-        }
+        return super.createEvent('unnamedEvent', { player, context }, (event) => {
+            let amount = Math.min(this.amount, player.hand.length);
+            if (this.location === 'archives') {
+                amount = Math.min(this.amount, player.archives.length);
+            }
 
-        let cards = _.shuffle(player.hand).slice(0, amount);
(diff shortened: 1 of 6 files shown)


server/game/GameActions/RandomDiscardAction.js
```diff
@@ -23,17 +23,17 @@ class RandomDiscardAction extends PlayerAction {
...
+        }
 
-            let cards = _.shuffle(player.hand).slice(0, amount);
-            if (this.location === 'archives') {
-                cards = _.shuffle(player.archives).slice(0, amount);
-            }
+        let cards = _.shuffle(player.hand).slice(0, amount);
+        if (this.location === 'archives') {
+            cards = _.shuffle(player.archives).slice(0, amount);
+        }
```
jeremylarner, line 35: These changes are significant.  You're determining which cards (and how many cards) are discarded when the event is created, rather than when it resolves.  This could potentially cause problems (if a card is randomly selected to be discarded, but then is discarded by another effect in the interim), or it could be correct (if we ever got a card which said 'when this card would be discarded from your hand, do X', setting the event like this is probably how we would implement it).

For what you need here, I'm not sure this change is necessary: ```
return super.createEvent('unnamedEvent', { player, context }, event => {
    event.cards = cards```
achieves the same thing

romanogit: Fixed. Thanks!

romanogit:
@drallieiv , it is a tricky method. It sets the amount to hand and revert to archives if the location is archive. Maybe it would be better to have an if and an else for clarity, but the code was already in place, I just changed what was needed for my card.

drallieiv:
@romanogit I was tired and missed that there were 2 scoped amount (local one, and the object one).
So I don't think the bug happens like I thought initially

PR #1581, opened 2020-06-13
Update FR translations

drallieiv:
Added FR translations for new texts, and fixed a few old.

Note that typography rules in french require a space before double ponctuation symbols ( ! ? : )
Which does not exist in english typography, where ? is just after the last letter, like the normal dot.


cryogen:
Thanks!

PR #1583, opened 2020-06-13
Add some dis mm cards

cryogen:
and other misc fixes

### DIFF
--- a/server/game/cards/04-MM/MarkOfDis.js
+++ b/server/game/cards/04-MM/MarkOfDis.js
@@ -11,7 +11,8 @@ class MarkOfDis extends Card {
                         condition: () => context.target.location === 'play area',
                         trueGameAction: ability.actions.lastingEffect(() => ({
                             duration: 'untilNextTurn',
-                            roundDuration: 2,
+                            roundDuration: 3,
+                            targetController: context.target.controller,
                             effect: ability.effects.restrictHouseChoice(context.target.getHouses())
                         }))
                     }))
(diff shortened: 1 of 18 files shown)


server/game/cards/04-MM/MarkOfDis.js
```diff
@@ -0,0 +1,28 @@
...
+class MarkOfDis extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
+                cardType: 'creature',
+                gameAction: ability.actions.sequential([
+                    ability.actions.dealDamage({ amount: 2 }),
+                    ability.actions.conditional((context) => ({
+                        condition: () => context.target.location === 'play area',
+                        trueGameAction: ability.actions.lastingEffect(() => ({
```
jeremylarner, line 12: should be `ability.actions.untilEndOfMyNextTurn`

cryogen:
@jeremylarner I fixed Etan's jar in commit 
02169bb but I'm not sure it's right. Can you have a look?

server/game/cards/04-MM/MarkOfDis.js
```diff
@@ -0,0 +1,26 @@
...
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
+                cardType: 'creature',
+                gameAction: ability.actions.sequential([
+                    ability.actions.dealDamage({ amount: 2 }),
+                    ability.actions.conditional((context) => ({
+                        condition: () => context.target.location === 'play area',
+                        trueGameAction: ability.actions.untilEndOfMyNextTurn({
+                            targetController: context.target.controller,
```
jeremylarner, line 13: This is wrong I think - targetController should be 'any', 'current' or 'opponent', shouldn't it?

PR #1599, opened 2020-06-17
Add some saurian mm cards

cryogen:
@jeremylarner Can you have a look at what I'm doing wrong with Legion's March please?

### DIFF
--- a/server/game/cards/04-MM/LegionsMarch.js
+++ b/server/game/cards/04-MM/LegionsMarch.js
@@ -1,0 +1,29 @@
+const Card = require('../../Card.js');
+
+class LegionsMarch extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            effect:
+                'deal 1 damage to each non-Dinosaur creature after using a Dinosaur creature for the remainder of the turn',
+            gameAction: ability.actions.forRemainderOfTurn((context) => ({
+                when: {
+                    onUseCard: (event) => {
+                        return event.player === context.player && event.card.hasTrait('dinosaur');
+                    }
+                },
+                gameAction: ability.actions.dealDamage((damageContext) => {
+                    return {
+                        amount: 1,
+                        target: damageContext.game.creaturesInPlay.filter(
+                            (c) => !c.hasTrait('dinosaur')
+                        )
+                    };
+                })
+            }))
+        });
+    }
+}
+
+LegionsMarch.id = 'legion-s-march';
+
+module.exports = LegionsMarch;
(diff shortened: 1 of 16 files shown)


server/game/cards/04-MM/SirenHorn.js
```diff
@@ -0,0 +1,24 @@
...
+    setupCardAbilities(ability) {
+        this.whileAttached({
+            effect: ability.effects.gainAbility('beforeFight', {
+                effect: 'move 1 amber from {0} to {1}',
+                effectArgs: (context) => context.event.card,
+                gameAction: [
+                    ability.actions.removeAmber((context) => ({
+                        target: context.source
+                    })),
+                    ability.actions.placeAmber((context) => ({
```
romanogit, line 13: I think this needs to be a `then` clause, otherwise you will place an amber on the fought creatured even if Siren Horn's parent has no amber.

test/server/cards/04-MM/SirenHorn.spec.js
```diff
@@ -0,0 +1,28 @@
...
+                        hand: ['siren-horn'],
+                        inPlay: ['senator-shrix']
+                    },
+                    player2: {
+                        inPlay: ['little-rapscal', 'troll', 'nexus']
+                    }
+                });
+            });
+
+            it('should make the fight target gain 1 amber and the upgraded creature lose 1 amber', function () {
```
romanogit, line 17: Could be useful to add a test case when Senator Shrix has no amber on it.

cryogen: Good idea!

server/game/cards/04-MM/SirenHorn.js
cryogen: Good catch, I've added a condition which should do the same thing

cryogen: Though thinking about it, that wouldn't cover the case where something prevents the amber moving rather than there just not being any..

server/game/cards/04-MM/LegionsMarch.js
```diff
@@ -0,0 +1,29 @@
...
+
+class LegionsMarch extends Card {
+    setupCardAbilities(ability) {
+        this.play({
+            effect:
+                'deal 1 damage to each non-Dinosaur creature after using a Dinosaur creature for the remainder of the turn',
+            gameAction: ability.actions.forRemainderOfTurn((context) => ({
+                when: {
+                    onUseCard: (event) => {
+                        return event.player === context.player && event.card.hasTrait('dinosaur');
```
jeremylarner, line 11: onUseCard is a card action, so the event has no `player` property.  I think you want `event.context.player === context.player`

Issue #1169, opened 2020-03-17
Sanctum House - Mass Mutation


romanogit:
Sanctum 57 / 33 New 58%
---
- [x] 126 - [Ardent Hero](https://cdn.keyforgegame.com/media/card_front/en/479_126_8654CP76HG8_en.png)

---

- [x] 128 - [Burning Glare](https://cdn.keyforgegame.com/media/card_front/en/479_128_8V9FCMPGMQ5C_en.png)
- [x] 129 - [Champion Anaphiel](https://cdn.keyforgegame.com/media/card_front/en/479_129_FR9V84W6X65C_en.png)
- [x] 132 - [Dæmo-Knight](https://cdn.keyforgegame.com/media/card_front/en/479_132_CV53WG3V8J78_en.png)
- [x] 135 - [General Xalvador](https://cdn.keyforgegame.com/media/card_front/en/479_135_F278PG3H957W_en.png)
- [x] 137 - [Lieutenant Gorvenal](https://cdn.keyforgegame.com/media/card_front/en/479_137_CP2G72CH5MGP_en.png)
- [x] 142 - [Æmberheart](https://cdn.keyforgegame.com/media/card_front/en/479_142_QVWRG3QMW2J4_en.png)
- [x] 143 - [Angry Mob](https://cdn.keyforgegame.com/media/card_front/en/479_143_VQJJCXPW6JWP_en.png)
- [x] 144 - [Baldric the Bold](https://cdn.keyforgegame.com/media/card_front/en/479_144_8W3R4977V5F4_en.png)
- [x] 147 - [Bring Low](https://cdn.keyforgegame.com/media/card_front/en/479_147_V3M3MQ3QXH43_en.png)
- [x] 150 - [Gizelhart’s Standard](https://cdn.keyforgegame.com/media/card_front/en/479_150_RCXWVVJ9W8QC_en.png)
- [x] 153 - [Purify](https://cdn.keyforgegame.com/media/card_front/en/479_153_6RRCF57PMCCG_en.png)
- [x] 159 - [Book of Malefaction](https://cdn.keyforgegame.com/media/card_front/en/479_159_HR3CMHC4P55R_en.png)
- [x] 160 - [Call to Action](https://cdn.keyforgegame.com/media/card_front/en/479_160_48VHJC6345XJ_en.png)
- [x] 165 - [Lady Loreena](https://cdn.keyforgegame.com/media/card_front/en/479_165_WM5FCHMM98VC_en.png)
[...]

PR #1613, opened 2020-06-22
Implement mm sanctum cards

Toeofdoom:
Implemented Bring Low, Purify and Book of Malefaction for issue #1169 :
image
image
image


server/game/cards/04-MM/BookOfMalefaction.js
```diff
@@ -0,0 +1,28 @@
+const Card = require('../../Card.js');
+
+class BookOfMalefaction extends Card {
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onStealAmber: (event, context) => event.player === context.player
+            },
+            gameAction: ability.actions.addWarrantCounter((context) => ({
+                target: this,
```
jeremylarner, line 10: Try and avoid using `this` in game actions - should be `context.source` here

PR #1615, opened 2020-06-22
Implement MM Xeno-* cards

romanogit:
image

image

image

image

image


server/game/cards/04-MM/XenoBot.js
```diff
@@ -13,7 +13,25 @@ class XenoBot extends Card {
             }
         });
 
-        // TODO Add fight action copy of umbra alien
```
cryogen, line 16: Thanks for putting this in, I'd forgotten

PR #1616, opened 2020-06-22
MM Shadows Borrow, Opportunist and Mastermindy

romanogit:
image

image

image


drallieiv:
about scheme counters which seems to be a new thing

will the chat log and ui automatically handle a message stating that a "scheme counter" was added.

If new text added, it should be added to translation files too. At least the en reference.

romanogit:
I honestly let the message be handled by the engine.
We do not translate game messages.

PR #1632, opened 2020-06-25
MM Gluttony, Greed and Sloth

romanogit:
image

image

image


test/server/cards/04-MM/Gluttony.spec.js
```diff
@@ -0,0 +1,40 @@
+fdescribe('Guttony', function () {
```
Toeofdoom, line 1: should probably remove the f before describe

PR #1637, opened 2020-06-28
Replaced the fallback/sealed Archon cardback to be TCO branded

SteVader:
New TCO branded Archon cardback was suggested here https://discordapp.com/channels/474540818577752064/484154914038087680/726231758789541979


SkyJedi:
I love this

PR #1644, opened 2020-07-06
Implemented Auto-vac 5150.

Toeofdoom:
Implemented Auto-Vac 5150 for #1168 

image

Also updated lash of broken dreams unit test - I think it tested whether a key was forged, but the player could have had a "forge a key" prompt open.


Toeofdoom:
I have noticed that this will not interact correctly with keyfrog - it will increase the cost immediately, even though the effect as written should only apply from the start of the next turn. This applies to several other key cost increase effects too. I think Lucky Dice had to deal with that problem too, so that could be used.

romanogit:
Interesting observation.

romanogit:
(sorry, misclicked)

PR #1743, opened 2020-07-16
Fixes fetchdata.js hung on async/await call

romanogit:


romanogit:
Still not good... not waiting for fetching images

romanogit:
Now it is working :)

romanogit:
Still not working, damn!

PR #1775, opened 2020-07-17
minor korean translation

wonkh11:
minor korean translation for missing one


cryogen:
Thank you so much for these translations and your others!

PR #1799, opened 2020-07-19
Using card.uuid as prompt identifier, instead of aggregating quantities.

### LINKED ISSUE #1791: Vandalize doesn't distinguish between identically named cards when one is enhanced
I was playing a game against a friend of mine when I played Vandalize. I drew two copies of Securi-Droid and another card, whose name isn't important. The list of cards was presented to me like this:

Securi-Droid (2)
Other card

According to my playing partner, one of his four copies of Securi-Droid is enhanced, and since the two that were on the board were _not_ enhanced, one of the two drawn by Vandalized should have been, yet they weren't presented as separate cards.

My opponent's deck was Bao, District Thegn.

https://www.keyforgegame.com/deck-details/76d60cc8-73ea-4351-b34b-8030ae19f65b
https://decksofkeyforge.com/decks/76d60cc8-73ea-4351-b34b-8030ae19f65b

romanogit:
This feature used to work for mavericks, but with Enhancements, it become more complex to aggregate, so it is better to have duplicate buttons.

Fixes #1791 

Screenshots bellow shows a prompt with 3 buttons:
- Exchange Officer in "native" house SA
- Exchange Officer maverick Shadows with Enhancements 
- Exchange Officer maverick Shadows without Enhancements

image

image

image

image

I think a good enhancement would be:
1) Put a house icon next to the card name when it is maverick
2) Write a blue text when the card has Enhancement

What do you think, @cryogen ?

### DIFF
--- a/public/locales/de.json
+++ b/public/locales/de.json
@@ -381,7 +381,6 @@
     "privacy.lobby": "Your lobby messages are stored so that we can display them to other users of the site and to detect patterns of abusive behaviour.  Your messages may also be moderated to remove offensive content, or deleted altogether at our discretion.",
     "Create": "Create",
     "{{card}}": "{{card}}",
-    "{{card}} ({{quantity}})": "{{card}} ({{quantity}})",
     "How much amber do you want to use from {{card}}?": "How much amber do you want to use from {{card}}?",
     "Do you wish to repeat this effect?": "Do you wish to repeat this effect?",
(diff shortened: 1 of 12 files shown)


romanogit:
I think my idea above would be very complicated :)

cryogen:
I'll merge this but we should definitely make it clearer which card is which button in a future revision

PR #1823, opened 2020-07-21
Simplified Lieutenant Governal logic

### LINKED ISSUE #1822: Lieutenant Gorvenal capture amber when opponent fought with a controlled creature
My opponent exiled Gloriana Attendant.
I fought with Gloriana and my opponent's Lieutenant Governor captured 1 amber

```
Lakmear: 6 Æmber  (0 keys) romanomagacho22: 1 Æmber  (0 keys)
Turn 6
Key phase - romanomagacho22
romanomagacho22 does not forge a key. They have 1 amber. The current cost is 6 Æmber 
House phase - romanomagacho22
romanomagacho22 chooses sanctum  as their active house this turn
Main phase - romanomagacho22
romanomagacho22 plays Smite
romanomagacho22 uses Smite to ready and fight with Gloriana's Attendant
romanomagacho22 uses Gloriana's Attendant to make Gloriana's Attendant fight Baldric the Bold
romanomagacho22 uses Gloriana's Attendant to place 1 Æmber  on Baldric the Bold
Lakmear uses Lieutenant Gorvenal to capture 1 Æmber  from their opponent , placing it on Lieutenant Gorvenal <---
romanomagacho22 uses Lyco-Knight to reap with Lyco-Knight
```

romanogit:
Fixes #1822


jeremylarner:
Are you sure this fixes the issue? It doesn't look like a great change to make otherwise...

romanogit:
It does not look good. I will review

romanogit:
@jeremylarner : the fix is right. The onFight event does not carry the player (maybe it should). Anyway, fight only occurs during active player turn and the context.source is Lieutenant.

I added a test case and it fails with the previous version.

jeremylarner:
@romanogit I think the problem here is that the ability should be looking at a pre-fight snapshot of the attacker.  I'll PR an alternative fix: #1889

romanogit:
Thanks @jeremylarner

PR #1844, opened 2020-07-22
Fix Auto-Legionary applying effect to all creatures

### LINKED ISSUE #1843: Upon activating Auto-Legionary i was able to use all of my in play ready cards for the remainder of the game.
This issue happened multiple times, however, I don't know how to take screenshots on my fiancee's computer sorry. Should be easy enought to replicate as it happens everytime since the MM update.

romanogit:
Fixes #1843


server/game/cards/03-WC/AutoLegionary.js
```diff
@@ -4,7 +4,7 @@ class AutoLegionary extends Card {
     setupCardAbilities(ability) {
         let effects = [
             ability.effects.changeType('creature'),
-            ability.effects.canUse((card, context) => card === context.source)
+            ability.effects.canUse((card) => card === this)
```
jeremylarner, line 7: Shouldn't this just be `ability.effects.canUse()`?

romanogit: It seems it needs a function

romanogit:
I will split this fix in two

PR #1847, opened 2020-07-22
Sinder should only allow selecting a creature

drallieiv:
Sinder
Reap: Destroy a friendly creature.

Currently allows selecting an artifact (or upgrade to be destroyed)


drallieiv:
Bug occurence before fix

Sinder

PR #1854, opened 2020-07-24
Fix Auto-Legionary and Mimic Gel

romanogit:
Fix Auto-Legionary and Mimic Gel by setting a base power effect.


romanogit:
Mimic Gel still cannot be used as belonging to active house

PR #1857, opened 2020-07-25
added await curvetext and fixed mav icons

SkyJedi:
bumped down the image a quality a smidge

### DIFF
--- a/client/archonMaker.js
+++ b/client/archonMaker.js
@@ -139,7 +139,7 @@ export const buildDeckList = async (deck, language, translate, allCards) => {
 
     let name;
     try {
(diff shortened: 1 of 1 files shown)


client/archonMaker.js
```diff
@@ -210,7 +210,7 @@ export const buildDeckList = async (deck, language, translate, allCards) => {
             y = y + 44;
         }
 
-        const rarity = new fabric.Image(
+        const rarity = await new fabric.Image(
```
cryogen, line 213: Are you sure this function is async, or returns a promise?

SkyJedi: You're right., it's a synchronous function.  Ill remove these

PR #1866, opened 2020-07-27
fixes error when opponent has no amber to capture

SkyJedi:


server/game/cards/04-MM/Ant110ny.js
```diff
@@ -4,7 +4,7 @@ class Ant110ny extends Card {
     setupCardAbilities(ability) {
         this.play({
             gameAction: ability.actions.capture((context) => ({
-                amount: context.player.opponent.amber
+                amount: context.player.opponent.amber ? context.player.opponent.amber : 0
```
cryogen, line 7: The error is probably when there is no opponent, ie a single player game

SkyJedi:
Yes, but ERROR

PR #1872, opened 2020-07-28
WTF, they update images

### LINKED ISSUE #1871: Enhancements overlap card text on MM upgrades
image

It seems that MM versions of upgrade cards have an additional border to the left of the card text to prevent this. i.e.

**Worlds Collide 161**
image

**Mass Mutation 098**
image

SkyJedi:
Fixes #1871

I sort the card DB in newest -> oldest so TCO will grab the newest images


cryogen:
This will only fix new installs but that's fine :)

PR #1878, opened 2020-07-28
Fixes Keyfrog message stating incorrect player forges the key

romanogit:
Also fixed a messaging issue because a player can forge with keyfrog out of turn


test/server/cards/04-MM/Keyfrog.spec.js
```diff
@@ -65,5 +69,96 @@ describe('Keyfrog', function () {
 
             expect(this.keyfrog.location).toBe('discard');
         });
+
+        it("should end the game when it is the last key and controller's turn", function () {
```
Godzil, line 74: Should the test name be "should end the game when it is the last key and **not** controller's turn"?

romanogit: Fixed

PR #1884, opened 2020-07-29
Keyword tip: A small feature on CardZoom

rsdevigo:
Create a small div aboveethe CardZoom with the rules of keywords. Some cards have no explanation of the keywords.
Here a example:
keyword


SkyJedi:
Can this be toggles off?  Any player that has experience will not want to see this

rsdevigo:
> Can this be toggles off? Any player that has experience will not want to see this

I can do this in profile or in bottom bar.

cryogen:
I'm not sure I like the look of this. It kind of gets in the way. It isn't the job of the site to educate the player on the rules of the game, that's the rulebook's job. A certain amount of game knowledge is assumed

rsdevigo:
Certainly some knowledge must be assumed, but a beginner, usually he learns the order of the shift and the keywords he assumes that the cards will inform, some do not. Perhaps this feature could only be found in the Beginner game rooms.

PR #1885, opened 2020-07-30
Creating a new type of card size, a mixed one

rsdevigo:
Creating a new type of card size, a mixed one, where the hand is 1.8 the normal size of the card and the rest is the normal size.

Here a example:

image

### DIFF
--- a/client/Components/GameBoard/Card.jsx
+++ b/client/Components/GameBoard/Card.jsx
@@ -149,6 +149,8 @@ class InnerCard extends React.Component {
                 return 1.4;
             case 'x-large':
                 return 2;
+            case 'mixed':
+                return 1.8;
         }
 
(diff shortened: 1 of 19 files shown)


cryogen:
What problem is this trying to solve?

I've never had someone come to me and ask for a bigger hand and smaller cards. 99.9% of people asking for changes want the other way around if anything

rsdevigo:
Well, when I played, because of the resolution of my screen, I had to use the game in normal size, but what bothered me was the size of the hand, because I couldn't see it completely, even though I only had space to enlarge it. But it was an attempt ...
But in addition to this change in hand size, this PR also has a hover effect on the cards, which increases a little bit just to give visual feedback to the player which card he is viewing.

PR #1891, opened 2020-08-01
Fix 2x Mimic Gel error

### LINKED ISSUE #1035: 2x Mimic Gel - Error
thumbnail_image1

Apologies for closing my TCO window before gathering all information.

Issue occured when trying to play a second Mimic Gel on my side.
First Mimic Gel was played to copy an opponents Grom-Nine Toes.
Second Mimic Gel was played a few turns later when first Mimic Gel had 2 damage on it, but received error and the second Mimic Gel was sent to discard pile.
We turned on manual mode and I tried to drag Mimic Gel from discard pile onto battlefield, but it didn't work. Then I tried to putting it to my hand which worked. Turned off manual mode and tried to play Mimic Gel again, but received same error and Mimic Gel was sent to the discard pile.

Unfortunately I left the game so I could not test to remove the first Mimic Gel to see if that was the problem.

Deck name is "Rhomforce, the Ruffian Roboticist"
https://www.keyforgegame.com/deck-details/6a3704a6-ced1-42db-ac96-fbd67d5a728d

romanogit:
Fixes #1035

### DIFF
--- a/server/game/Effects/CopyCard.js
+++ b/server/game/Effects/CopyCard.js
@@ -1,5 +1,3 @@
-const _ = require('underscore');
-
 const EffectValue = require('./EffectValue');
 const GainAbility = require('./GainAbility');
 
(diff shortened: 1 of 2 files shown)


jeremylarner:
The change you've made in the logic here looks like it should break when copying a Mimic Gel which has gained actions from somewhere else - could you add a test for that?

romanogit:
I added a test for that.

Em sáb, 1 de ago de 2020 3:50 AM, jeremylarner <notifications@github.com>
escreveu:

PR #1901, opened 2020-08-08
August event banner

SteVader:


SteVader:
Forgot to add the image file, I guess that's why the checks failed first time.

client/pages/Lobby.jsx
```diff
@@ -106,6 +106,11 @@ class Lobby extends React.Component {
                         <div className='main-header' />
                     </Col>
                 </div>
+                <div>
+                    <Col sm={{ span: 10, offset: 1 }}>
+                        <div className='event-banner' />
```
cryogen, line 111: If you put a <Link href=... /> around this it will make it clickable

cryogen: or inside of it, not sure which

PR #1906, opened 2020-08-12
Added more distinct headers to the start of each round in message log.

SteVader:
Hopefully makes reading back through game log much easier as the start and end of each turn is much easier to see:

image

**NOTE:** The `startofround` div doesn't currently appear for turn 1 of player 1. This would make more sense if that was added.

### DIFF
--- a/client/Components/GameBoard/Messages.jsx
+++ b/client/Components/GameBoard/Messages.jsx
@@ -65,7 +65,6 @@ const Messages = ({ messages, onCardMouseOver, onCardMouseOut }) => {
                 switch (fragment.type) {
                     case 'endofround':
                     case 'phasestart':
-                    case 'startofround':
                         messages.push(
                             <div
                                 className={'font-weight-bold text-white separator ' + fragment.type}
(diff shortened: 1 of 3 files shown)


romanogit:
Can we make slightly different colors for each player?

SteVader:
Great suggestion. That's actually closer to what I had in mind when I started this, but I wasn't sure I'd be able to achieve it.

I've based this off the current releases theme configuration, so that it will automatically fall into place with subsequent releases e.g.

```css
background-image: url('~assets/img/panel-primary.png');
border-top: 1px solid theme-color('primary');
```

Perhaps I could tint the background of each phase message slightly. Something like this mockup:

image

It definitely emphasises each player's turn, making things even more readable. There's a lot of padding and spacing that I'd have to combat in between each row of the messages though.

cryogen:
I like that. Can we make it more subtle and put it in a separate change?

SkyJedi:
With the overall turn header could we remove the user name from each phase
header?

cryogen:
Once we get the background I'd be inclined to do that yeah

SteVader:
> I like that. Can we make it more subtle and put it in a separate change?

More subtle? I mean... this is almost my subtlest!

I of course jest. Leave it with me and I'll try to get a look at it soonest. When you say "separate change" though, do you mean keep this and branch off of it for the extra stuff? Or bin this and do it all as a new branch and PR?

cryogen:
As in I'll merge this and please do the rest on another branch/pr :)

PR #1917, opened 2020-08-16
Add username to password recovery email. Fixes #1916

### LINKED ISSUE #1916: User has no way to retrieve its username
1. Password reset can be done either with username or email address.
2. Login can only be done with username + password.
3. Password reset email does not provide the user's username.
4. Thus, it seems that there is no way for a user to recover its username for now.

tomsihap:
Fixes #1916.


romanogit:
Reworked in https://github.com/keyteki/keyteki/pull/2863
Thanks

PR #1931, opened 2020-08-26
Bandaid fix for unserializable ward message

Toeofdoom:
This seems to be what was breaking the server - the "damage source" for bonus damage icons has a reference to "game", so it breaks if you try to serialize it in a message within the game.


cryogen:
I've fixed this slightly more correctly in #1962. I've stolen your test though, thanks :)

PR #1954, opened 2020-09-09
Fixes Harvest Time and One Last Job not considering warded creatures

### LINKED ISSUE #1934: Harvest Time awards amber to controller even if it targets a creature that was warded
If one of the cards to be purged by harvest time is warded, and so doesn't get purged, it shouldn't count among cards that were "purged this way" and shouldn't award amber to its controller. Unfortunately, I didn't think to get the log from the game, but currently it incorrectly gives amber also for a warded card

romanogit:
Fixes #1934


romanogit:
This is fixed by https://github.com/keyteki/keyteki/issues/1895

PR #1978, opened 2020-09-18
Incremental messaging

cryogen:
Replace the current system of just sending the entire game state every update with one that uses a diff/patch system. The resulting network traffic is considerably reduced and is now flat throuought a game, rather than getting worse as time goes on


SkyJedi:
Woot woot woot

Issue #1907, opened 2020-08-13
Deck searching not working (only for me)


Ugluk4242:
I have not been able to search for decks for some weeks. Filtering by 'Expansion' it works well.

![Search_Deck_1](https://user-images.githubusercontent.com/68158046/90173159-06e67c00-dd72-11ea-9cf6-f9b37c7e32f0.JPG)
![Search_Deck_2](https://user-images.githubusercontent.com/68158046/90173160-06e67c00-dd72-11ea-9dd0-05e48f4ff272.JPG)

PR #1985, opened 2020-09-23
Fix deck name search to actually look at the name column

### LINKED ISSUE #1972: Deck search looks at id not name
and is also case sensitive

cryogen:
Also, make it case insensitive
Fixes #1972


cryogen:
Might also fix #1907

PR #1992, opened 2020-09-27
Key forged order fix

### LINKED ISSUE #1991: the key images are being forged all nimble-bimbly and its making us look like scrubs
image
Scrubs I say!

SkyJedi:
image
NO MORE SCRUBS


cryogen:
apparently I don't know how sort works :)

PR #1998, opened 2020-09-29
fabric canvas renoodle

SkyJedi:
move cardback and decklists away from dataUrls and into pure fabric generated canvases.


SkyJedi:
fixes #1593

PR #2010, opened 2020-10-05
Check is players are still available for rematch

### LINKED ISSUE #1921: One should probably not be able to accept a rematch after the player who asked for one has left the game
So this happened to me a couple of times. Once a few months ago and once this week. Sometimes an opponent may afk or tab away after a completed game and won't see that you asked for a rematch and therefore not answer to accept or reject the rematch. If this happens, I may just leave the game before getting a response, and join a new game. The problem comes if the other player comes back and accepts the rematch at a later point. In that case, you'll be booted out of the game you just started with another player and into a rematch with the original opponent. Just leaving in the middle of a game like that, involuntarily, feels disrespectful to the second opponent so it would be good if this possibility could be eliminated.

SkyJedi:
fixes #1921

### DIFF
--- a/server/game/gamesteps/RematchPrompt.js
+++ b/server/game/gamesteps/RematchPrompt.js
@@ -3,7 +3,6 @@ const AllPlayerPrompt = require('./allplayerprompt');
 class RematchPrompt extends AllPlayerPrompt {
     constructor(game, requestingPlayer) {
         super(game);
-
         this.requestingPlayer = requestingPlayer;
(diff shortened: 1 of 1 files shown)


cryogen:
This only half fixes #1921. There's a second part to check that a user isn't in a new unrelated game before yanking them into a rematch.

SkyJedi:
can you be in a new game if you're not marked as having left your previous game?

cryogen:
By normal means, no. But what's happening here is rematch is triggering and dragging someone through :)

PR #2059, opened 2020-11-11
Suppress vertical scrollbar on whole play area

SteVader:
Several components cause the play area to exceed 100% body height. This leads to an unsightly (and useless) scrollbar that sits next to the message log (which has it's own scrollbar). Suppress the outer scrollbar as it's never of any use.

before:

image

after:

image

### DIFF
--- a/client/Components/GameBoard/GameBoard.scss
+++ b/client/Components/GameBoard/GameBoard.scss
@@ -6,7 +6,7 @@
 
 .game-board {
     position: absolute;
-    bottom: 0;
+    bottom: 3px;
     left: 0;
     right: 0;
     top: $navbar-height;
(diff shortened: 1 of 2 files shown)


SteVader:
I might have missed a use case where that outer body scrollbar makes sense/is required, but I couldn't find one - even using super short browser windows.

cryogen:
Have you tried this on a tablet/phone?

Also what was causing the scrollbar? That usually only shows up when something is broken, or on a phone/tablet

SteVader:
I couldn't nail down exactly what's causing it. I had been convinced that it's the blue glow on the bottom of the home side player bar, as that's the amount it wants to scroll by:

yuck-scroll

I haven't tried on phone, but it's got exactly the same amount of scroll on iPad.

I'll take another look at "fixing" it rather than suppressing it later today. Admittedly I could have been a smidgen more sober than I was last night :)

SteVader:
Okay, this is 100% less of a sweep it under the carpet solution. See what you think.

PR #2078, opened 2020-11-25
Fixes Gigantic Card issues

### LINKED ISSUE #1789: Amberlution doesn't kill Ultra Gravitron and then Ultra Gravitron is duplicated
My board had Bot Bookton, Ultra Gravitron, and other creatures
Reap with Bot Bookton, Plays Amberlution 
Ultra Gravitron remains on board (I did NOT check graveyard at this point)
Manual Mode can not remove Ultra Gravitron from board, can't even select it
Next turn Play It's Coming 
Ultra Gravitron (fully assembled image is in graveyeard) along with top half image
Put top half in hand, now board had full gravitron, discard has full gravitron, hand has top half (quit game at this point)

image

### LINKED ISSUE #1811: Gigantic creatures behave differently depending on which half you select
Hello,

It seems there is an issue with gigantic creatures enhanced. 
Only the half part selected to play the whole creature triggers its bonus. 
Don’t know if the both part could be enhanced by the algorithm but in any case if in tco I don’t choose the enhanced half to play the creature the bonus icon are not resolved. 
It’s happen on my niffle Kong. Assume it’s the same for other gigantic.

### LINKED ISSUE #1898: Troop Call shouldn't return both halves of Niffle Kong (apparently?)
I'm not convinced this is a "bug," but I think the ruling is correct and can't find an argument against it. I stumbled upon the following thread in the FFG KeyForge forum:

https://community.fantasyflightgames.com/topic/310312-bouncing-gigantic-creatures/

I can't really argue with the logic. It claims that Troop Call should return only the bottom half of Niffle Kong when pulling cards from discard. (It will return both halves from play because they're treated as a single creature when in play, as spelled out below.)

**TROOP CALL**
**Play:** Return each friendly Niffle creature from your discard pile and from play to your hand.

**GIGANTIC (RULEBOOK, PAGE 12)**
A gigantic creature counts as 2 cards while out of play, but as a single card while in play. _[This is why Troop Call will return both halves from play.]_
...
Both halves of a gigantic creature have the same title, house, and card type. Otherwise, each half has the attributes printed on it: the top half has bonus icons, while the bottom half has power, armor, and the text box.

---

Ergo, both halves are creatures ("card type" is defined on page 5), but only the bottom half has the Niffle trait, since that's the half with the text box. (From page 16 of the rulebook: "Traits are descriptive attributes (such as “Knight” or “Specter”) that may be referenced by other cards. **Traits are listed at the top center of a card’s text box.** Traits have no inherent game effect, but may be referenced by card abilities.")

### LINKED ISSUE #1902: Purify puts Ultra Gravitron into play
Sorry, don't have screenshots but it will be easyto replicate.

Player A
Has half of UG in hand

Player B 
Plays purify, purges one of Player A's mutant creatures

Cards are discarded from top of Player A's deck until a non mutant creature is discarded.
UG, a non mutant creature is discarded and put into play (almost certainly because Player A has the other half in hand, I would not expect this bug to happen if that wasn't the case).

What should happen
As the card is played from the discard pile and not Player A's hand there is no opportunity for the two halves of UG to be combined into a single creature and played. As UG cannot be put into play the card should return to the discard pile. No more cards would be discarded by Purify as it's effect ends when a non mutant creature is discarded.

Please note, the bottom half of Niffle Kong and Deusillas are mutants so would simply be discarded by Purify. The top half of those two and both halves of U?G are non mutant creatures so would trigger what Purify is searching for and end the effect but cannot be put into play.

### LINKED ISSUE #1965: "Blast from the Past" deals damage even when it returns the top half of a gigantic creature
My opponent played Blast from the Past, chose the top half of Deusillus, and dealt 20 points of damage to my Niffle Kong. According to the rulebook, the top half of a gigantic creature does not have any power associated with it:

_**Ruleboook, p. 12, "Gigantic":** Both halves of a gigantic creature have the same title, house, and card type. Otherwise, each half has the attributes printed on it: the top half has bonus icons, while the bottom half has power, armor, and the text box._

This issue is effectively linked to https://github.com/keyteki/keyteki/issues/1898.

### LINKED ISSUE #1974: No shield on Ultra Gravitron
Shield 3 is missing on Ultra Gravitron.
Thanks, bye!

Crucible_bug

### LINKED ISSUE #1986: Æmberlution should remove the destroyed creatures before playing the new one
When [Æmberlution](https://decksofkeyforge.com/cards/mberlution) is played the board should be cleared before playing the creature from hand as it may change the way you put them on the board.

The board state before playing amberlution
BeforeAemberlution

Just after playing the card; 
DuringAmberlution

All the creatures that are destroyed are still visible and TCO ask to play the new one, this is a bit confusing.

After playing my own creatures, we can see the destroyed one still being there:
OpponentCreatures

Also not the same issue but it is for the same card it is not really clear from which hand the cards are being played:
Maybe changing the text to something "Which flank do you want to place this creature from your opponent hand on?"

### LINKED ISSUE #2079: Saurian Egg should be destroyed if half gigantic is discarded.
Since the card text is a comma separated list, if only half of a gigantic creature is discarded, the egg should be destroyed.

### LINKED ISSUE #2080: Playing Amberlution when Ultra Gravitron is in hand causes both halves to be put into play as full creatures.
Extracted from #1789

### LINKED ISSUE #2081: Amberlution does not allow order of put into play
Amberlution does not allow order of put into play

Similar to #300

romanogit:
Fixes all of the Gigantic Card issues

Fixes #1789
Fixes #1811
Fixes #1898
Fixes #1902
Fixes #1965
Fixes #1974
Fixes #1986
Fixes #2079 
Fixes #2080
Fixes #2081

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,449 +5,431 @@
     "requires": true,
     "dependencies": {
         "@babel/cli": {
-            "version": "7.12.8",
-            "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.12.8.tgz",
-            "integrity": "sha512-/6nQj11oaGhLmZiuRUfxsujiPDc9BBReemiXgIbxc+M5W+MIiFKYwvNDJvBfnGKNsJTKbUfEheKc9cwoPHAVQA==",
+            "version": "7.10.1",
+            "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.10.1.tgz",
+            "integrity": "sha512-cVB+dXeGhMOqViIaZs3A9OUAe4pKw4SBNdMw6yHJMYR7s4TB+Cei7ThquV/84O19PdIFWuwe03vxxES0BHUm5g==",
             "requires": {
-                "@nicolo-ribaudo/chokidar-2": "2.1.8-no-fsevents",
-                "chokidar": "^3.4.0",
+                "chokidar": "^2.1.8",
                 "commander": "^4.0.1",
                 "convert-source-map": "^1.1.0",
                 "fs-readdir-recursive": "^1.1.0",
                 "glob": "^7.0.0",
-                "lodash": "^4.17.19",
+                "lodash": "^4.17.13",
                 "make-dir": "^2.1.0",
                 "slash": "^2.0.0",
                 "source-map": "^0.5.0"
+            },
+            "dependencies": {
+                "commander": {
+                    "version": "4.1.1",
+                    "resolved": "https://registry.npmjs.org/commander/-/commander-4.1.1.tgz",
+                    "integrity": "sha512-NOKm8xhkzAjzFx8B2v5OAHT+u5pRQc2UCa2Vq9jYL/31o2wi9mxBA7LIFs3sV5VSC49z6pEhfbMULvShKj26WA=="
+                },
+                "slash": {
+                    "version": "2.0.0",
+                    "resolved": "https://registry.npmjs.org/slash/-/slash-2.0.0.tgz",
+                    "integrity": "sha512-ZYKh3Wh2z1PpEXWr0MpSBZ0V6mZHAQfYevttO11c51CaWjGTaadiKZ+wVt1PbMlDV5qhMFslpZCemhwOK7C89A=="
+                }
             }
         },
         "@babel/code-frame": {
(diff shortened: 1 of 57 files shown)


romanogit:
Done done done :)

PR #2083, opened 2020-11-26
Display own upgrades over the top of opponents battleline.

### LINKED ISSUE #2076: Suggestion: Can Upgrades be displayed over the top of the bottom of enemy creatures?
Would it be possible to have upgrades on friendly creatures displayed over the top of the bottom part of enemy creatures?  At the moment, it can be impossible to see all the upgrades on a friendly creature because they are hidden underneath the bottom of enemy creatures.

We can see the creature by hovering over it and seeing it enlarged over where the log is, if there's anything we need to read that we can't see because of the upgrades.

Just an idea.

image

image

SteVader:
Fixes #2076

Not sure if there's a reason we've avoided layering things in this order before now? 

Changes this:
image

To this:
image

### DIFF
--- a/client/Components/GameBoard/GameBoard.scss
+++ b/client/Components/GameBoard/GameBoard.scss
@@ -97,6 +97,7 @@
 
     &.our-side {
         margin: 0;
+        z-index: 20;
 
         .taunt {
             margin-top: -15px;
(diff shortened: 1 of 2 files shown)


client/Components/GameBoard/GameBoard.scss
```diff
@@ -97,6 +97,7 @@
 
     &.our-side {
         margin: 0;
+        z-index: 20;
```
cryogen, line 100: we have variables in _variables.scss for this :)

SteVader: Ah, okay - nice. I've made a change that hopefully makes sense now then.

PR #2097, opened 2020-12-02
Fixes Sir Marrows double capture and Po's Pixies interaction

### LINKED ISSUE #190: Multiple Sir Marrows each capture an amber instead of offering a choice
When 2 Sir Marrows are in play in opponent side, and the I reap, I could decide which take the aember, but both take 1 aember. I think it is a bug.
I think in real game, I decide which take and the other wouldnt.

### LINKED ISSUE #567: Sir Marrows should capture from common supply when Po's Pixies is in play
Sir Marrows should capture from common supply when Po's Pixies is in play.

I had a Po's Pixies and opponent had a Sir Marrows. He captured my amber after I reap and the amber came from my pool, not the common supply.

Sir Marrows text says: "After your opponent gains A by reaping, Sir Marrows captures it. " 

Po's Pixies says: A stolen or captured from your pool is taken from the common supply instead.

romanogit:
Fixes #190
Fixes #567

### DIFF
--- a/server/game/cards/02-AoA/SirMarrows.js
+++ b/server/game/cards/02-AoA/SirMarrows.js
@@ -9,7 +9,7 @@ class SirMarrows extends Card {
             },
             gameAction: ability.actions.sequential([
                 ability.actions.capture((context) => ({
-                    amount: context.event.amount || 0
+                    amount: context.event.amount
                 })),
                 ability.actions.changeEvent((context) => ({
                     event: context.event,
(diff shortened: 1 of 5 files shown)


server/game/cards/02-AoA/SirMarrows.js
```diff
@@ -4,17 +4,21 @@ class SirMarrows extends Card {
...
             gameAction: ability.actions.sequential([
-                ability.actions.placeAmber({
-                    amount: 1,
-                    target: this
-                }),
-                ability.actions.loseAmber((context) => ({
-                    target: context.player.opponent,
-                    amount: 1
+                ability.actions.capture((context) => ({
+                    amount: context.event.amount || 0
```
cryogen, line 12: Why the || 0? Can this ever be undefined?

romanogit: I don't think it can, just being over careful, I will remove it.

PR #2100, opened 2020-12-02
Fixes Epic Quest after used by Nexus - bug in conditional action for PlayerAction

### LINKED ISSUE #2098: Epic Quest forging for opponent
image

So, this happened. I used Epic Quest and it forged a key for my opponent, winning him the game. Maybe something to do with the fact he used Nexus two turns before in order to prevent me from winning?

romanogit:
Fixes #2098

### DIFF
--- a/server/game/cards/03-WC/Memolith.js
+++ b/server/game/cards/03-WC/Memolith.js
@@ -11,7 +11,7 @@ class Memolith extends Card {
                     card.location === 'hand' ||
                     (card.parent === context.source && card.location === 'grafted'),
                 gameAction: ability.actions.conditional({
-                    condition: (context) => context.target && context.target.location === 'hand',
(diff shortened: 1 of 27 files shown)


server/game/cards/03-WC/Memolith.js
```diff
@@ -11,7 +11,7 @@ class Memolith extends Card {
                     card.location === 'hand' ||
                     (card.parent === context.source && card.location === 'grafted'),
                 gameAction: ability.actions.conditional({
-                    condition: (context) => context.target.location === 'hand',
+                    condition: (context) => context.target && context.target.location === 'hand',
```
cryogen, line 14: target should never be undefined here, if it is, I'd rather it blow up because it's a bug in the engine

PR #2131, opened 2020-12-09
Fix for two Custom Offices and message to pay amber

### LINKED ISSUE #1493: Custom Office and Tentacus do not have a game message saying opponent gained amber
Customs Office and Tentacus does not have a game message saying opponent gained amber.

It caused confusion: see #1479 and #1490

image

image

### LINKED ISSUE #2023: Two Customs Offices
I was playing a game where I had two customs offices in play. My opponent had 1 aember and was able to play a niffle grounds. I received one aember when they played it, but I should have received two. I think it was okay that they played the niffle grounds because it had a pip on it which would have brought them up to two, but I only received the one. 

This was the game if it is helpful: https://www.thecrucibletracker.com/games/5c7df600-0fdd-11eb-b0bc-11e54f3d3599

romanogit:
Fixes #1493
Fixes #2023

### DIFF
--- a/server/game/baseability.js
+++ b/server/game/baseability.js
@@ -136,7 +136,8 @@ class BaseAbility {
      */
     canPayCosts(context) {
         let cost = this.cost.concat(context.player.getAdditionalCosts(context));
-        return cost.every((cost) => cost.canPay(context));
+        let aggregatedCost = {};
+        return cost.every((cost) => cost.canPay(context, aggregatedCost));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


server/game/baseability.js
```diff
@@ -136,7 +136,8 @@ class BaseAbility {
      */
     canPayCosts(context) {
         let cost = this.cost.concat(context.player.getAdditionalCosts(context));
-        return cost.every((cost) => cost.canPay(context));
+        let aggregatedCost = {};
```
cryogen, line 139: This is a bit confusing. Can there be anything other than amber on this? If not just make it be amber. If so, at least initialise amber to 0 in here

romanogit: In theory, canPay is agnostic. Technically, it could be amber, chain, key (see costs.js)

PR #2134, opened 2020-12-09
Toggle button text during confirm action (Delete/Cancel)

### LINKED ISSUE #1445: When you delete a deck the options are unclear
On the decks screen, when you click to delete a deck, you get two options: Delete and Confirm.  Delete would be the obvious choice here, but Delete cancels the action.

SteVader:
Fixes #1445

I appreciate this might tread on the toes of New World(tm) - so feel free to bin this PR if so. That aside, I decided to finally _learn_ some React (rather than just keep staring it in the hope of guessing how it works). 

Is this a completely cack-handed way to tackle this?

### DIFF
--- a/client/Components/Form/ConfirmButton.jsx
+++ b/client/Components/Form/ConfirmButton.jsx
@@ -19,10 +19,11 @@ const ConfirmButton = ({ children, onClick }) => {
     const handleClick = (event) => {
         event.preventDefault();
         setShowConfirm(!showConfirm);
-        if (btnText === 'Cancel') {
+
+        if (showConfirm) {
             setBtnText(children);
         } else {
-            setBtnText('Cancel');
+            setBtnText(<Trans>Cancel</Trans>);
         }
     };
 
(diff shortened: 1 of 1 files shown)


client/Components/Form/ConfirmButton.jsx
```diff
@@ -14,23 +14,29 @@ import { Button } from 'react-bootstrap';
...
     const [showConfirm, setShowConfirm] = useState(false);
+    const [btnText, setBtnText] = useState(children);
 
     const handleClick = (event) => {
         event.preventDefault();
         setShowConfirm(!showConfirm);
+        if (btnText === 'Cancel') {
+            setBtnText(children);
+        } else {
+            setBtnText('Cancel');
```
cryogen, line 25: You'll need to pull in a translation object and pass this to it, so it's translated

```
const { t } = useTranslation();

setBtnText(t('Cancel'));
```

SteVader: I couldn't get this to work, but the translation object is already imported for the confirm button, so I did the following:

```jsx
if (showConfirm) {
    setBtnText(children);
} else {
    setBtnText(<Trans>Cancel</Trans>);
}
```

Which seems to work properly. Is that better?

PR #2142, opened 2020-12-14
Fixes discarded cards not counted for alpha check

### LINKED ISSUE #2139: Swindle not checking if actions have been taken before it is played
My opponent discarded a card and then played Swindle, but as Swindle is an Alpha (and Omega) card, one shouldn't be able to take any actions (e.g. discarding a card) before (or after) playing it.

image

romanogit:
Fixes #2139


romanogit:
This one is important to merge as a newly introduced after the alpha check for actions in phase, not round.

PR #2143, opened 2020-12-14
Fixes Finch Cloak message

### LINKED ISSUE #2141: Finch Cloak
Is it possible for the log to be more specific about what Finch Cloak has done please?

image

romanogit:
Fixes #2141

### DIFF
--- a/server/game/cards/03-WC/FinchCloak.js
+++ b/server/game/cards/03-WC/FinchCloak.js
@@ -23,7 +23,7 @@ class FinchCloak extends Card {
             effectArgs: (context) =>
                 context.player.amber < context.player.opponent.amber
                     ? 'steal 1 amber'
-                    : 'gain both players 1 amber'
+                    : 'make both players gain 1 amber'
         });
     }
 }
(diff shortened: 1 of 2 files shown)


server/game/cards/03-WC/FinchCloak.js
```diff
@@ -18,7 +18,12 @@ class FinchCloak extends Card {
                         amount: 1
                     }))
                 ])
-            }))
+            })),
+            effect: '{1}',
+            effectArgs: (context) =>
+                context.player.amber < context.player.opponent.amber
+                    ? 'steal 1 amber'
+                    : 'gain both players 1 amber'
```
cryogen, line 26: What does this message resolve to? Seems like it's missing a "make"

romanogit: player1 uses Finch Cloak to gain both players 1 amber

Looks like Shakesperean english?

I can write 'to make both players gain 1 amber'

cryogen: Yep that works better :)

PR #2177, opened 2021-01-05
Fixes Mimic Gel and Cyber-Clone copying Gigantic (top part played first) 

### LINKED ISSUE #2176: Mimic Gel is not copying gigantic (probably when played through top part)
Mimic Gel is not copying gigantic (probably when played through top part)

romanogit:
Fixes #2176


romanogit:
Reverting changes not related to Gigantic

PR #2215, opened 2021-01-19
Tireless Crocag and Harland Mindlock test case

romanogit:


romanogit:
It is more complex than this change... closing to think further

PR #2237, opened 2021-01-25
Fixes "after next card is played" bugs

### LINKED ISSUE #2137: Murkens and Soft Landing / Wild Bounty
Soft Landing readies Murkens and Wild Bounty resolves icons from enhanced Murkens

image

image

### LINKED ISSUE #2227: Action that causes playing encounter suit triggers it
The scenario that was observed:
Exhume targeting Murkens
Murkens plays card from top of deck
Top card is Encounter Suit
Encounter Suit played on Archimedes
Archimedes gets warded by Encounter Suit

Archimedes should not be warded, because Encounter Suit says before the action is resolved, which is a point that has passed.

I suspect a Wild Wormhole playing Encounter Suit will have the same result but I have not tested.

### LINKED ISSUE #2235: When resolving bonus icons “an additional time”, do the additional time immediately after resolving that icon
When resolving bonus icons “an additional time”, do the additional time immediately after resolving that icon

romanogit:
Fixes #2235
Fixes #2227
Fixes #2137


server/game/GameActions/ResolveBonusIconsAction.js
```diff
@@ -7,13 +7,96 @@ class ResolveBonusIconsAction extends CardGameAction {
         this.effectMsg = "resolve {0}'s bonus icons";
     }
 
+    resolveIcon(card, context, event, icon) {
+        if (icon === 'amber') {
```
cryogen, line 11: This feels like it should be a switch

romanogit: Fixed to use switch/case

PR #2283, opened 2021-03-07
Added several untested generated cards

Toeofdoom:
Contains autogenerated code for 39 new cards - this code has not been tested, I'm opening the PR to provide a clear status/avoid duplicating work. Most code should work fine as it fits into existing patterns, however unusual combinations might not be exactly correct. e.g. Rocketeer tryska readies neighbors as they come into play - the automatic implementation is might not set this up in a way supported by the engine if the entersPlayReady() ability could be checked before the "neighbor" condition makes any sense.

Also introduces SimpleEventTracker (intended to make it easier to count events that have occurred during a turn.) - worth having someone with more knowledge of the engine confirm if it will work as intended.


romanogit:
Closing...

PR #2292, opened 2021-03-12
Fixes Imperial Traitor not revealing when hand has no sanctum

### LINKED ISSUE #1840: Imperial Traitor doesn't reveal hand if there are no sanctum cards in it
Main phase - blinkingline
blinkingline plays Imperial Traitor
blinkingline gains an Æmber  due to Imperial Traitor's bonus icon
blinkingline uses Imperial Traitor

Opponent's hand was not revealed.  Even if I can't purge a Sanctum card i should still know what is in the hand.

romanogit:
Applied the same selectorCondition created for dark tidings which makes cards selectable, but Done is not enabled.
Fixes #1840


romanogit:
selectorCondition was deprecated before becoming a reality so need to find another solution

PR #2296, opened 2021-03-15
Fixes Bracchus and bypassing amber prompt

### LINKED ISSUE #2206: Able to forge "bypassing" creature's ambers and paying only 2 ambers !!!
Here is the setup:

- I have 2 amber
- Senator Bracchus has 4 amber
- Liutenant Governal has 2 amber
- Champion Anaphiel has 2 amber

Senator Bracchus allows me to use amber on any creature to forge.

Prompt is in the following order:
- Bracchus: 0 1 2 3 4

I select ZERO

- Governal 0 1 2

I select ZERO

- Anaphiel 0 1 2

I select ZERO

I forge paying 2.

I simulate in my local environment (using different creatures):

image

romanogit:
Queue  forge key prompt sequentially, instead of all at once
Fixes #2206

### DIFF
--- a/server/game/player.js
+++ b/server/game/player.js
@@ -621,6 +621,11 @@ class Player extends GameObject {
             let max = Math.min(modifiedCost, sourceAmber);
             let min = Math.max(0, modifiedCost - this.amber - totalAvailable + sourceAmber);
             if (max === min) {
+                this.game.addMessage(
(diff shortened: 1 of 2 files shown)


romanogit:
This is important fix that is nice to go with DT

PR #2301, opened 2021-03-16
Fixes Impspector message when no cards are purged

### LINKED ISSUE #2260: Impspector's message when there is nothing to purge
I just killed my opponent's Impspector with an empty hand and thus, no card got purged. That worked fine but the message in the log could be improved.

grafik

romanogit:
Fixes #2260


romanogit:
Merging some easy ones...

PR #2310, opened 2021-03-18
Ghosthawk should be optional

### LINKED ISSUE #2306: Ghosthawk isn't optional
The ability says: "Play: You may reap with each neighboring creature, one at a time."

But currently it's not possible to play it and *not* reap with the neighboring creatures (you'd want that if there's a CÆNDLE unit in play for example).

romanogit:
Fixes #2306


romanogit:
Merging some easy ones...

PR #2322, opened 2021-03-20
implemented Infighting and test

ryanch:


server/game/cards/05-DT/Infighting.js
```diff
@@ -0,0 +1,24 @@
+const Card = require('../../Card.js');
+
+class Infighting extends Card {
+    //Play: Each creature deals damage equal to its power to its right neighbor.
+    setupCardAbilities(ability) {
+        this.play({
+            gameAction: ability.actions.dealDamage((context) => ({
+                // iterate over all creatures in the game, each creature takes damage from the creature on its left.
+                target: context.game.creaturesInPlay.filter(
```
romanogit, line 9: I think it can be simplified to:
target: (context.opponent ? context.opponent.creaturesInPlay.slice(1).map(c => c.neighbors[0]) : []).concat(context.players.creaturesInPlay.slice(1).map(c => c.neighbors[0]))

ryanch: I was thinking you might have feedback on this one. Took me a while to come up with that I used there. I tried pasting in your code there for the targets - it did not work - it was giving me an undefined error. But if you can help me understand what you are proposing I can try to re-work it.

It looks like your code is merging the list of left neighbor creatures. My code is looking for creatures who have a neighbor to the left, and only returning the creatures who have left neighbors (not a list of left neighbors). 

Consider these creatures:
A B C D

I want a list of who needs to be damaged by the creature to their left:
B C D

Not a list of creatures who will be doing the damaging:
A B C

romanogit: so it is even simplier:

target: (context.opponent ? context.opponent.creaturesInPlay.slice(1) : []).concat(context.players.creaturesInPlay.slice(1))

ryanch: Yes, very much simpler to just combine the two lists of creatures from both sides without the first character. I updated the PR with this simpler logic, and added 2 tests - just to make sure the logic didn't break when there was no creatures. Thanks for pointing this out.

PR #2328, opened 2021-03-20
implemented primal relic and test

ryanch:


romanogit:
I see uncharted land does the same with this, so, it is ok

PR #2330, opened 2021-03-21
created onAmberSpentOnForge event, and implemented Mookling

ryanch:
In this PR, I did the following:
1) Added onAmberSpentOnForge event (and improved the messaging)
2) updated The Sting to use this new event
3) Created Mookling and Mookling Evil Twin cards with tests

### DIFF
--- a/server/game/cards/01-Core/TheSting.js
+++ b/server/game/cards/01-Core/TheSting.js
@@ -6,13 +6,6 @@ class TheSting extends Card {
             effect: ability.effects.skipStep('key')
         });
 
-        /*
-        this.persistentEffect({
-            targetController: 'opponent',
-            effect: ability.effects.forgeAmberGainedByOpponent()
-        });
-        */
-
         this.reaction({
             when: {
                 onAmberSpentOnForge: (event, context) => event.player === context.player.opponent
--- a/server/game/cards/01-Core/TheSting.js
+++ b/server/game/cards/01-Core/TheSting.js
@@ -17,8 +17,8 @@ class TheSting extends Card {
             when: {
                 onAmberSpentOnForge: (event, context) => event.player === context.player.opponent
             },
-            gameAction: ability.actions.gainAmber((event) => ({
-                amount: event.cost
+            gameAction: ability.actions.gainAmber((context) => ({
+                amount: context.event.amberSpent
             }))
         });
 
--- a/server/game/cards/01-Core/TheSting.js
+++ b/server/game/cards/01-Core/TheSting.js
@@ -6,9 +6,20 @@ class TheSting extends Card {
             effect: ability.effects.skipStep('key')
         });
 
+        /*
         this.persistentEffect({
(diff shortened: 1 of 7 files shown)


server/game/cards/05-DT/Mookling.js
```diff
@@ -0,0 +1,14 @@
+const Card = require('../../Card.js');
+
+class Mookling extends Card {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'opponent',
+            effect: ability.effects.modifyKeyCost(() => this.getPower(false))
```
romanogit, line 7: this is tricky, should avoid using "this"

server/game/cards/01-Core/TheSting.js
```diff
@@ -6,9 +6,13 @@ class TheSting extends Card {
             effect: ability.effects.skipStep('key')
         });
 
-        this.persistentEffect({
-            targetController: 'opponent',
-            effect: ability.effects.forgeAmberGainedByOpponent()
```
romanogit, line 11: Can you cleanup this effect "forgeAmberGainedByOpponent" from effect.js?

server/game/player.js
```diff
@@ -670,6 +670,7 @@ class Player extends GameObject {
 
     finalizeForge(key, modifiedCost, cost) {
         this.modifyAmber(-modifiedCost);
+        this.game.raiseEvent('onAmberSpentOnForge', { player: this, amberSpent: cost });
 
         if (this.anyEffect('forgeAmberGainedByOpponent')) {
```
romanogit, line 675: Can you remove this block, since "forgeAmberGainedByOpponent" is now deprecated

server/game/cards/05-DT/Mookling.js
ryanch: So, I tried to avoid using it, but it seems that "context.source" is undefined. Example, this fails:

`effect: ability.effects.modifyKeyCost((context) => context.source.getPower(false))`

I looked at other cards using persistentEffect and they also use "this" to find the card with the actual effect. Maybe A) I'm missing something, or B) there is a something up which is not assigning context.source to be the card with the effect?

server/game/cards/01-Core/TheSting.js
ryanch: Done. Tests passed as well.

server/game/player.js
ryanch: Done. Tests passed as well.

server/game/cards/05-DT/Mookling.js
ryanch: I'll keep Mookling around with it.

ryanch: Figured it out.. though I have no idea why this works:
`effect: ability.effects.modifyKeyCost((_, context) => context.source.getPower(false))`

PR #2337, opened 2021-03-21
DT: AnkylosaurusFormation, Saurarium, TheFirstScroll, AltruistsRostrum, MedicusLacus

SkyJedi:


server/game/cards/05-DT/AnkylosaurusFormation.js
```diff
@@ -0,0 +1,38 @@
+const Card = require('../../Card.js');
+
+class AnkylosaurusFormation extends Card {
+    //Play: Choose one:
+    //• A friendly creature gains skirmish for the rest of your turn.
+    //• Exalt a friendly creature. Each friendly creature gains skirmish for the rest of your turn.
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
```
romanogit, line 9: This card is a choose. Need to use mode select.

SkyJedi: i started on that, but then though this way would eliminate a UX step.  One creature I always going to be targeted to get skirmish, the choice is really do you wanna exalt it or not for the larger effect

Right now its Choose creature, exalt yes/no

Moving it to the mode select would be choose skirmish/ or exalt  then target a creature. ...i guess thats the same amount of clicking.  

I can switch it if ya want

romanogit: That's tricky 👍

server/game/cards/05-DT/Saurarium.js
```diff
@@ -0,0 +1,20 @@
+const Card = require('../../Card.js');
+
+class Saurarium extends Card {
+    //Each creature with the lowest power cannot reap.
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'any',
+            match: (card, context) => {
+                let lowestPower = context.game.creaturesInPlay.sort((a, b) => a.power - b.power)[0]
```
romanogit, line 9: if there are no creatures in play, this may cause an "undefined" exception (undefined.power)

PR #2345, opened 2021-03-23
implemented officers blaster, with leftNeighbor and rightNeighbor

ryanch:
In this PR I'm adding support for the "officers blaster", which is described here:
https://archonarcana.com/Spoilers?cardnumber=298

Two call outs.

**One:**
In the blaster impl I have this line:
```
                            upgrade: context.source.upgrades.find(
                                (upgrade) => upgrade.name === "Officer's Blaster"
                            )
```
I'm not super excited about this, but the context object is not a pointer to the upgrade, it is a pointer to the creature. So, to find the blaster to move it I'm doing this search.

**Two:**
I'm proposing adding a leftNeighbor and rightNeighbor. 

Consider this code copied from Card.js https://github.com/keyteki/keyteki/blob/master/server/game/Card.js#L973:
```
        if (index < 0) {
            return neighbors;
        } else if (index > 0) {
            neighbors.push(creatures[index - 1]);
        }

        if (index < creatures.length - 1) {
            neighbors.push(creatures[index + 1]);
        }
```
In the code above, if there is only two creatures on the board then the value of [0] is ambiguous. [0] can be the left creature or the right creature.

Consider this board, where target creature is the creature we are looking at:
```
['Other Creature', 'Target Creature'] 
target.neighbors[0] = Other Creature (the left creature)
```

Consider this board, where target creature is the creature we are looking at:
```
['Target Creature', 'Other Creature']
target.neighbors[0] = Other Creature (the right creature)
```


ryanch:
If you hate this idea, for neighbors, I can redo the card without it.

server/game/GameActions/AttachAction.js
```diff
@@ -44,7 +44,9 @@ class AttachAction extends CardGameAction {
             { card: this.upgrade, parent: card, context: context },
             (event) => {
                 if (event.card.location === 'play area') {
-                    event.card.parent.removeAttachment(event.card);
+                    if (event.card.parent) {
```
romanogit, line 47: I did not understand why this "if" was needed

server/game/cards/05-DT/OfficersBlaster.js
```diff
@@ -0,0 +1,31 @@
...
+                ability.effects.gainAbility('destroyed', {
+                    target: {
+                        cardType: 'creature',
+                        numCards: 1,
+                        // note, context.source here is the creature that was upgraded, not this upgrade
+                        cardCondition: (card, context) => card === context.source.rightNeighbor,
+                        gameAction: ability.actions.attach((context) => ({
+                            // note, the context here is from the context of the creature - not the upgrade
+                            // to avoid using "this" looking up the upgrade on the create
+                            upgrade: context.source.upgrades.find(
```
romanogit, line 19: I do not think we need a card selector here, since rightNeighbor is always one.
The search by name might cause issues with two Blasters on the same user, so, maybe, it is better to use "this" for upgrades.

server/game/GameActions/AttachAction.js
ryanch: I thought it was needed, but I just took it out and it all works now. I think this was a hold over before I figured out the the "this" problem I was having.

server/game/cards/05-DT/OfficersBlaster.js
ryanch: Removed

ryanch:
I've updated the PR, using "this" and no longer using getRightNeighbor

server/game/cards/05-DT/OfficersBlaster.js
```diff
@@ -0,0 +1,48 @@
...
+
+class OfficersBlaster extends Card {
+    //This creature gains +2 power.
+    //This creature gains, “Destroyed: Attach $this to this creature's right neighbor.”
+    setupCardAbilities(ability) {
+        this.whileAttached({
+            effect: [
+                ability.effects.modifyPower(2),
+                ability.effects.gainAbility('destroyed', {
+                    target: {
```
romanogit, line 11: there is not need for a target here. You can simply do:
gameAction: ability.actions.attach(() => ({
    target: context.source.controller.creaturesInPlay[context.source.controller.creaturesInPlay.indexOf(context.source) + 1]
    // note, using "this" here to get a pointer to the blaster
    upgrade: this
}))

romanogit:
I will merge, since it is working. I can review later the suggestion I gave you.

PR #2366, opened 2021-03-28
updated Selective Preservation to prompt for each power

ryanch:
This pr updates Selective Preservation to prompt for each unique power level, instead of all creatures at once.

Note, I added the ```effect: 'Select creatures for each power to not destroy.',``` so that I could have a no-op for the pre-then block.

### DIFF
--- a/server/game/cards/05-DT/SelectivePreservation.js
+++ b/server/game/cards/05-DT/SelectivePreservation.js
@@ -3,20 +3,44 @@ const Card = require('../../Card.js');
 class SelectivePreservation extends Card {
     setupCardAbilities(ability) {
         this.play({
-            target: {
-                activePromptTitle: 'Choose a creature of each power value to not destroy',
-                cardType: 'creature',
-                mode: 'exactly',
-                numCards: (context) =>
-                    new Set(context.game.creaturesInPlay.map((card) => card.power)).size,
-                selectorCondition: (selectedCards, context) =>
-                    new Set(context.game.creaturesInPlay.map((card) => card.power)).size ===
-                    new Set(selectedCards.map((card) => card.power)).size,
-                gameAction: ability.actions.destroy((context) => ({
-                    target: context.game.creaturesInPlay.filter(
-                        (card) => context.target && !context.target.includes(card)
+            effect: 'Select creatures for each power to not destroy.',
+
+            then: (preThenContext) => {
+                let unqiuePowers = Array.from(
+                    new Set(
+                        preThenContext.game.creaturesInPlay.map((creature) =>
+                            creature.getPower(false)
+                        )
                     )
-                }))
+                );
+                unqiuePowers.sort((a, b) => (a > b ? 1 : -1));
+
+                let targets = [];
+                for (let i = 0; i < unqiuePowers.length; i++) {
+                    let power = unqiuePowers[i];
+                    let targetKey = 'power' + power;
+                    targets[targetKey] = {
+                        activePromptTitle: {
+                            text: 'Choose a {{power}} creature to not destroy',
+                            values: { power: power }
+                        },
+                        cardType: 'creature',
+                        numCards: 1,
+                        cardCondition: (card) => card.getPower(false) === power,
+                        gameAction: ability.actions.destroy((context) => ({
+                            target: context.game.creaturesInPlay.filter(
+                                (card) =>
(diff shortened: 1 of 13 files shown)


server/game/cards/05-DT/SelectivePreservation.js
```diff
@@ -3,20 +3,44 @@ const Card = require('../../Card.js');
...
+                    let targetKey = 'power' + power;
+                    targets[targetKey] = {
+                        activePromptTitle: {
+                            text: 'Choose a {{power}} creature to not destroy',
+                            values: { power: power }
+                        },
+                        cardType: 'creature',
+                        numCards: 1,
+                        cardCondition: (card) => card.getPower(false) === power,
+                        gameAction: ability.actions.destroy((context) => ({
```
romanogit, line 30: I believe the gameAction needs to be outside all targets and you need to make sure the card is not in any target selection

server/game/cards/05-DT/SelectivePreservation.js
```diff
@@ -3,20 +3,44 @@ const Card = require('../../Card.js');
...
-                selectorCondition: (selectedCards, context) =>
-                    new Set(context.game.creaturesInPlay.map((card) => card.power)).size ===
-                    new Set(selectedCards.map((card) => card.power)).size,
-                gameAction: ability.actions.destroy((context) => ({
-                    target: context.game.creaturesInPlay.filter(
-                        (card) => context.target && !context.target.includes(card)
+            effect: 'Select creatures for each power to not destroy.',
+
+            then: (preThenContext) => {
+                let unqiuePowers = Array.from(
```
romanogit, line 9: I believe the preference in the code for unique is to use _.uniq

server/game/cards/05-DT/SelectivePreservation.js
```diff
@@ -3,20 +3,44 @@ const Card = require('../../Card.js');
...
+                    new Set(
+                        preThenContext.game.creaturesInPlay.map((creature) =>
+                            creature.getPower(false)
```
romanogit, line 12: you can use creature.power, instead of getPower(false).

server/game/cards/05-DT/SelectivePreservation.js
```diff
@@ -3,20 +3,44 @@ const Card = require('../../Card.js');
...
+                );
+                unqiuePowers.sort((a, b) => (a > b ? 1 : -1));
+
+                let targets = [];
+                for (let i = 0; i < unqiuePowers.length; i++) {
+                    let power = unqiuePowers[i];
+                    let targetKey = 'power' + power;
+                    targets[targetKey] = {
+                        activePromptTitle: {
+                            text: 'Choose a {{power}} creature to not destroy',
```
romanogit, line 24: You need to add this text to locale file

server/game/cards/05-DT/SelectivePreservation.js
```diff
@@ -3,20 +3,44 @@ const Card = require('../../Card.js');
...
-                mode: 'exactly',
-                numCards: (context) =>
-                    new Set(context.game.creaturesInPlay.map((card) => card.power)).size,
-                selectorCondition: (selectedCards, context) =>
-                    new Set(context.game.creaturesInPlay.map((card) => card.power)).size ===
-                    new Set(selectedCards.map((card) => card.power)).size,
-                gameAction: ability.actions.destroy((context) => ({
-                    target: context.game.creaturesInPlay.filter(
-                        (card) => context.target && !context.target.includes(card)
+            effect: 'Select creatures for each power to not destroy.',
```
romanogit, line 6: No need to add "effect" message to locale file

server/game/cards/05-DT/SelectivePreservation.js
cryogen: the preference is to never use underscore where possible :)

romanogit: ops, my bad! I have been using, so I'll try to avoid.

server/game/cards/05-DT/SelectivePreservation.js
romanogit: You could also use the ellipsis operator, I guess: [...new Set(a)]

romanogit:
I will try to merge and "fix" this one on a branch I am working because this selectorCondition is blocking me.

Issue #2184, opened 2021-01-11
Dark Tidings Spoilers 


romanogit:
Dark Tidings Spoilers (keeping all here while we do not have official names)
https://archonarcana.com/Spoilers?order_by=Number&reprints=New%20Cards

House: Logos
- [x] ([001](https://archonarcana.com/Spoilers?cardnumber=001)) Armadrone (Evil Twin) - This card is incomplete and subject to change.
- [x] ([002](https://archonarcana.com/Spoilers?cardnumber=002)) Binary Moray (Evil Twin) - This card is incomplete and subject to change.
- [x] ([003](https://archonarcana.com/Spoilers?cardnumber=003)) Captain Kresage (Evil Twin) - This card is incomplete and subject to change.
- [x] ([004](https://archonarcana.com/Spoilers?cardnumber=004)) Old Egad (Evil Twin) - This card is incomplete and subject to change.
- [x] ([005](https://archonarcana.com/Spoilers?cardnumber=005)) PI Sweven (Evil Twin)
- [x] ([006](https://archonarcana.com/Spoilers?cardnumber=006)) Professor Gwyne? (Evil Twin) - This card is incomplete and subject to change.
- [x] ([007](https://archonarcana.com/Spoilers?cardnumber=007)) Talmage Steelheart (Evil Twin) - This card is incomplete and subject to change.
- [x] ([008](https://archonarcana.com/Spoilers?cardnumber=008)) Doctor Verokter? - This card is incomplete and subject to change.
- [x] ([009](https://archonarcana.com/Spoilers?cardnumber=009)) Roof Laboratory? - This card is incomplete and subject to change.
- [x] ([010](https://archonarcana.com/Spoilers?cardnumber=010)) Reckless Experiment ? - This card is incomplete and subject to change.
- [x] ([011](https://archonarcana.com/Spoilers?cardnumber=011)) Groundbreaking Discovery? - This card is incomplete and subject to change.
- [x] ([015](https://archonarcana.com/Spoilers?cardnumber=015)) Backup Plan ? - This card is incomplete and subject to change.
- [x] ([016](https://archonarcana.com/Spoilers?cardnumber=016)) Captain Kresage
- [x] ([017](https://archonarcana.com/Spoilers?cardnumber=017)) Chronophage
- [x] ([018](https://archonarcana.com/Spoilers?cardnumber=018)) Honors Keysis
[...]

PR #2367, opened 2021-03-28
implement Redhand Registry

ryanch:
Card Description:
https://archonarcana.com/Spoilers?cardnumber=106&order_by=Number

Related issue:


server/game/cards/05-DT/RedhandRegistry.js
```diff
@@ -0,0 +1,29 @@
+const Card = require('../../Card.js');
+
+class RedhandRegistry extends Card {
+    //After A is stolen from you, your opponent skips the “forge a key” step during their next turn.
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onStealAmber: (event, context) => event.player === context.player
+            },
+            // this looks strange, but we need to create the skip forge event
```
romanogit, line 10: I think the problem is that you own the artifact, but it might trigger during your turn or during opponent's turn, so a lastingEffect with a targetComtroller opponent might no be effective always. I wonder if we should pass a "player" to it.

romanogit: Also, the delayed effect might vanish if the artifact is destroyed on the same turn, for example, due a Big Fuse

ryanch: Okay, so looking at this again, I found that there is a untilEndOfMyNextTurn which works perfect!

PR #2370, opened 2021-03-28
implement Drawn By The Depths and test

ryanch:
Card Description:
https://archonarcana.com/Spoilers?cardnumber=313&order_by=Number

Related Issue:

Note, to make this card, I updated MoveToBottomAction so that you can choose to impact the opponent's deck.

### DIFF
--- a/server/game/GameActions/MoveToBottomAction.js
+++ b/server/game/GameActions/MoveToBottomAction.js
@@ -1,6 +1,11 @@
 const CardGameAction = require('./CardGameAction');
 
 class MoveToBottomAction extends CardGameAction {
+    constructor(propertyFactory) {
+        super(propertyFactory);
+        this.targetPlayer = 'current';
+    }
+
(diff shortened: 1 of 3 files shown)


server/game/GameActions/MoveToBottomAction.js
```diff
@@ -1,6 +1,11 @@
 const CardGameAction = require('./CardGameAction');
 
 class MoveToBottomAction extends CardGameAction {
+    constructor(propertyFactory) {
+        super(propertyFactory);
+        this.targetPlayer = 'current';
```
romanogit, line 6: I think you should use a "player" attribute: See CaptureAction and UseAction. using "current" and "opponent" can be really tricky for example, if you the opponent triggers an effect, so, for consistenct, let's use player.

You could also assume that, since the cards are always coming from the deck, you could using card.owner instead of context.player to change the deck.

PR #2372, opened 2021-03-28
implemented Card 321 and tests

ryanch:
Card Description:
https://archonarcana.com/Spoilers?cardnumber=321&order_by=Number

Related Issue:

Note, for this card I added a new counter type 'melerukh' I named it this because the card text just says 'counter' but the creature title has 'melerukh' in it. Not sure if we should do something generic instead?


romanogit:
> Not sure if we should do something generic instead?

That's a good question. I think we are flooding the Gameactions with addXXXCounter for card-specific counters, such as doom, ignorance, knowledge, glory, etc... I think for card-specific counters we should use a generic addCounter action where we pass the type. I will elaborate on this. For now, it is good.

romanogit:
BTW, the card translation says the counter name is "awakening"

image

ryanch:
Updated to use 'awakening' for counter name. Thanks for catching that, I was using the Chinese translation.

PR #2380, opened 2021-03-31
Fixes Incubation Chamber log message

### LINKED ISSUE #2263: Minor issue - Incubation Chamber log message is incorrect when opponent uses it but does not archive a card
Minor log issue but the log says player uses Incubation Chamber to archive Incubation Chamber if they use the card but do not archive a card.

image

romanogit:
Fixes #2263


romanogit:
Merging some easy ones...

PR #2392, opened 2021-04-13
Blypyp effect should allow enter ready and not ready after in play

romanogit:
Consistent with Soft Landing


romanogit:
This is a nice fix that is nice to go with DT with the ready reactions

Issue #2427, opened 2021-04-16
Grand Alliance Council not working properly


Blaph:
Grand Alliance Council says "Play: Choose a creature from each house. Destroy each creature not chosen", but doesn't allow choosing if the creatures in play are the only ones of their house (In our case: floomf for me, rustmiser for my opponent).
![Bug here](https://user-images.githubusercontent.com/30389835/115024442-64face80-9ec0-11eb-94fe-d7a153e58d5a.png)

Chat: https://pastebin.com/hf5H7dRy

cmorenoserrano:
This happened to me as well just now. It costed me the game, but it's fine.

ryanch:
taking a look at this one

ryanch:
added test case to reproduce this. now working on fix.

ryanch:
I have a fix, working to get my env working again to make a PR for it.

Osarek:
seems to be resolved :D

PR #2485, opened 2021-04-18
fix for issue 2427

ryanch:
fixes the issue reported here: https://github.com/keyteki/keyteki/issues/2427

I fixed this by making it so that creatures can't be targeted for destruction if there is only 1 creature of that house in play, with this in the filter:

`context.game.creaturesInPlay.filter((card) => card.hasHouse(house)).length > 1`

### DIFF
--- a/server/game/cards/05-DT/GrandAllianceCouncil.js
+++ b/server/game/cards/05-DT/GrandAllianceCouncil.js
@@ -14,22 +14,17 @@ class GrandAllianceCouncil extends Card {
                 },
                 cardType: 'creature',
                 numCards: 1,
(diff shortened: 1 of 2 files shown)


romanogit:
@ryanch : take a look at Selective Preservation. It is better to place the gameAction outide targets:
```
return {
                    alwaysTriggers: true,
                    targets: targets,
                    gameAction: ability.actions.destroy((context) => ({
                        target: context.game.creaturesInPlay.filter(
                            (card) => !Object.values(context.targets).includes(card)
                        )
                    }))
                };
```

ryanch:
@romanogit man - you really know your stuff. Yeah that was better and totally solved the issue. updated the PR.

PR #2537, opened 2021-04-25
enhancement "fix"

SkyJedi:


server/services/DeckService.js
```diff
@@ -580,8 +580,8 @@ class DeckService {
         }
     }
 
-    solveEnhancements(list, deckCards) {
-        const cards = list.map((c) => deckCards.find((x) => x.id === c));
+    countEnhancements(list, deckCards) {
+        const cards = list.map((c) => deckCards.find((x) => x.id === c)).filter(Boolean);
```
romanogit, line 584: what is the meaning of filter(Boolean)?

SkyJedi: The list is the master vaults card array that also contains the tide card.  At this point the tide card has already been removed from deckCards and yields an 'undefined' in the find() function.

.filter(Boolean) removes that undefined.

romanogit: I do not understand why you pass list and deckCards to the same function. Why not work only with deckCards, which has tide removed right before you call countEnhancements.

SkyJedi: List is the array of 36 or 37 card uuids. deckCards is the card objects but only one of each, so it misses duplicated cards in the deck.  

I might overhaul how this all is processed after I get it working again

romanogit: it misses duplicates, but it guarantees enhanced cards are there, so you do not have to worry about the whole list.

cryogen:
Something about this doesnt seem quite right. But I'll merge it and look at it later

PR #2551, opened 2021-04-27
Flagging auto-enhanced decks as valid

### LINKED ISSUE #2547: Auto-enhanced Decks - Can the Enhancements show as Verified please?
Thank you for implementing the auto-enhancement of decks where the enhancements can only be placed one way (i.e. when there is only one type of enhancement and the number of enhancements matches the number of enhanced cards exactly).

Is it possible for these decks to also automatically show as Valid with Enhancements Verified?  Thank you.

SkyJedi:
Fixes #2547


romanogit:
I believe this is not needed anymore my friend @SkyJedi

PR #2570, opened 2021-05-03
Add option to choose first player

AleksandrShepelev:
Reason: during BO3 tournament games this option needed to choose first player for 2nd and 3rd game, because rematch don't allow spectators and even if rematch works, sometimes games are interrupted. need way to choose first player. 

Idea: make checkbox that allow your opponent to choose first player. This way we exclude abuse from game creator because if they mark this checkbox in random game, they give advantage to opponent, not to them. in tournament game, opponent just will choose correct player or winner will create game 2


AleksandrShepelev:
did it another way
https://github.com/keyteki/keyteki/pull/2574

PR #2574, opened 2021-05-03
Add manual command to change first player

AleksandrShepelev:
Manual command to change first player (for triad matches for example)

### DIFF
--- a/server/game/chatcommands.js
+++ b/server/game/chatcommands.js
@@ -28,7 +28,8 @@ class ChatCommands {
             '/stop-clocks': this.stopClocks,
             '/start-clocks': this.startClocks,
             '/token': this.setToken,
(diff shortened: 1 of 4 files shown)


server/game/chatcommands.js
```diff
@@ -329,6 +330,29 @@ class ChatCommands {
...
+        for (const p of players) {
+            if (p.name === firstPlayerName) {
+                this.game.activePlayer = p;
+                this.game.addAlert('info', '{0} changed first player to {1}', player, p);
+                return;
+            }
+        }
+        this.game.addAlert(
+            'danger',
+            'Cannot change first player: player {0} not exists',
```
cryogen, line 352: Small thing, but this should say "player {0} does not exist", I'll change it for you :)

Issue #2586, opened 2021-05-08
Blasters should not be able to bounce once attached to their "owner". 


blinkingline:
Once attached, the blaster's ability refers only to the specific blaster and not any other copies of the card - using the creature with the attached blaster does not allow its controller to move a different blaster. 
If the card is attached to the blaster's owner (For example Chan's Blaster is attached to Commander Chan) , the creature referenced is only the creature the blaster is attached to and not other copies of it - The creature with the attached blaster can not be used to move the blaster to a different copy. This was changed in the self-referential rules of Rulebook 1.6 from a previous ruling in Crucible Cast 17. 

"Blaster bouncing" was also confirmed not to work in an email from FFG:
![image](https://user-images.githubusercontent.com/7368564/117544475-74100080-afd6-11eb-829f-beb41d28a4f4.png)

romanogit:
Blaster boucing was so fun :(

PR #2588, opened 2021-05-08
Fix issue #2586

DustyMMiller:
This is to fix https://github.com/keyteki/keyteki/issues/2586

I tried just doing it with one function, but figured this was better because you can see that you are about to do 2 damage in the logs, so less likely to accidentally damage your own creature.

This failed the tests because I didn't want to update the tests for the blasters without approval/confirmation.

### DIFF
--- a/server/game/cards/03-WC/BlasterCard.js
+++ b/server/game/cards/03-WC/BlasterCard.js
@@ -34,11 +34,42 @@ class BlasterCard extends Card {
         return result;
     }
 
+    getDefaultAbility(ability) {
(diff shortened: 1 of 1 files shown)


romanogit:
I think you can write it simpler:

```
getDefaultAbility(ability) {
    return   {
             target: {
                   optional: true
                    cardType: 'creature',
                    gameAction: ability.actions.dealDamage({ amount: 2 })
                }
      };
```

romanogit:
Makes sense

PR #2671, opened 2021-06-16
setting cookieSecure: true and making localStorage preferable over cookie for i18n language preference

### LINKED ISSUE #2670: i18next cookie warning
I'm getting this warning in Firefox:

> Cookie “i18next” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies

romanogit:
Fixes #2670

### DIFF
--- a/client/i18n.js
+++ b/client/i18n.js
@@ -70,12 +70,14 @@ i18n
         nsSeparator: false,
         detection: {
             // order and from where user language should be detected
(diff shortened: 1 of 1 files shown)


romanogit:
@cryogen 
I do not have https local, so this is worthy a quick test right after deploy

PR #2679, opened 2021-06-19
Keyforge Live Banners (doover)

Cloggin:
(Jacked up the last so started over.)

Same plan, though, taking full advantage of the carousel to add 4x banners in specific sequence to the Lobby.

PR #2680, opened 2021-06-19
Keyforge Live Banners

SteVader:
On behalf of @Cloggin due to #2679 build failing.


SteVader:
@Cloggin - mine built, but I still can't tell you why. If I had to guess, I'd point at linting maybe? I have Prettier installed in VSCode.

Cloggin:
@SteVader It's cause I tried to do it at all, and the universe said "bro, you promised you'd only do mobile apps, what's this web shite?"

I'm still rocking visual studio 2019, I'm sure it has to do with copy/pasting/formatting/tabsandspaces/siliconvalleyreruns.

Thanks for the assist, here.

cryogen:
The issue was you tried to edit it without doing a 'npm install' so all the auto linting didn't happen.

However the bigger issue here is I'd like to keep it to a one banner per event if possible :)

I don't mind hosting the banners but if they start getting out of hand I might have to just start not doing them

Cloggin:
No rule of six in the banners huh? 

*Sad face.*

Alright roll with just the first one then. Please and thanks.

cryogen:
Sorry to be a party pooper :)

SteVader:
I've adjusted accordingly.

PR #2763, opened 2021-09-06
Change position and style of game timer

### LINKED ISSUE #2761: Timer position does not look good - should be at the same "line" as hand
image

SteVader:
Possibly fixes #2761

Totally open to this not being merged if it's hated. Was something I'd toyed with in the past.


cryogen:
Love this

Issue #947, opened 2019-11-18
No way to specify which cross house cheating you are using


DaveCordeiro:
![image](https://user-images.githubusercontent.com/57851149/69024880-5a19d280-098a-11ea-9ae9-1847bc3fed6a.png)
https://pastebin.com/jCQxtX4H
Here, the player with Val played a kirby then played trust no one, which it counted as the play for Val, then she could not play the off house creatures in her hand.

blinkingline:
I thought I had added this previously, but we did get a message back from FFG on this:

> I have three cards in my battleline; Captain Val Jericho in the Center of the Battleline with Com. Officer Kirby as a neighbor. On my turn, I reap with Com. Officer Kirby, granting me a play allowance of a non-Star Alliance action, artifact, or upgrade for the rest of the turn. Captain Val Jericho is also granting a play allowance of a card that is not of the active house. If I play a Shadows action, does this use up both allowances?
> This follows the same precedent as playing two Phase Shifts, which gives you two play allowances and only uses one on the first non-Logos card you play. Playing a Shadows action in this case will only use up one of your two play allowances, leaving you one more to use. You can choose whether to use Com. Officer Kirby's or Captain Val Jericho's play allowance first, leaving the other open (with the caveat that Captain Val has to be in the center of your battleline when you use her play allowance).

Issue #2153, opened 2020-12-18
Review FFG Rulings 16 December 2020


romanogit:
https://archonarcana.com/File:FFG_Rulings_16_December_2020.pdf

- [ ] If Shadow Self's neighbor is hit by Special Delivery and Shadow Self is Destroyed, Shadow Self is not purged.
- [ ] Dexus looks at the position of cards where they are originally played, not the state of the board after Play: abilities resolve.
- [ ] If there is enough Æmber to forge a key including using Æmber on cards like Senator Shrix, then a key must be forged, regardless of whether or not it is in the Forge a Key step.
- [ ] Keyforgery will be getting an errata in the next update.
- [ ] If The Archivist is face down in a player's Archive, it has no effect. "Puts into Archive" is different from "Archive".
- [ ] When resolving bonus icons “an additional time”, do the additional time immediately after resolving that icon. So, in this case, the second order is correct (Æmber, Æmber, Damage, Damage, Capture, Capture).
- [ ] Auto-Encoder resolves immediately after the discard of a card.
- [ ] Master of the Grey will prevent you from resolving bonus icons via abilities like those on Ensign El-Samra, Reclaimed by Nature, or Ultra Gravitron.
- [ ] If you play a Maverick Fandangle with 4 or more Æmber, it enters play ready.
- [ ] Scrivener Favian can resolve capture icons created by Amphora Captura as a steal.
- [ ] After Reckless Rizzo uses their Action: ability, they have a lasting effect that causes them to lose elusive. New instances of the elusive keyword (from Opportunist, for example) will be lost as soon as they are gained.
- [ ] Shadow Self's replacement effect takes the place of step 4 of dealing damage, meaning Shadow Self's Armor will not protect it from redirected damage.
- [ ] The Pale Star is essentially saying “creatures cannot have power other than 1 and armor other than 0”.
- [ ] You can play a card like Explo-rover as an upgrade around Lifeward.
- [ ] Matter Maker will let you play Explo-rover as an upgrade off-house.
- [ ] Duskwitch's Omega does not resolve when cloned by Cyber-Clone
- [ ] Mimicry copying a Shadows card counts as resolving a Shadows card for Dark Harbinger
[...]

PR #2771, opened 2021-10-10
Costs: paying w/ effects & effect selection prompt

dgbrokaw:
This PR is made to fix the Helper Bot task in [this list](https://github.com/keyteki/keyteki/issues/2153). I added tests to Helper Bot's, Com. Officer Kirby's, and Jericho's specs to cover this interaction. It should also apply to CXO Taber, but I didn't add tests for that. Should I?

What is `noActiveHouseForPlay`? Does there need to be a case before the effects part to cover the positive case of this effect?

In this PR, I could see:
- adding comments in costs.js to justify why effects go above active house (i.e. a rules reference)
- moving the "beforeEach" in Helper Bot and Jericho specs into the `it` because there's only one test
- deleting the Jericho test because it's redundant (because I think they all count as a `StaticEffect`?)

What do you think?

EDIT (2021/10/16):
I added a prompt when there are multiple house-cheating effects coming from different cards. This was done to fix https://github.com/keyteki/keyteki/issues/947.

image

### DIFF
--- a/server/game/costs.js
+++ b/server/game/costs.js
@@ -138,7 +138,7 @@ const Costs = {
                         !context.game.effectsUsed.includes(effect)
                 );
 
-                let effect = effects.find((effect) => {
+                let matchedEffects = effects.filter((effect) => {
                     let value = effect.getValue(context.player);
                     if (value.condition && !value.condition(context.source)) {
                         return false;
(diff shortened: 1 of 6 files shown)


dgbrokaw:
In the current implementation, there are two groups of effects:
Group 1: canPlayHouse, canPlayNonHouse (Helper Bot, Kirby, Jericho, etc.)
Group 2: canPlay (Matter Maker, United Action).

I don't know if there's a rules justification for these groups. Would it be that lasting effects take precedence over constant effects? Is there a rule for that?

dgbrokaw:
@cryogen @romanogit This needs an update to put canPlay effects into the same selection as the other house-cheating effects, IMO. Unfortunately the most recent answer from FFG does not address Matter Maker/United Action:

image

Maybe I can send out another email to FFG for clarification. But without knowing more I would think all these effects should be grouped together, as there is little evidence to the contrary.

dgbrokaw:
No response from FFG yet. I'm working on a new commit for this PR. Stay tuned.

dgbrokaw:
The new commit refactors the payEvent function for playing cards. I reversed my change that put canPlay effects as the last paying effect (because there was no clarification from FFG).

The results are:
- canPlay effects are almost the same as they were before, except you will be allowed to choose between them and other play allowance effects
- duplicate play allowances do not trigger a prompt, so nothing changes for players that stack Phase Shifts
- Helper Bot is fixed
- Captain Val Jericho and Kirby-like effects stacking are fixed.

PR #2799, opened 2022-01-11
Fix Brain Drain message

### LINKED ISSUE #2748: Brain Drain should say ".. to the top of their opponent's deck"
> LuLu uses Brain Drain to return Strange Ordination to the top of their deck

sdsantos:
Fixes #2748

I went if the Lost In The Woods style of message, to ensure the message it's always right, although it's slightly more verbose. This also might fix the message for other cards like Barehanded, although I haven't tried.


romanogit:
Fixed build error and repulled here https://github.com/keyteki/keyteki/pull/2858 thanks

PR #2811, opened 2022-04-02
migration to change the default value of the "order simultaneous abil…

dgbrokaw:
…ities" setting


cryogen:
Can you update the 'base' table definition as well? The migration is only really used by me in production - the table definitions set up the clean slate version

dgbrokaw:
I think this is what you meant.

dgbrokaw:
I pressed the wrong button there.

cryogen:
Yep thanks!

PR #2812, opened 2022-04-02
don't show a special tag for decks with enhancements

dgbrokaw:


romanogit:
I believe this is not needed anymore

PR #2813, opened 2022-04-02
always show the tide and let the active player raise the tide as many…

dgbrokaw:
… times as they want

Maybe we want to have a setting for this: "show tide controls when not playing DT." If either player has this turned on, it will show the tide no matter which set is being played. If both players have this turned off, it will work as it does now.

### DIFF
--- a/client/Components/GameBoard/PlayerStats.jsx
+++ b/client/Components/GameBoard/PlayerStats.jsx
@@ -61,7 +61,6 @@ const PlayerStats = ({
     size,
     spectating,
     stats,
-    tideRequired,
     user
 }) => {
     const { t } = useTranslation();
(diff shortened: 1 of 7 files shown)


client/Components/GameBoard/PlayerStats.jsx
```diff
@@ -145,19 +144,17 @@ const PlayerStats = ({
...
-                        key='tide'
-                        onClick={onClickTide}
-                        className='img-fluid tide'
-                        src={Constants.TideImages[stats.tide]}
-                        title={t(`${stats.tide}-tide`)}
-                    />
-                </div>
-            );
-        }
+        return (
```
dgbrokaw, line 147: Should I still check for stats here?

cryogen: Not sure why the check was there before, it should always be set. Guess we'll find out pretty quick after this goes live if it was needed..

PR #2834, opened 2022-08-17
Amber animations

dgbrokaw:
Demo video: https://youtu.be/h4tw1sq101E.

- the game object has an animations array
- game actions push animation names to the array
- the gameboard has an Animations component in between the player boards
- the Animations component contains Amber component instances, one for each animation in the array
- the Animations component modifies the name of the animation to match the perspective of the player
- the animation names correspond to CSS classes that have animation keyframes
- at the end of each turn, the animations array is cleared

All aember movement is between the starred locations here:
amber-animations-v2

A problem that remains is the aember dumped from effects like Praefectus Ludo when the board is wiped. Animations are spaced out when added at the same time from the same source. In this case, each creature adds one animation, and they all animate at the same time, making it appear as if there were only one aember moving.

I'm open to suggestions for improvement. I don't think I like how the various game actions are passing additional info to ModifyAmberAction. There was already a "reap" case, and I added to that. Maybe if we could get a value from ModfiyAmberAction saying whether the aember was redirected or not, and then we can have each action set up their own animations with that info?

### DIFF
--- a/client/Components/GameBoard/Amber.jsx
+++ b/client/Components/GameBoard/Amber.jsx
@@ -1,19 +1,13 @@
 import React from 'react';
-import { useTranslation } from 'react-i18next';
 
 import './Amber.scss';
 
 const AmberImage = require(`../../assets/img/amber.png`);
 
 const Amber = ({ animation, delay = 0 }) => {
-    const { t: translation } = useTranslation();
     return (
         <div>
-            <img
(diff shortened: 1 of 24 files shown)


dgbrokaw:
The TeamCity build failed. Any ideas what's wrong? I can't see the details.

dgbrokaw:
I'm adding comments.

Issue #2832, opened 2022-07-16
Hecatomb gives aember for warded Dis creatures


Tutenkharnage:
I played Hecatomb against an enemy board with five warded Dis creature. All five lost their wards, but my opponent gained 5 aember even though the wards replaced each creature's destruction, so the Dis creatures weren't "destroyed this way."

PR #2835, opened 2022-08-22
make hecatomb sequential so destroyed effects trigger at the same tim…

dgbrokaw:
…e, and also does not award amber for creatures that were not destroyed

Fixes https://github.com/keyteki/keyteki/issues/2832.


dgbrokaw:
Now I'm trying to improve the effect message :-)

romanogit:
Fixed the effect message and created a new PR
https://github.com/keyteki/keyteki/pull/2856

PR #2836, opened 2022-08-24
Update node to 12 in Dockerfile

### LINKED ISSUE #2833: ExploratoryCraft: flat is not a function
error: preThenContext.target.map(...).flat is not a function.

https://github.com/keyteki/keyteki/blob/eca831ec8ba43ebb231c098275f131e06c445bdb/server/game/cards/05-DT/ExploratoryCraft.js#L17

Turns out the provided Dockerfile uses node v10, which doesn't have Array.prototype.flat() yet.

https://github.com/keyteki/keyteki/blob/b436d2901d04097b4a450a243b8da8e2f04e2b91/Dockerfile#L1

It's likely safe to just bump the node version to 11, I'll be testing this locally for the next few days.

ericchu94:
Some javascript functions used are not available in node 10, such as [Array.prototype.flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) and [Array.prototype.flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat).

These functions are used in the following cards:
https://github.com/keyteki/keyteki/blob/eca831ec8ba43ebb231c098275f131e06c445bdb/server/game/cards/05-DT/ExploratoryCraft.js#L17
https://github.com/keyteki/keyteki/blob/eca831ec8ba43ebb231c098275f131e06c445bdb/server/game/cards/05-DT/WipeClear.js#L12

Upgrade to node 12 addresses this.

Fixes #2833

Testing done:
- [x] Played games with impacted cards

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 COPY package.json /usr/src/app


cryogen:
Thanks!

PR #2844, opened 2022-09-17
Make Groundbreaking Discovery  destroy upgrades

### LINKED ISSUE #2826: Groundbreaking discovery does not destroy upgrades on warded creature
If a creature has a ward when groundbreaking discovery is triggered (such as via encounter suit), then the ward is removed but upgrades remain. Groundbreaking discovery says to destroy all cards in play, so I believe the upgrades should be removed even if the creature is not?

No log or screen shots but should be easy to verify with this deck. It happened when zenzi had encounter suit, light of the archons, aember vac on it.
https://decksofkeyforge.com/decks/4f269c70-55b7-44a2-a148-3acb2172ca74

jtrussell:
Previously, upgrades on warded creatures would remain if the creature they were attached to managed to survive GBD.

It might have been cleaner to go upstream and include upgrades in whatever is initially populating `context.game.cardsInPlay`, I wasn't sure if upgrades were being left out of that list intentionally.

Fixes #2826

### DIFF
--- a/server/game/cards/05-DT/GroundbreakingDiscovery.js
+++ b/server/game/cards/05-DT/GroundbreakingDiscovery.js
@@ -12,7 +12,10 @@ class GroundbreakingDiscovery extends Card {
                 ),
             gameAction: ability.actions.sequential([
                 ability.actions.destroy((context) => ({
-                    target: context.game.cardsInPlay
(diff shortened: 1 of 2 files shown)


romanogit:
@cryogen , this one looks good to merge

cryogen:
Thanks!

Issue #2876, opened 2022-10-22
Cross expasion Alliance is possible


Osarek:
Im fixing that in the Hour

![image](https://user-images.githubusercontent.com/57572496/197355742-1ae712ea-5e88-4c60-90c7-42e02bb9285a.png)

from Cota https://www.keyforgegame.com/deck-details/754f2397-e8c1-4044-9e10-08a3a8acf14e
and aoa https://www.keyforgegame.com/deck-details/ef0c8859-ee05-4359-8500-909cdb37946d
and Dt https://www.keyforgegame.com/deck-details/73b1a1a7-2cf1-4dc6-956f-79589b0d4a98

Osarek:
PR is ready https://github.com/keyteki/keyteki/pull/2878

Osarek:
Its fix on prod,  can not reproduce anymore

PR #2878, opened 2022-10-22
#2876 alliance multiple expasion

Osarek:
image


server/services/DeckService.js
```diff
@@ -440,6 +441,17 @@ class DeckService {
             }
 
             decksByUuid[dbDeck.uuid] = dbDeck;
+            expansionDeck.push(dbDeck.expansion);
```
cryogen, line 444: Just store the expansionid the first time round and then compare against it, returning error if it doesn't match. No need to do all this array comparing

cryogen (changes requested):

server/services/DeckService.js
Osarek: fixed

Osarek:
But i have a "Cannot read property 'house' of undefined"
when doing a valid alliance, seems that is also on master

image

cryogen (approved):

PR #2880, opened 2022-10-23
Fix/#2879 allow all decks in alliance

Osarek:
No regression in decklist,
No regression in deck selection for alliance

Allaince deck are now available in alliance

image


server/services/DeckService.js
```diff
@@ -284,6 +284,9 @@ class DeckService {
             } else if (filterObject.name === 'isAlliance') {
                 filter += `AND ${this.mapColumn(filterObject.name)} = $${index++} `;
                 params.push(filterObject.value);
+            } else if (filterObject.name === 'showAllDecks' && !filterObject.value) {
```
cryogen, line 287: Please don’t change this file. It’s suppose to handle filtering on database columns only

cryogen (changes requested):

cryogen (approved):

PR #2882, opened 2022-10-24
Feature/more stats in deck summary

Osarek:
image


romanogit:
I'd rather have a more detailed deck statistics where we click on a link bellow the normal stats and we would navigate to a page that could show what you are showing.
I also think that it would be nice to have a last 15 (or 10 or 50, selectable?!) matches statistics. Sometimes you improve with a deck and having old data in the calculation could be misleading.

Osarek:
Hi,
I totally agree about that.
It will also prevent to calculate all mode stat for each deck everything.

What do you think about a link in the in game chat section, that open a new
tab of the select deck stats( specialy on opponent deck)

Le mar. 25 oct. 2022 à 02:10, Romano Silva ***@***.***> a
écrit :

romanogit:
I would never add stats for opponent's deck. That's something I believe @cryogen would also oppose.
I would put a link on the deck list or stats to a full stats page. 
This stats page would show what you are proposing. But only stats of decks you have regsistered.

server/services/DeckService.js
```diff
@@ -310,10 +310,21 @@ class DeckService {
...
+                    'CASE WHEN "CompetitiveWinCount" + "CompetitiveLoseCount" = 0 THEN 0 ELSE (CAST("CompetitiveWinCount" AS FLOAT) / ("CompetitiveWinCount" + "CompetitiveLoseCount")) * 100 END AS "CompetitiveWinRate" ' +
+                    'FROM ( ' +
                     'SELECT d.*, u."Username", e."ExpansionId" as "Expansion", (SELECT COUNT(*) FROM "Decks" WHERE "Name" = d."Name") AS DeckCount, ' +
                     '(SELECT COUNT(*) FROM "Games" g JOIN "GamePlayers" gp ON gp."GameId" = g."Id" WHERE g."WinnerId" = $1 AND gp."DeckId" = d."Id") AS "WinCount", ' +
-                    '(SELECT COUNT(*) FROM "Games" g JOIN "GamePlayers" gp ON gp."GameId" = g."Id" WHERE g."WinnerId" != $1 AND g."WinnerId" IS NOT NULL AND gp."PlayerId" = $1 AND gp."DeckId" = d."Id") AS "LoseCount" ' +
+                    '(SELECT COUNT(*) FROM "Games" g JOIN "GamePlayers" gp ON gp."GameId" = g."Id" WHERE g."WinnerId" != $1 AND g."WinnerId" IS NOT NULL AND gp."PlayerId" = $1 AND gp."DeckId" = d."Id") AS "LoseCount", ' +
```
cryogen, line 321: This should all be fetched in one join and then grouped. If you can't figure out the SQL for grouping it, then grouping it in code is acceptable

client/pages/DeckStat.jsx
```diff
@@ -0,0 +1,186 @@
...
+    }));
+    const dispatch = useDispatch();
+    const { t } = useTranslation();
+
+    useEffect(() => {
+            // if (user != null && deck != null) {
+            dispatch(
+                loadStats({
+                    userId: 1,
+                    deckId: 5
```
Osarek, line 62: 1rst promblem,
so here im supposed to use user.id and deck.id, the one loaded at line 45, 48, the value of user and deck are nul here but not at line 73,74

client/pages/DeckStat.jsx
```diff
@@ -0,0 +1,186 @@
...
+    let { stats } = useSelector((state) => ({
+        stats: state.stats
+    }));
+    const dispatch = useDispatch();
+    const { t } = useTranslation();
+
+    useEffect(() => {
+            // if (user != null && deck != null) {
+            dispatch(
+                loadStats({
```
Osarek, line 60: And 2nd probleme, i dont know how the load the result of this api call into the variable stats

client/pages/DeckStat.jsx
cryogen: You're in a useEffect callback. This is called whenever the things in the array change. So you want deck.id and user.id to be in the array. Then only set these values if they are present

client/pages/DeckStat.jsx
cryogen: You don't. It's the useSelector line above

client/pages/DeckStat.jsx
Osarek: So this problem is now fix, thanks

PR #2890, opened 2022-10-29
Fix/alliance deck in normal archonte

Osarek:


client/Components/Decks/DeckList.jsx
```diff
@@ -185,6 +185,14 @@ const DeckList = ({ deckFilter, onDeckSelected, standaloneDecks = false }) => {
                     value: data.filters[k].filterVal
                 }));
 
+                //filter comming from elsewhere than the table change
+                if (
```
cryogen, line 189: if(deckFilter?.isAlliance)

Osarek:
there is 3 case of calling this method

1/ Decks page : deckFilter?.isAlliance=undefined
--> we don't need to add the filter
2/ Deck selection for Alliance : deckFilter?.isAlliance=true
--> we need to add the filter
3/ Deck selection for Normal: deckFilter?.isAlliance=false
--> we need to add the filter


should i do
if (deckFilter?.isAlliance == false || deckFilter?.isAlliance == true) 
or 
if (typeof deckFilter.isAlliance !== 'undefined' )

cryogen:
> 

I see what you're trying to do.  Just checking if(deckFilter) should be enough in that case

Osarek:
I think that should be what I want
Did not test that for the moment 

if (deckFilter?.isAlliance?)

Osarek:
> > 
> 
> I see what you're trying to do. Just checking if(deckFilter) should be enough in that case

You're right, that's enough to cover the 3 cases.
thanks

PR #2919, opened 2022-11-26
Card generator pr

Toeofdoom:
This is me finally getting around to creating a PR for the card parser - there's a lot here, a good place to start is server/scripts/generatecards/readme.md for an overview of how it works.

Specific areas I'd be interested in feedback for:
* The testing method - you can run these to see how well the generated cards meet existing test cases by having it overwrite hundreds of cards (current results - fails 104/19799). This is also why several cards were renamed or moved - means this command actually overwrites them instead of leaving 2 copies around!
    * npm server/scripts/generateCards.js --overwrite
    * npm test
* Does the code make any sense, especially the parser and generator parts? Any suggestions for improving this?
* Did I miss anything because I don't usually write javascript? The package-lock.json isn't updated because I think my older version of node changed it back to version 1...


romanogit:
Can you rebase the PR to WoE branch?

romanogit:
Nevermind, figured out how to do it... 😛

PR #2955, opened 2023-05-17
Download decks csv

Neugebauer:
This change adds a button to download your non-Alliance decks win/loss records as a csv file. 
Casual users will use this to see decks sorted by games played as well as their all-deck totals.
Advanced users will will use this data to add to their deck quality calculation projects.


Osarek:
Does it contains also deck that we have deleted?

PR #2958, opened 2023-05-19
WoE: hook up `readiesDuringReadyPhase` and implement Frost Giant

strib:
It looks like the `readyDuringReadyPhase` wasn't actually used anywhere, so I took my best guess about how to hook it up.  Very open to changing this to something better!

### DIFF
--- a/server/game/GameActions/ReadyAction.js
+++ b/server/game/GameActions/ReadyAction.js
@@ -8,6 +8,9 @@ class ReadyAction extends CardGameAction {
     }
 
     canAffect(card, context) {
+        if (card.game.currentPhase === 'ready' && !card.readiesDuringReadyPhase()) {
+            return false;
+        }
         return (
             card.location === 'play area' &&
             card.checkRestrictions('ready', context) &&
@@ -16,12 +19,9 @@ class ReadyAction extends CardGameAction {
     }
 
     getEvent(card, context) {
-        return super.createEvent('onCardReadied', { card: card, context: context }, () => {
-            if (card.game.currentPhase === 'ready' && !card.readiesDuringReadyPhase()) {
-                return;
-            }
-            card.ready();
-        });
+        return super.createEvent('onCardReadied', { card: card, context: context }, () =>
+            card.ready()
+        );
     }
 }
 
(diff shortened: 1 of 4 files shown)


romanogit (changes requested):

server/game/GameActions/ReadyAction.js
```diff
@@ -16,9 +16,12 @@ class ReadyAction extends CardGameAction {
     }
 
     getEvent(card, context) {
-        return super.createEvent('onCardReadied', { card: card, context: context }, () =>
-            card.ready()
-        );
+        return super.createEvent('onCardReadied', { card: card, context: context }, () => {
+            if (card.game.currentPhase === 'ready' && !card.readiesDuringReadyPhase()) {
```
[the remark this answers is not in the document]
strib: Good call, I didn't notice `canAffect`.  Done!

server/game/GameActions/ReadyAction.js
```diff
@@ -8,6 +8,9 @@ class ReadyAction extends CardGameAction {
     }
 
     canAffect(card, context) {
+        if (card.game.currentPhase === 'ready' && !card.readiesDuringReadyPhase()) {
```
romanogit, line 11: I think `doesNotReady` is actually not used anywhere. Maybe Kiri Giltspine could have used it in the past.
I think for consistency, we could follow how Kiri Giltspine is done or Storm Surge.

romanogit (changes requested):

server/game/GameActions/ReadyAction.js
strib: I did look at Kiri Giltspine but it is actually worded slightly differently because it is a lasting effect tied to the controller at the time the effect is put into place.  So if it changes sides (like in one of its tests), it actually _would_ ready.  But Frost Giant is a persistent effect and does not matter who controls it, so i can't just use the same code there.  Let me know if I'm missing something.

I updated Storm Surge to use `doesNotReady` and the tests still pass, and it looks a bit cleaner to me.  What do you think?

romanogit: Can you commit the  updated Storm Surge?

strib: Oops, I thought I had, sorry.  It's there now, ready for another look.

PR #2960, opened 2023-05-19
WoE: implement Kretchee

strib:
Added new effects for capturing and exalting one extra amber.

Seems like this might be a good one to have an extra log message for, but I haven't figured out how to do that yet.  Any tips would be appreciated!

### DIFF
--- a/server/game/GameActions/ExaltAction.js
+++ b/server/game/GameActions/ExaltAction.js
@@ -22,8 +22,8 @@ class ExaltAction extends AddTokenAction {
             { card: card, context: context, amount: this.amount },
             () => {
                 let extra = 0;
-                if (this.amount > 0 && player.anyEffect('exaltOneMoreFromPool')) {
-                    extra = 1;
+                if (this.amount > 0) {
+                    extra = player.sumEffects('exaltMoreFromPool');
                 }
 
                 card.addToken('amber', this.amount + extra);
--- a/server/game/GameActions/ExaltAction.js
+++ b/server/game/GameActions/ExaltAction.js
@@ -16,10 +16,18 @@ class ExaltAction extends AddTokenAction {
     }
 
     getEvent(card, context) {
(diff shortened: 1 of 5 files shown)


server/game/effects.js
```diff
@@ -122,6 +122,8 @@ const Effects = {
     restrictHouseChoice: (house) => EffectBuilder.player.static('restrictHouseChoice', house),
     stealFromPool: () => EffectBuilder.player.static('stealFromPool'),
     captureFromPool: () => EffectBuilder.player.static('captureFromPool'),
+    captureOneMoreFromPool: () => EffectBuilder.player.static('captureOneMoreFromPool'),
```
romanogit, line 125: I believe the implementation is good, but we could make this more flexible by adding an amount. You could name, for example: "captureMoreFromPool" or "captureExtraFromPool"

server/game/GameActions/ExaltAction.js
```diff
@@ -16,10 +16,18 @@ class ExaltAction extends AddTokenAction {
...
 
     getEvent(card, context) {
+        let player = this.player || context.player.opponent;
         return super.createEvent(
             'onExalt',
             { card: card, context: context, amount: this.amount },
-            () => card.addToken('amber', this.amount)
+            () => {
+                let extra = 0;
+                if (this.amount > 0 && player.anyEffect('exaltOneMoreFromPool')) {
```
romanogit, line 25: You need to count the effects because there could be two or more Kretchee in play

romanogit (changes requested):

server/game/GameActions/ExaltAction.js
strib: Oops, great catch.  Fixed.

server/game/effects.js
strib: Good suggestion, done!

PR #2961, opened 2023-05-20
WoE: implement Broken Axe Outpost

strib:


server/game/cards/06-WoE/BrokenAxeOutpost.js
```diff
@@ -0,0 +1,25 @@
...
+                controller: 'self',
+                gameAction: ability.actions.returnToDeck({ bottom: true })
+            },
+            then: {
+                target: {
+                    cardType: 'creature',
+                    controller: 'opponent',
+                    gameAction: ability.actions.dealDamage({ amount: 6 })
+                },
+                message: '{0} uses {1} and puts {2} on the bottom of their to deal 6 damage to {3}'
```
strib, line 17: I have no idea if this is right.  I haven't been able to figure out how the log code works.  Pointers very much appreciated!

romanogit: you can do `export DEBUG_TEST=1`
and then run your tests.

romanogit: I think the message should be `{0} uses {1} to put {2} on the bottom of their deck and deal 6 damage to {3}`

romanogit (changes requested):

server/game/cards/06-WoE/BrokenAxeOutpost.js
strib: Great tip, thanks!  `returnToDeck` already prints a message, and in the `then` clause I don't seem to have a `{3}` available, so now there are just 2 separate messages which seems fine:

```
2023-05-21 16:53:15 player1 uses Broken Axe Outpost to return Bumpsy to the bottom of their deck
2023-05-21 16:53:15 player1 uses Broken Axe Outpost to deal 6 damage to Kelifi Dragon
```

PR #2969, opened 2023-05-21
DT with KO locale

romanogit:
Trying again...


cryogen:
Has this been included with the woe update?

romanogit:
no, this is only the KO locale, probably we need to merge master in the data repo and then update this PR to have WOE as well

PR #2972, opened 2023-05-21
WoE: implement Conductor Jărroyă

strib:


test/server/cards/06-WoE/ConductorJarroya.spec.js
```diff
@@ -0,0 +1,29 @@
...
+                    inPlay: ['conductor-jarroya', 'toad', 'shigziso-buggy', 'gauntlet-of-command']
+                },
+                player2: {
+                    inPlay: ['shisnyasi-buggy']
+                }
+            });
+        });
+
+        it('it should ready friendly buggy artifacts', function () {
+            this.player1.useAction(this.shigzisoBuggy);
```
romanogit, line 17: I believe the block from 17 to 19 are not needed as they are asserting Shugziso Buggy which you eventually set exhausted in line 21.

romanogit (changes requested):

test/server/cards/06-WoE/ConductorJarroya.spec.js
strib: 17-19 deal with the friend Shigziso Buggy, but line 21 deals with the enemy Shisnyasi Buggy.  I believe 17-19 are still needed.

Man these are going to be hard to play with.

romanogit: Yes, it is like playing Lord of the Rings in Sindarin

PR #2973, opened 2023-05-22
WoE: implement Corner the Market

strib:


strib:
There are a few problems with this PR:

* It doesn't actually replace the discard, it only moves the card _after_ it hits the discard.  So, for example, I think Auto-Encoder would still trigger when it shouldn't.  I suspect we need something more like in Dimension Door, but I don't know how to do that for discarding.  Any pointers?
* Even as it is, it fails in the test on an "undefined" error while trying to archive.  I tried to model it after Scout Pete, but apparently there's a difference between discarding and archiving that I haven't been able to track down yet.  Any help would be much appreciated!

strib:
The other thought I had here is maybe we can just replace the "Play" option for each card with an "Archive" option.  Might be a little easier?

strib:
Ok, I think I figured out how to interrupt the discard event and redirect the card to archive if desired.  Please take a look!

PR #2975, opened 2023-05-22
WoE: implement Envoy Of Ekwirre

strib:


server/game/GameActions/SwapEverythingAction.js
```diff
@@ -0,0 +1,51 @@
+const CardGameAction = require('./CardGameAction');
+
+class SwapEverythingAction extends CardGameAction {
```
romanogit, line 3: Not sure I like the name, though.
What do you think about adding a pair of attributes in the SwapAction: `swapTokens = false`, `swapUpgrades = false`?

strib: Good idea, done!  Please take another look.

romanogit: Beautiful

PR #2982, opened 2023-05-23
WoE: implement Exchange Program

strib:


romanogit:
I need to double check on swap before review

strib:
I updated this to use `takeControlOn` like #3227 .  I noticed a bug in `takeControlOn` while testing so I fixed it here too.  (We need to allow splicing in at the end of the array as well.)

PR #2984, opened 2023-05-24
WoE: implement Niffle Paw

strib:


server/game/cards/06-WoE/NifflePaw.js
```diff
@@ -0,0 +1,27 @@
...
+                onReap: (event, context) => event.card === context.source.parent
+            },
+            gameAction: ability.actions.destroy((context) => ({ target: context.source.parent })),
+            then: {
+                alwaysTriggers: true,
+                target: {
+                    numCards: 1,
+                    cardType: ['creature'],
+                    cardCondition: (card) => card.location === 'play area',
+                    gameAction: ability.actions.attach((context) => ({
```
strib, line 16: I can't figure out why I'm not seeing a log message for this attachment action.  It's probably something simple.  Help appreciated!

romanogit: Because that's on a `then` statement. you have to provide a message explicitly.

strib: Thanks, fixed!

romanogit: message should follow the format "{0} uses {1} to move ... "... something like that....

romanogit (changes requested):

server/game/cards/06-WoE/NifflePaw.js
strib: So it's ok to have "uses Niffle Paw to move Niffle Paw to X"?  That seemed weird to me but happy to do it.

strib: Ok did it that way.

romanogit: Yes, that's how "effects" are worded. But in this case, I think it should say, for example:

"romano uses "creature" to move Niffle Paw to "another creature"

PR #2985, opened 2023-05-24
WoE: implement Outnegotiate

strib:


server/game/cards/06-WoE/Outnegotiate.js
```diff
@@ -0,0 +1,23 @@
...
+        this.play({
+            target: {
+                controller: 'opponent',
+                location: 'discard',
+                gameAction: [
+                    ability.actions.steal((context) => ({
+                        target: context.target.owner,
+                        amount: context.target.bonusIcons.length
+                    })),
+                    ability.actions.returnToDeck({ bottom: true })
```
strib, line 14: Having log issues again here; I don't see the return to deck message get printed in the log.

romanogit: Because it is an array of events, there is a flag that allows multiple messages to be displayed. 
Add `effectStyle: all` at the `target` level.

strib: Thanks, done!

Issue #2902, opened 2022-11-17
WoE - Ekwidon house - Specific issues created for pending cards


romanogit:
This list is still based on Spoilers, so, implement with care:
- [x] (057) **Æmbitrage**: Your keys cost +1 Aember. During your "draw cards" step, refill your hand to 1 additional card.
- [x] (058) **Antiquities Dealer**: Action: Gain 2 Aember if you control at least one artifact.
- [x] (059) **Avid Collecting**: Play: Return a friendly token creature to your hand.
- [x] (060) **Conductor Jărroyă**: After Reap: Ready each friendly Buggy artifact.
- [x] (061) **Corner the Market**: Play: During your opponent`s next turn, they cannot play cards, and each time they discard a card from their hand, they may archive that card.
- [x] (062) **Cursed Relic**: Enhance Aember Aember Aember Aember Aember Aember. You cannot play Cursed Relic or choose to discard it from your hand.   
- [x] (063) **Envoy of Ekwirrĕ**: Reap: Swap Envoy of Ekwirrĕ with one of its neighbors. Also swap all Aember, damage, counters, and upgrades on those creatures.
- [x] (064) **Flea Market**: Action: Look at a random card in your opponent`s hand. You may give your opponent 1 Aember. If you do, play that card as if it were yours.
- [x] (065) **Freelancer**: At the start of each turn, the active player takes control of this creature. This creature may be used as if it belonged to 
the active house.
- [x] (066) **Insurance Policy**: Play: Lose 1 Aember. This creature gains, "Destroyed: If it is not your turn, gain 4 Aember."
- [x] (067) **Molluscaller**: Reap: For the remainder of the turn, each friendly Strange Shell gets +3 power and loses all abilities.
- [x] #3130
- [x] (069) **Talent Scout**: Talent Hunter can be used as if it belonged to the active house. Play: Look at your opponent`s hand and play a creature as if it were yours. Your opponent gains control of Talent Hunter.
- [x] (070) **The Æmber Road**: Omni: Put 1 trade counter on The Æmber Road, then gain 1 Aember for each trade counter on it. Give control of The Æmber 
Road to your opponent.
- [x] (071) **Token of Appreciation**: Play: Make a token creature. Forge a key at +7 Aember current cost, reduced by 1 Aember for each friendly token creature.
- [x] (072) **Tourist Trap**: Play: Make a token creature. Action: Choose a friendly token creature and an enemy creature. If you do, swap control of those creatures.
[...]

PR #2990, opened 2023-05-24
Woe: implement Shizyoku Swopper

strib:
Issue: #2902

### DIFF
--- a/server/game/GameActions/SwapAction.js
+++ b/server/game/GameActions/SwapAction.js
@@ -27,6 +27,10 @@ class SwapAction extends CardGameAction {
 
     getEvent(card, context) {
         return super.createEvent('onSwap', { card: card, context: context }, () => {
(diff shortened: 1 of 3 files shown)


romanogit:
I am not sure if `swap` can be used to take control. I will double check.

strib:
@romanogit hmm you are right, I don't think `swapControl` actually updates the controller.  But I don't think 2 calls to `takeControl` would be correct, since the swap has to happen all at once to avoid side effects.  What do you suggest?  Should I generalize `swap()` to also update the controllers?  Anything else that needs to be called, like `updateEffectsContexts()`?

romanogit:
I asked a question about it in discord channel

strib:
I'm pretty confident the existing swap rule covers it.  Just like in the same battleline, it should happen simultaneously and not trigger things like Picaroon.

romanogit:
Took your code and changed it to takeControlOn on #3227

Issue #2904, opened 2022-11-17
WoE - Sanctum house - Specific issues created for pending cards


romanogit:
This list is still based on Spoilers, so, implement with care:
- [x] (177) **Batariel of the Grey**: After Reap: Ready each Disciple.
- [x] (179) **Friendship**: When this creature would be dealt damage, if it has 1 or more neighbors, that damage is divided among its neighbors as evenly as possible instead.
- [x] #3181
- [x] (182) **Heavy Subsidies**: Keys cost +6 Aember. Each creature gains, "After Reap: Gain 1 Aember."
- [x] (190) **Renewed Life**: Play: Fully heal a creature. Make a token creature for each damage healed this way.
- [x] #3182
- [x] #3183
- [x] #3184
in play. After Reap: Gain 2 Aember.
- [x] (198) **Grey Abess**: Play: Make a token creature. Each friendly token creature gets +1 armor.
- [x] (201) **Hymn to Duma**: Each friendly creature gains, "Omni: Destroy this creature. A friendly creature captures 2 Aember."
- [x] (202) **Katerina the Summoner**: Play/After Reap: Capture 1 Aember. Destroyed: Make a token creature for each Aember on Katerina the Summoner.    
- [x] #3185
- [x] (205) **Martyr of the Vault**: Destroyed: If your opponent has 7 Aember or more, they lose all but 5 Aember.
- [x] (208) **Squire Recruitment**: Play: Make a token creature for each friendly Knight creature.
- [x] (209) **Absolve**: Play: Heal 1 Damage from each creature. Each creature healed this way captures 1 Aember from its opponent.
- [x] (210) **Battlevangelist**: After Fight: Make a token creature.
- [x] (211) **Celestial Gorm**: Omni: Destroy Celestial Gorm. Return each other artifact to its owner`s hand.
- [x] (213) **Colonel Mariana**: Play: Each friendly Knight captures 1 Aember.
- [x] (215) **Kassiel the Benevolent**: Enhance Aember Capture Capture
- [x] #3186
- [x] (217) **Membership Drive**: Play: Make a token creature. Gain 1 Aember for each friendly token creature.
- [x] (218) **Muster**: Play: Make a token creature. If your opponent has more Aember than you, archive Muster.
- [x] #3187
This creature gains taunt. (This creature&#039;s neighbors cannot be attacked unless they have taunt.)
- [x] (220) **Outflank**: Play: Each Sanctum flank creature captures 2 Aember from its opponent.
[...]

PR #2992, opened 2023-05-24
WoE: Implement Absolve

jtrussell:
See #2904

### DIFF
--- a/server/game/cards/06-WoE/Absolve.js
+++ b/server/game/cards/06-WoE/Absolve.js
@@ -1,0 +1,42 @@
+const Card = require('../../Card.js');
+
+class Absolve extends Card {
+    // Play: Heal 1D from each creature. Each creature healed this way captures 1A from its opponent
+    setupCardAbilities(ability) {
+        this.play({
+            gameAction: ability.actions.heal((context) => ({
+                amount: 1,
+                target: context.game.creaturesInPlay.filter((card) => card.hasToken('damage'))
+            })),
+            then: {
+                gameAction: [
+                    ability.actions.capture((context) => ({
+                        target: context.preThenEvents
+                            .filter(
+                                (event) =>
+                                    !event.cancelled &&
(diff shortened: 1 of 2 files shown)


romanogit:
This one is big, I will review tomorrow.

server/game/cards/06-WoE/Absolve.js
```diff
@@ -0,0 +1,50 @@
...
+    // Play: Heal 1D from each creature. Each creature healed this way captures 1A from its opponent
+    setupCardAbilities(ability) {
+        this.play({
+            //effect: 'heal 1 damage from all creatures',
+            gameAction: ability.actions.heal((context) => ({
+                amount: 1,
+                target: context.game.creaturesInPlay.filter((card) => card.hasToken('damage'))
+            })),
+            then: {
+                message: '{1} heals {3}, capturing {0} {4} amber',
```
romanogit, line 13: I think this message can be avoided because the previous effect will state which creatures will be healed.

test/server/cards/06-WoE/Absolve.spec.js
```diff
@@ -0,0 +1,39 @@
+describe('Absolve', function () {
+    describe("Absolve's ability", function () {
```
romanogit, line 2: It would be interesting to add a test case where you heal more creatures than amber from your opponent and from yourself.

The engine should prompt for creatures to capture.

jtrussell: Yep, there's something similar for Pandemonium. I'll add a test case 👍 .

jtrussell:
@romanogit  message removed and added a test for when there are more damaged creatures than amber to capture 🙂.

test/server/cards/06-WoE/Absolve.spec.js
jtrussell: Added!

PR #2994, opened 2023-05-25
WoE: Implement Batariel of the Grey

jtrussell:
Marking as a draft - wasn't seeing opposing disciples show up as cards in play for player 2, i.e. `this.player2.player.cardsInPlay.length === 0` in the test below. I'll revisit but figured I'd toss up the draft in case it was related to something not fully implemented with tokens (non-token cards seemed to show up as I expected).

See: #2904

### DIFF
--- a/server/game/cards/06-WoE/BatarielOfTheGrey.js
+++ b/server/game/cards/06-WoE/BatarielOfTheGrey.js
@@ -1,0 +1,16 @@
+const Card = require('../../Card.js');
+
+class BatarielOfTheGrey extends Card {
+    // After Reap: Ready each Disciple.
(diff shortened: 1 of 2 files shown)


romanogit:
That's because "Disciple" is a token creature which I believe is not yet implemented. For token creatures to work, one needs to define the attribute "token" in each player (hopefully it works for player2 as well):

See https://github.com/keyteki/keyteki/blob/woe/test/server/cards/06-WoE/B0T.spec.js

jtrussell:
Thanks for the tip, @romanogit! I updated the test to handle tokens in the expected way 🙂.

PR #2998, opened 2023-05-25
WoE: implement Mass Buyout

strib:
Issue: #2902

### DIFF
--- a/test/server/cards/06-WoE/MassBuyout.spec.js
+++ b/test/server/cards/06-WoE/MassBuyout.spec.js
@@ -1,0 +1,37 @@
+describe('Mass Buyout', function () {
+    describe("Mass Buyout's ability", function () {
+        beforeEach(function () {
+            this.setupTest({
+                player1: {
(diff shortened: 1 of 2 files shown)


test/server/cards/06-WoE/MassBuyout.spec.js
```diff
@@ -0,0 +1,37 @@
+describe('Mass Buyout', function () {
```
romanogit, line 1: Can you add test cases for warded creatures and include creatures that goes to other places other than discard, such as Bad Penny? I think it might fail with warded creatures.

strib: The test already has a warded creaure (`Gub`).  I can add Bad Penny in later, but this code is just copied directly from Final Analysis so it should have the same behavior as that one.

strib: (The ward is what makes player2 in the second test only gain 1 amber, rather than 2).

PR #3000, opened 2023-05-25
WoE: implement Shrewd Investor

strib:
Issue: #2902

### DIFF
--- a/server/game/cards/06-WoE/ShrewdInvestor.js
+++ b/server/game/cards/06-WoE/ShrewdInvestor.js
@@ -1,6 +1,6 @@
 const Card = require('../../Card.js');
 
-class ShrewInvestor extends Card {
+class ShrewdInvestor extends Card {
     // Play: You may have your opponent gain 1Aember. If you do,
     // capture 4Aember.
     setupCardAbilities(ability) {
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/ShrewdInvestor.js
```diff
@@ -0,0 +1,23 @@
+const Card = require('../../Card.js');
+
+class ShrewInvestor extends Card {
```
romanogit, line 3: Class name should be `ShrewdInvestor`

strib: Oops, good catch.  My fingers have problems with that word for some reason.  Fixed!

server/game/cards/06-WoE/ShrewdInvestor.js
```diff
@@ -0,0 +1,23 @@
...
+    setupCardAbilities(ability) {
+        this.play({
+            optional: true,
+            gameAction: ability.actions.sequential([
+                ability.actions.gainAmber((context) => ({
+                    target: context.player.opponent
+                })),
+                ability.actions.capture({ amount: 4 })
+            ]),
+            message: '{0} uses {1} to cause {2} to gain 1 amber, and then capture 4 amber onto {0}',
```
romanogit, line 15: I think you should remove this `onto {0}` as zero is the player's name

strib: Yep, fixed.

romanogit:
Is the DockerFile supported to be pushed?

strib:
Sorry, no, I just noticed and reverted that.

That new `npm run build`` step really slows down my dev cycle though, I need to find a better way around it...

Issue #2903, opened 2022-11-17
WoE - Mars house  - Specific issues created for pending cards


romanogit:
This list is still based on Spoilers, so, implement with care:
- [x] #3030
- [x] #3031
- [x] #3032
- [x] #3033
- [x] #3034
- [x] (126) **Nautilixian**: Nautilixan cannot be used unless you have used 1 or more friendly Mars creatures this turn. Each friendly Pilot creature cannot be dealt damage or destroyed.
- [x] (129) **R-Evolution**: This creature and each creature with the same name as it gain, "After Reap: Deal 2 Damage to an enemy creature."
- [x] (130) **Replay Pod**: Each friendly Mars creature gains "Destroyed: Place this creature facedown under Replay Pod."   Action: Put each card under 
Replay Pod into your hand. Purge Replay Pod.
- [x] (131) **Revo Hooligans**: Elusive. After Reap: Put a paint counter on an artifact. While that artifact has a paint counter, it belongs to house Mars (instead of its other houses).
- [x] (135) **Airlock**: Action: Discard a non-Mars card from your hand. If you do, draw a card.
- [x] (140) **Ironyx Banner**: Play: Make a token creature. Each friendly Ironyx creature gets +1 power and +1 armor.
- [x] (141) **Memrox the Red**: Your opponent`s cards cannot leave your archives. Action: Gain 1 Aember for each card in your archives.
- [x] #3035
- [x] #3036
- [x] (148) **Suspended Animation**: Enhance Damage Damage. Play: Put a damaged enemy creature into your archives. If it leaves your archives, put it into its owner`s hand instead.
- [x] (149) **Yxl the Iron Captain**: Play: Each friendly Ironyx creature captures 2 Aember.
- [x] (150) **Zizok**: Skirmish. Splash-attack 1. After an enemy creature is destroyed by Zizok`s splash-attack damage, ready Zizok.
- [x] (152) **Braindart**: Play: Enrage an enemy creature. It captures 1 Aember from its own side.
- [x] #3037
- [x] (154) **Genetic Blast**: Play: Deal 2 Damage to a creature and each other creature with the same name as that creature.
- [x] (155) **Glylyx Weaponsmith**: Enhance Damage Damage Draw. (These icons have been already added to cards in your deck.)
- [x] (156) **Ironyx Propaganda**: This creature gains, "After Fight/After Reap: Make a token creature."
- [x] (157) **Ironyx Rebel**: Deploy.
[...]

PR #3017, opened 2023-05-26
WoE: implement Revo Hooligans

strib:
Issue: #2903

### DIFF
--- a/server/game/cards/06-WoE/RevoHooligans.js
+++ b/server/game/cards/06-WoE/RevoHooligans.js
@@ -1,0 +1,34 @@
+const Card = require('../../Card.js');
+
+class RevoHooligans extends Card {
+    // After Reap: Put a paint counter on an artifact. While that
+    // artifact has a paint counter, it belongs to house Mars (instead
+    // of its other houses).
+    setupCardAbilities(ability) {
+        this.reap({
+            effect: 'put a paint counter on {0}, making it belong to house Mars',
+            target: {
+                cardType: 'artifact',
+                gameAction: [
+                    ability.actions.addPaintCounter(),
+                    ability.actions.cardLastingEffect((context) => ({
+                        duration: 'lastingEffect',
+                        target: context.target,
+                        until: {
+                            onAddToken: (event) =>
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/RevoHooligans.js
```diff
@@ -0,0 +1,34 @@
...
+    // artifact has a paint counter, it belongs to house Mars (instead
+    // of its other houses).
+    setupCardAbilities(ability) {
+        this.reap({
+            effect: 'put a paint counter on {0}, making it belong to house Mars',
+            target: {
+                cardType: 'artifact',
+                gameAction: [
+                    ability.actions.addPaintCounter(),
+                    ability.actions.cardLastingEffect((context) => ({
```
romanogit, line 14: I think this should be a persistentEffect with a conditional clause.

strib: I'm having trouble figuring out how to give a persistent effect to another card within a game action.  Could you help me figure out what this should look like exactly?  I can't find any examples.

Dockerfile
```diff
@@ -11,6 +11,6 @@ ENV NODE_ENV production
 
 COPY . /usr/src/app
 
-RUN npm run build
+# RUN npm run build
```
romanogit, line 14: ops, again

server/game/cards/06-WoE/RevoHooligans.js
romanogit: but the effect is on itself.

Dockerfile
strib: Sorry, fixed.

server/game/cards/06-WoE/RevoHooligans.js
strib: I don't understand.  Revo is putting an effect on an artifact, via the paint counter.  The effect will survive even if Revo dies.  I don't know of any other card in the game that does this.  Could you tell me which functions to call to make this a persistent effect, rather than the way I coded it?

PR #3023, opened 2023-05-27
WoE: implement Yxl the Iron Captain

strib:
This is the first time multiple creatures are capturing an amount that is more than 1, so it required a change to make sure too much amber wasn't taken from the opponent.  For now the leftover just goes to the last-clicked creature, but maybe we want to improve that interface a bit in the future.

Issue: #2903

### DIFF
--- a/server/game/GameActions/CaptureAction.js
+++ b/server/game/GameActions/CaptureAction.js
@@ -67,7 +67,7 @@ class CaptureAction extends CardAction {
             }
 
             let extra = 0;
-            if (event.amount > 0) {
+            if (amount > 0) {
                 extra = player.sumEffects('captureMoreFromPool');
             }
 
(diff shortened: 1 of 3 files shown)


server/game/GameActions/CaptureAction.js
```diff
@@ -58,16 +58,20 @@ class CaptureAction extends CardAction {
...
             if (!player.anyEffect('captureFromPool')) {
-                player.modifyAmber(-event.amount);
+                if (amount > player.amber) {
+                    amount = player.amber;
+                }
+                player.modifyAmber(-amount);
             }
 
             let extra = 0;
             if (event.amount > 0) {
```
romanogit, line 70: Should be "amount" here `if (amount > 0)`

strib: Done!

PR #3026, opened 2023-05-27
WoE: implement Myx the Tallminded

strib:
Issue: #2903

### DIFF
--- a/server/game/cards/06-WoE/MyxTheTallminded.js
+++ b/server/game/cards/06-WoE/MyxTheTallminded.js
@@ -1,0 +1,22 @@
+const Card = require('../../Card.js');
+
+class MyxTheTallminded extends Card {
+    // Your opponent's keys cost +1 Aember for each friendly Mars
+    // creature in play.
+    setupCardAbilities(ability) {
+        this.persistentEffect({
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/MyxTheTallminded.js
```diff
@@ -0,0 +1,22 @@
...
+
+class MyxTheTallminded extends Card {
+    // Your opponent's keys cost +1 Aember for each friendly Mars
+    // creature in play.
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'opponent',
+            effect: ability.effects.modifyKeyCost(
+                (player) =>
+                    (player.opponent &&
```
romanogit, line 11: What happens if opponent takes control of Myx?
I wonder if "player" here is who owns or who controls the source.

strib: This is all copied from EDAI so it should work fine, but I can add a test.

romanogit: No need,  I tried to find a reference but I could not. Hoefully EDAI also works when taken control

PR #3029, opened 2023-05-27
Six new unfathomable

DylanGranger:
Implementation and testing for the following cards:
Abyssal Sight
Azure Basin Outpost
Brackish Shoreline
Covetous Hema
Dragnet
Skullback Crab


romanogit:
I tick'ed them

PR #3054, opened 2023-05-28
WoE: implement Grondal

### LINKED ISSUE #3033: (121) Grondal: Play: Make 2 token creatures. At the start of your turn, destroy the least powerful creature.

strib:
Issue: #2903
Fixes #3033

### DIFF
--- a/test/server/cards/06-WoE/Grondal.spec.js
+++ b/test/server/cards/06-WoE/Grondal.spec.js
@@ -1,0 +1,37 @@
+describe('Grondal', function () {
+    describe("Grondal's ability", function () {
+        beforeEach(function () {
+            this.setupTest({
+                player1: {
+                    house: 'mars',
+                    token: 'grumpus',
+                    hand: ['grondal'],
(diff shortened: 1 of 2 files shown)


test/server/cards/06-WoE/Grondal.spec.js
```diff
@@ -0,0 +1,37 @@
...
+                player2: {
+                    inPlay: ['urchin', 'dust-pixie']
+                }
+            });
+        });
+
+        it('makes two tokens on play', function () {
+            this.player1.playCreature(this.grondal);
+            this.player1.clickPrompt('Left');
+            //this.player1.clickPrompt('Left');
```
strib, line 20: @romanogit it's not prompting left/right twice when making two tokens.  There might be a bug there but I'm not sure and haven't had time to look into it yet.  It definitely makes two tokens so I'm not sure why it's not prompting for the second one here.

strib: Rebased and fixed.

Issue #2901, opened 2022-11-17
WoE - Brobnar house - Specific issues created for pending cards


romanogit:
This list is still based on Spoilers, so, implement with care:
- [x] (001) **Ancient Battleground**: Each friendly creature cannot reap and gains, "`&#039;&#039;After Fight:&#039;&#039;&#039; Gain 1[[File:Enhance_aember.png|18px|frameless|class=aemberImg|link=|alt=Aember icon]]."
- [x] #3042
- [x] (008) **Frost Giant**: Frost Giant does not ready during your "ready cards" step.
- [x] #3043
your hand. Shuffle your deck. Gain chains equal to the number of cards put into your hand this way.
- [x] (011) **Harmal Atoon**: Each other friendly Brobnar creature gains, "Destroyed:Return this creature to its owner`s hand."
- [x] (013) **Kretchee**: Each time a player exalts a creature or a creature captures Aember, put 1 Aember on that creature from the common supply.     
- [x] #3044
- [x] #3045
- [x] #3046
- [x] (023) **Broken Axe Outpost**: Action: Put a friendly creature on the bottom of its owner`s deck. If you do, deal 6 D to an enemy creature.        
- [x] (024) **Crim Torchtooth**: Splash-attack 4. (When this creature attacks, also deal 4 Damage to each of the attacked creature`s neighbors.) After Fight: Enrage each of Crim Torchtooth&#039;s neighbors.
- [x] #3047
- [x] (028) **Ged Hammer**: Destroyed: Ready and enrage each other friendly Brobnar creature.
- [x] #3048
- [x] (035) **Vow of Blood**: Enhance Damage DamagePlay: Deal 2 Damage to each damaged enemy creature.
- [x] (036) **Vulka**: Each friendly Brobnar creature gains splash-attack 1.
- [x] (038) **Brikk Nastee**: After another friendly Brobnar creature fights, gain 1 Aember.
- [x] (040) **Crogg The Clumsy**: Splash-attack 2. (When this creature attacks, also deal 2 Damage to each of the attacked creature`s neighbors.)       
- [x] (042) **Fresh Marks**: Play: Destroy a friendly creature. If you do, exalt 3 enemy creatures.
- [x] #3049
- [x] (044) **Might Club**: Action: Ready and enrage a creature.
- [x] (045) **Pelf**: Skirmish. (When you use this creature to fight, it is dealt no damage in return.) After Fight: Your opponent loses 1 Aember.      
- [x] #3050
[...]

PR #3062, opened 2023-05-30
WoE: implement Nirbor Flamewing

### LINKED ISSUE #3044: (016) Nirbor Flamewing: At the start of your turn, if Nirbor Flamewing is in your discard, you may destroy a friendly creature. If you do, put Nirbor Flamewing into play ready. Destroyed: Make a token creature.

strib:
Issue: #2901
Fixes #3044

### DIFF
--- a/server/game/cards/06-WoE/NirborFlamewing.js
+++ b/server/game/cards/06-WoE/NirborFlamewing.js
@@ -1,0 +1,104 @@
+const Card = require('../../Card.js');
+const EventRegistrar = require('../../eventregistrar.js');
+
+class NirborFlamewing extends Card {
+    // At the start of your turn, if Nirbor Flamewing is in your
+    // discard pile, you may destroy a friendly creature. If you do,
+    // put Nirbor Flamewing into play ready.
+    //
+    // Destroyed: Make a token creature.
+    setupCardAbilities(ability) {
+        this.leftPlaySinceStartOfRound = false;
+
+        this.tracker = new EventRegistrar(this.game, this);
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/NirborFlamewing.js
```diff
@@ -0,0 +1,76 @@
...
+        });
+
+        // If Nirbor dies during the start of turn pahase (e.g., to
+        // another Nurbor, or by General Order 24, etc), we should have the
+        // option of immediately bringing it back.  But we can't fire
+        // immediately on the destroyed event in the case of 2 Nirbors
+        // because then we won't get to the `then` clause. The
+        // `addEventToWindow` call is trying to get around that but it's
+        // not quite right.
+        this.reaction({
```
strib, line 39: This case is broken for the 2x Nirbor scenario but it might be good enough to go with for now, and we can revisit later.  Or, if you know how to fix it, please let me know!

strib:
This is fixed now, and it was a doozy.  It's the first instance in the game where the player can loop as long as they want during the "begin round" phase and it required some extra stuff and a lot of my brain cells, but I think this should work.  Let me know if you have any questions.

chenj209:
> This is fixed now, and it was a doozy. It's the first instance in the game where the player can loop as long as they want during the "begin round" phase and it required some extra stuff and a lot of my brain cells, but I think this should work. Let me know if you have any questions.

It is still bugged. When the nirbor flamewing dies as a token creatue, the ability to resurrect is not registered properly. I think when a token creatures dies, maybe there should be a check the see if anything on the behind creature should be dealt with

PR #3074, opened 2023-05-31
WoE: implement Gezrahi Blacksmith

strib:
Issue: #2902


romanogit:
What's the best practice for making new localized strings for prompts? Do I just put it in English for now in all the translation files, or is there a better process?

> You can put in all translation files but the text can be in English.

romanogit:
@strib you have to merge these

strib:
Yep fixed now, should pass.

Issue #2905, opened 2022-11-17
WoE - Saurian house - Specific issues created for pending cards


romanogit:
This list is still based on Spoilers, so, implement with care:
- [x] (231) **Aedile Tullia**: Each time you exalt a friendly non-token creature, make a token creature.
- [x] (233) **Aristotlmimus**: After Reap: If Aristotlmimus has a wisom counter, draw 1 card, archive 1 card, and play a card from your archives. Otherwise, you may play a card from your archives.
- [x] (235) **Conscription**: Play: If there are more enemy creatures than friendly creatures, make a number of token creatures equal to the difference.
- [x] (237) **Edict of Nerotaurus**: After a creature reaps, the next creature used this turn cannot reap. After a creature fights, the next creature used this turn cannot fight.
- [x] (238) **Election**: Each creature gains, "After Reap: Put a `for&#039; or &#039;against&#039; counter on Election".  If there are 6 &#039;for&#039; counters on Election, destroy Election and each creature.  If there are 6 &#039;against&#039; counters on Election, destroy each artifact.
- [x] #3139
- [x] (246) **Platopelta**: After Reap: If Platopelta has a wisdom counter, draw 1 card, archive 1 card, and you may put a wisdom counter on Aristotlmimus. Otherwise, archive 1 card.
- [x] (247) **Rampaging Brutodon**: Play: Make a token creature.
In order to use Rampaging Brutodon, you must destroy a friendly creature.
- [x] (250) **Socraterosaurus**: After Reap: Draw 1 card. You may put a wisdom counter on Platopelta.
- [x] (252) **Tectonic Shift**: Play: Divide each player`s battleline in half as evenly as possible without changing the order of creatures. For each battleline, destroy each creature in one of the halves.
- [x] (253) **Training Costs**: You cannot choose to discard Training Costs from your hand. Play: Lose 2 Aember. If you do not, shuffle Training Costs into your deck.
- [x] (254) **Æmbrosia Outpost**: Action: Put a friendly creature on the bottom of its owner`s deck. If you do,  move 1 Aember from a friendly creature 
to your pool.
- [x] (256) **Chaosodon**: Splash attack 3. (When this creature attacks, also deal 3 Damage to each of the attacked creature`s neighbors.) Before Fight: Deal 3 Damage to each of Chaosodon&#039;s neighbors.
- [x] (259) **Hoplite&#039;s Dory**: This creature gets +2 power for each exhausted creature to its left.
[...]

PR #3089, opened 2023-06-01
WoE: implement Conscription

strib:
Issue: #2905

### DIFF
--- a/server/game/cards/06-WoE/Conscription.js
+++ b/server/game/cards/06-WoE/Conscription.js
@@ -6,29 +6,15 @@ class Conscription extends Card {
     // difference.
     setupCardAbilities(ability) {
         this.play({
-            gameAction: ability.actions.conditional({
-                condition: (context) =>
+            gameAction: ability.actions.makeTokenCreature((context) => ({
+                amount:
                     !!context.player.opponent &&
                     context.player.opponent.creaturesInPlay.length >
-                        context.player.creaturesInPlay.length,
-                trueGameAction: ability.actions.makeTokenCreature((context) => ({
-                    amount:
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/Conscription.js
```diff
@@ -0,0 +1,38 @@
...
+                    !!context.player.opponent &&
+                    context.player.opponent.creaturesInPlay.length >
+                        context.player.creaturesInPlay.length,
+                trueGameAction: ability.actions.makeTokenCreature((context) => ({
+                    amount:
+                        context.player.opponent.creaturesInPlay.length -
+                        context.player.creaturesInPlay.length
+                }))
+            }),
+            effect: 'make {1} token creature{2}',
```
romanogit, line 20: I believe this is the default effect msg

strib: Unfortunately the `conditional` action replaces it with `do something conditionally` with no other info.  I didn't see a way to turn that off but let me know if there is.

romanogit: You can remove the condition and make `amount = 0` when the condition is not applicable
```
this.play({
                gameAction: ability.actions.makeTokenCreature((context) => ({
                    amount: !!context.player.opponent &&
                    context.player.opponent.creaturesInPlay.length >
                        context.player.creaturesInPlay.length ? 
                        context.player.opponent.creaturesInPlay.length -
                        context.player.creaturesInPlay.length : 0
                }))
            }),
```

strib: Fixed along with Election, thanks!

PR #3090, opened 2023-06-01
WoE: implement Edict of Nerotaurus

strib:
Issue: #2905


server/game/cards/06-WoE/Conscription.js
```diff
@@ -0,0 +1,38 @@
+const Card = require('../../Card.js');
+
+class Conscription extends Card {
```
romanogit, line 3: dupe?

strib: Weird, I don't know how that happened, sorry.  All fixed now, please take a look.

PR #3091, opened 2023-06-02
WoE: implement Election

strib:
Issue: #2905

### DIFF
--- a/server/game/cards/06-WoE/Election.js
+++ b/server/game/cards/06-WoE/Election.js
@@ -1,0 +1,56 @@
+const Card = require('../../Card.js');
+
+class Election extends Card {
+    // Each creature gains, "After Reap: Put a 'Yea' or 'Nay'
+    // counter on Election".
+    //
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/Election.js
```diff
@@ -0,0 +1,56 @@
...
+                    }
+                }
+            })
+        });
+
+        this.reaction({
+            when: {
+                onAddToken: () => this.tokens.yea === 6 || this.tokens.nay === 6
+            },
+            effect: '{1}',
```
romanogit, line 35: You could simplify here as well:

```javascript
            effect: '{1}',  // no need
            gameAction: ability.actions.destroy((context) => ({
                target: this.tokens.yea === 6 ? 
                   context.game.creaturesInPlay.concat(this) :
                   context.game.cardsInPlay.filter((card) => card.type === 'artifact')
            })
```

strib: Good idea thanks, fixed in #3089.

PR #3093, opened 2023-06-02
WoE: implement Rampaging Brutodon

strib:
Issue: #2905


server/game/costs.js
```diff
@@ -226,6 +228,19 @@ const Costs = {
...
+        canPay: (context) => context.game.activePlayer.creaturesInPlay.length > 0,
+        payEvent: (context) => {
+            let ability = new BaseAbility({
+                target: {
+                    controller: 'self',
+                    type: 'creature'
+                }
+            });
+            ability.resolveTargets(context);
+            return context.game.actions.destroy().getEvent(context.target, context);
```
strib, line 242: I need help with how to make a cost that includes getting a new target.  This does get a target, but it's a later point in the pipeline, so the `context.target` passed here to the destroy event isn't resolved yet.  But I can't figure out how to make a destroyed event with the resolved target.  Any tips?  Or ideas on how to implement this cost in a different way?

romanogit: how is this "cost" paid? You simply destroy a creature? Or is it like, a creature destroyed this turn?

strib: You destroy a creature right before you start using it, just for the purpose of using Brutodon:

image

romanogit: Crazy, need to think about this one.

strib: I did my best on this one, within my limited understanding.  From rules discussions, I've determined that:

* Rampaging Brutodon can destroy itself if desired.
* The targeted creature must be successfully destroyed in order to use Brutodon.
* The targeted creature must be successfully destroyed in order to even _exhaust_ Brutodon.

I think this is the first cost we have where `canPay` would return true but the payment event itself could fail, so that required some extra work and some weird code to propagate the cancellation backwards to the cost event.  If there is a better way I missed, please let me know!

romanogit:
Nice one

PR #3096, opened 2023-06-02
WoE: Implement Cleric

jtrussell:
See #2904

### DIFF
--- a/server/game/cards/06-WoE/Cleric.js
+++ b/server/game/cards/06-WoE/Cleric.js
@@ -1,0 +1,20 @@
+const Card = require('../../Card.js');
+
+class Cleric extends Card {
+    // When cleric enters play, capture 1A
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onCardEntersPlay: (event, context) => event.card === context.source
+            },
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/Cleric.js
```diff
@@ -0,0 +1,20 @@
...
+class Cleric extends Card {
+    // When cleric enters play, capture 1A
+    setupCardAbilities(ability) {
+        this.reaction({
+            when: {
+                onCardEntersPlay: (event, context) => event.card === context.source
+            },
+            gameAction: ability.actions.capture((context) => ({
+                amount: 1,
+                target: context.event.card
```
romanogit, line 12: I think the amber can be captured on any creature, not necessarily Cleric

jtrussell: That would be cool if true! Is there a source for that? I thought creatures that instruct you to capture would always capture onto themselves.

image

strib: Yeah I agree that it has to capture on the new Cleric.  Otherwise it would say "on a friendly creature" or something.

PR #3098, opened 2023-06-02
WoE: implement Training Costs

strib:
Issue: #2905


server/game/cards/06-WoE/TrainingCosts.js
```diff
@@ -0,0 +1,36 @@
+const Card = require('../../Card.js');
+
+class TrainingCosts extends Card {
+    // You cannot choose to discard Training Costs from your hand.
+    //
+    // Play: Lose 2Aember. If you do not, shuffle Training Costs into
+    // your deck.
+    setupCardAbilities(ability) {
+        this.startingAmber = 0;
```
romanogit, line 9: I think you should use the information in the event, which should be the actual amount of amber lost

romanogit: well, I just noticed it is not:
```
getEvent(player, context) {
        return super.createEvent(
            'onModifyAmber',
            { player: player, amount: this.amount, context: context },
            (event) => event.player.modifyAmber(-event.amount)
        );
    }
```

We should probably do here, something like
```
        return super.createEvent(
            'onModifyAmber',
            { player: player, amount: Math.max(player.amber, this.amount), context: context },
            (event) => event.player.modifyAmber(-event.amount)
        );
    }
```

This way you can add a condition to the `then` clause

strib: It needed to be `Math.min`, but yes, that's nice.  Done!  Please take another look.

PR #3103, opened 2023-06-02
Nav Revamp

pxe1833:
Changes to make the navigation more responsive and stylish. Changes the font of all navlinks to match the site logo, adds animations on hover, etc. Language selector has been updated to match, despite not containing links, persay.

NEW:
image

OLD:
image


client/Components/Navigation/LanguageSelector.jsx
```diff
@@ -69,18 +61,26 @@ const LanguageSelector = () => {
...
-            drop='left'
-            onSelect={onLanguageSelect}
-        >
-            {languages.map((language) => (
-                <NavDropdown.Item key={language.value} eventKey={language.value}>
-                    {language.name}
-                </NavDropdown.Item>
-            ))}
-        </NavDropdown>
+        <div className={'navbar-item navbar-link'}>
```
cryogen, line 64: Why have you replaced the react bootstrap control here?

client/Components/Navigation/NavbarDropdown.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Link from './Link';
+
+const NavbarDropdown = ({ title, childLinks, rightSideMenu }) => {
+    return (
+        <div className={'navbar-item navbar-link'}>
```
cryogen, line 7: These should be react bootstrap components

client/Components/Navigation/Navigation.scss
```diff
@@ -4,17 +4,139 @@
...
+    background-size: 0 0.1em;
+    background-position-y: 100%;
+    background-position-x: 0;
+    background-repeat: no-repeat;
+
+    &:hover {
+        cursor: default;
+        /* the only thing worse than this is JavaScript transitions */
+        /* add 24px to the previous level to get the next */
+        .navbar-dropdown.link-1 {
```
cryogen, line 48: You can use scss to loop these, see Card.scss for an example

client/Components/Navigation/ServerStatus.jsx
```diff
@@ -5,69 +5,40 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
...
-                pingClass = 'text-success';
-            } else if (responseTime < 300) {
-                pingClass = 'text-warning';
-            } else {
-                pingClass = 'text-danger';
-            }
+    const pingLevel = `ping-${
+        connected
+            ? responseTime
+                ? (responseTime < 150 && 'good') || (responseTime < 300 && 'mid') || 'bad'
```
cryogen, line 25: not sure why you've changed these from standard bootstrap classes

cryogen (commented):
Not massively keen on the underline on links. I think I'd also prefer to keep the avatar on the profile menu. Other than that looks good, thanks!

pxe1833:
@cryogen So, having gone back and attempted to re-insert the bootstrap controls here, some work, and some do not, with this design. The more important are the `Nav.Link` and `NavDropdown.Item` components, which, as things exist, are wrapped in the `Link` component to make navigation work the way you want (using the `navigate` action defined with Redux) which creates an illegal nested <a> tag. Would a hybrid approach be acceptable, using my custom component(s) where needed to avoid this issue, and the reinstating the Bootstrap components where they do not interfere with proper tag hierarchy?

client/Components/Navigation/ServerStatus.jsx
pxe1833: The standard bootstrap colors looked a bit harsh compared to the mellow pink-ish of the rest of the theme, so I picked colors that signify the same but are a little more tame

cryogen: They're theme colours though, so should be the same as everywhere they're used - if we don't like them then we change them everywhere. I think they're ok though

PR #3104, opened 2023-06-03
WoE: implement Inspiring Oration

strib:
Issue: #2905

### DIFF
--- a/server/game/cards/06-WoE/InspiringOration.js
+++ b/server/game/cards/06-WoE/InspiringOration.js
@@ -1,0 +1,30 @@
+const Card = require('../../Card.js');
+
+class InspiringOration extends Card {
+    // Play: Exalt a friendly creature. Make a token creature for each
+    // Aember on that creature.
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/InspiringOration.js
```diff
@@ -0,0 +1,30 @@
...
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
+                cardType: 'creature',
+                controller: 'self',
+                gameAction: ability.actions.exalt()
+            },
+            then: (preThenContext) => ({
+                alwaysTriggers: true,
+                message: '{0} uses {1} to make {3}{4}',
```
strib, line 15: I couldn't figure out how to avoid duplicating the message here, it seems like the `then` gameAction message is never printed.

romanogit: Yes, the then message is never printed. Thats to avoid poluting the chat log. Usually it should be a single message per effect.

PR #3111, opened 2023-06-03
WoE: Implement Grey Abess

jtrussell:
See #2904

### DIFF
--- a/server/game/cards/06-WoE/GreyAbess.js
+++ b/server/game/cards/06-WoE/GreyAbess.js
@@ -1,0 +1,18 @@
+const Card = require('../../Card.js');
+
+class GreyAbess extends Card {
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/GreyAbess.js
```diff
@@ -0,0 +1,18 @@
+const Card = require('../../Card.js');
+
+class GreyAbess extends Card {
```
romanogit, line 3: Either the comment or the implementation is incorrect

jtrussell: Yeah, comment is a copy/paste relic. Comments are the worst 😅.

If it’s alright I’m happy to put a pic of the card in each PR instead.

jtrussell: Updated 👍

Issue #2906, opened 2022-11-17
WoE - Star Alliance house- Specific issues created for pending cards


romanogit:
This list is still based on Spoilers, so, implement with care:
- [x] (292) **Cultural Exchange**: Play: Your opponent puts each card from their archives into their hand.
- [x] (296) **Mirror Shell**: Play: Make a token creature. This creature gains, "After Fight/After Reap: For the remainder of the turn, each friendly token creature is a copy of this creature."
- [x] (297) **Norbyn&#039;s Cavern**: Each friendly Æmberling gets +3 power and gains skirmish.
- [x] (301) **Pupgrade**: Pupgrade may be played as an upgrade instead of a creature, with the text: "This creature gets +3 power and gains, `Destroyed: Put Pupgrade on the right flank of your battleline as a creature, ready.&#039;"
- [x] (303) **Recruiting Station**: Omni: Choose a house. Discard the top card of your deck. If it belongs to the chosen house, make a token creature.  
- [x] (304) **Shakedown Sullivan**: After Reap: Choose a creature and discard the top card of its controller`s deck. If that card shares a house with the chosen creature, its controller gains 1 Aember. Otherwise, destroy that creature.
- [x] (305) **Stronger Together**: Play: Ready each Star Alliance creature that has 2 non-Star Alliance neighbors of different houses.
- [x] (307) **Temporal Purge**: Play: Flip each token creature faceup. Put each non-creature card flipped this way into its owner`s discard pile.       
- [x] (310) **Badge of Unity**: This creature belongs to house Star Alliance in addition to its other houses.
- [x] (311) **Camaraderie**: Play: Exhaust a friendly Star Alliance creature. If you do, draw 2 cards for each of its non-Star Alliance neighbors.      
- [x] (312) **Collector Boren**: Play: Put an upgrade from your disard pile into your hand. After an upgrade is attached to Collector Boren, ready Collector Boren.
- [x] (314) **FOF Transponder**: This creature gains, "Destroyed: Make a token creature. Attach FOF Transponder to that creature."
- [x] #3147
- [x] #3146
- [x] #3145
- [x] #3144
- [x] (327) **CPO Zytar**: Enhance Damage Draw Draw. (These icons have already been added to cards in your deck.)
- [x] (328) **Echo Reflector**: This creature gains, "Your opponent`s keys cost +3 A."
[...]

PR #3122, opened 2023-06-03
WoE: implement Mirror Shell

strib:
Issue: #2906


romanogit:
We have a problem with `copyCard` that I believe it does not change a creature from token to non-token and vice-versa.

That can be very confusing and could, in my opinion, bring difficulties to rendering. We would need to know the creature is a token for UI, but not a token for engine.

PR #3127, opened 2023-06-04
WoE: implement Pupgrade

strib:
Issue: #2906

### DIFF
--- a/server/game/cards/06-WoE/Pupgrade.js
+++ b/server/game/cards/06-WoE/Pupgrade.js
@@ -2,6 +2,10 @@ const Card = require('../../Card.js');
 const Effects = require('../../effects');
 
 class Pupgrade extends Card {
+    // Pupgrade may be played as an upgrade instead of a creature,
+    // with the text: "This creature gets +3 power and gains,
+    // 'Destroyed: Put Pupgrade on the right flank of your battleline
+    // as a creature, ready.'"
     setupCardAbilities(ability) {
         this.persistentEffect({
             location: 'any',
--- a/server/game/cards/06-WoE/Pupgrade.js
+++ b/server/game/cards/06-WoE/Pupgrade.js
@@ -17,14 +17,25 @@ class Pupgrade extends Card {
                 ability.effects.modifyPower(3),
                 ability.effects.gainAbility('destroyed', {
                     gameAction: [
+                        ability.actions.conditional((context) => ({
+                            condition: context.source.controller !== this.controller,
+                            trueGameAction: ability.actions.cardLastingEffect((context) => ({
+                                duration: 'lastingEffect',
+                                target: this,
+                                effect: [
+                                    ability.effects.takeControl(context.source.controller),
+                                    ability.effects.takeControlOnRight()
+                                ]
+                            }))
+                        })),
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/Pupgrade.js
```diff
@@ -0,0 +1,38 @@
...
+            effect: [
+                ability.effects.modifyPower(3),
+                ability.effects.gainAbility('destroyed', {
+                    gameAction: [
+                        ability.actions.cardLastingEffect({
+                            duration: 'lastingEffect',
+                            target: this,
+                            effect: Effects.changeType('creature')
+                        }),
+                        ability.actions.putIntoPlay(() => ({
```
romanogit, line 21: this can be tricky as this is also not "giving" control and the upgrade may be on an enemy creature.

strib: I don't follow your comment here -- Pupgrade gives the ability to the creature that it's on, so there's no "giving control" problem here that I can see.  (Also I forgot to put the actual card text in this one before, and I just added it now.)  Can you please take another look and/or explain more about the issue you see?

romanogit: > 'Destroyed: Put Pupgrade on the right flank of your battleline as a creature, ready.'"

Imagine you take control of a creature with Upgrade and that creature get's destroyed. Where does Upgrade go? To the right of you battle line, meaning, you take control of Pupgrade

strib: I agree, and I think this implementation will do that.  It's giving the destroyed effect to whoever controls the creature with the upgrade, just like any other upgrade that gives a destroyed effect.  Why do you think it won't work?

strib: Ah I wrote a test for it and I understand now.  The controller in the destroyed effect is all messed up.  Working on it.

strib: Ok it works now.  I'm not sure if it temporarily puts into play on the owner's battleline first before taking control back over the opponent, but I think all the effects should get applied at the same time so it should be fine...

romanogit: Not sure, but I am still a bit skeptical.
Instead of using takeControl, you could use "myControl" on PutIntoPlayAction.
I also think that if you make the changeType and PutIntoPlay sequential, you can avoid the upgrade change done in PutIntoPlayAction as it would become a creature first, just like I did for makeToken

strib: Does `myControl` work for the case where the active player is the opponent of the creature with the upgrade?  Like if the opponent kills your creature on their turn, we don't want Pupgrade to go under their control.  But I don't understand that code very well so not sure.

strib: Unfortunately neither of these suggestions work.  `myControl` and sequential both break the tests.

romanogit:
Used your test cases and a partial Pupgrade implementation.

I pushed in #3242

Hopefully, the MoveUpgradeToFlank will work for 99% of the cases, until we have clarifications about future issues.

PR #3156, opened 2023-06-06
WoE: implement Recorded History

### LINKED ISSUE #3141: (333) Recorded History: Play: Reveal up to 3 cards of different houses from your hand. Archive each card revealed this way.

strib:
Issue: #2906
Fixes #3141

### DIFF
--- a/server/game/cards/06-WoE/RecordedHistory.js
+++ b/server/game/cards/06-WoE/RecordedHistory.js
@@ -1,0 +1,49 @@
+const Card = require('../../Card.js');
+
+class RecordedHistory extends Card {
+    // Play: Reveal up to 3 cards of different houses from your
+    // hand. Archive each card revealed this way.
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
+                optional: true,
+                controller: 'self',
+                location: 'hand',
+                gameAction: ability.actions.archive({ reveal: true })
+            },
+            then: (preThenContext) => ({
+                target: {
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/RecordedHistory.js
```diff
@@ -0,0 +1,49 @@
+const Card = require('../../Card.js');
+
+class RecordedHistory extends Card {
+    // Play: Reveal up to 3 cards of different houses from your
+    // hand. Archive each card revealed this way.
+    setupCardAbilities(ability) {
```
romanogit, line 6: I believe there is a better way to implement this. I will check. The problem with this implementation is, besides verbose, it will archive 1 at a time and who knows in the future there could be an archive interrupt.

strib: Thanks, would love to make this cleaner.

romanogit: Take a look at https://github.com/keyteki/keyteki/pull/3211

strib: Ok thanks, I changed it to use the `BaseCardSelector`.  The only drawback is that the card still looks "selectable" according to the prompts -- it just doesn't do anything when you click it.  Hopefully that's ok. (I assume Hallafest will work the same way.)

romanogit:
Very nice

Issue #2907, opened 2022-11-17
WoE - Unfathomable house


romanogit:
This list is still based on Spoilers, so, implement with care:
- [x] (347) **Brackish Shoreline**: After a creature fights, it doesn`t ready during the "ready cards" step this turn.
- [x] (348) **Bryozoarch**: When your opponent plays an action card, instead of resolving its play effect, destroy the creature on your left flank.     
- [x] (350) **Extrematode Infection**: Play: Put 3 hatch counters on Extrematode Infection.  At the start of your turn, remove 1 hatch counter from Extrematode Infection. Then if it has no hatch counters, destroy this creature and make 3 token creatures.
- [x] (351) **Giltspine School**: Play: Make 3 token creatures. Token creatures do not ready during their controller`s "ready cards" step.
- [x] (352) **Hypoxia**: Play: Destroy each token creature.
- [x] #3175
- [x] (356) **Plunder**: Play: Reveal a random unrevealed card from your opponent`s hand. Then, you may either repeat this effect or your opponent discards the last revealed card.
- [x] (357) **Primordial Vault**: At the start of your turn, make a Cultist. Action: Destroy 4 friendly Cultists. If you do, search your deck and discard pile for Tangaika, reveal it, add it to your hand, and shuffle your deck.
- [x] (359) **Sibyl Waimare**: At the start of your opponent`s turn, that player discards the top card of their deck and exhausts each creature of that 
card&#039;s house.
- [x] (361) **Tangaika**: You cannot play Tanagaika unless you control 4 or more Cultists. Splash-attack 5. After Fight: Gain 1 Æmber.
- [x] (365) **Weak Link**: This creature gains, "While this creature is exhausted, your keys cost +6 Aember."
- [x] (368) **Azure Basin Outpost**: Action: Put a friendly creature on the bottom of its owner`s deck. If you do, exhaust 3 enemy creatures.
- [x] (369) **Catch and Release**: Play: Return each creature to its owner`s hand. Each player discards random cards from their hand until they have 6 or fewer cards in hand. Gain 2 chains.
- [x] (370) **Dragnet**: Play: Return a creature and each neighbor that shares a house with it to their owners` hands.
- [x] (373) **Kelping Hands**: Omni: Destroy Kelping Hands. For the remainder of the turn, each friendly creature gains poison.
- [x] (374) **Physaloha**: Enhance Damage Damage. (These icons have already been added to cards in your deck.) Damaged creatures do not ready during the "ready cards" step.
[...]

PR #3158, opened 2023-06-06
WoE: implement Bryozoarch

strib:
Issue: #2907

### DIFF
--- a/server/game/cards/06-WoE/Bryozoarch.js
+++ b/server/game/cards/06-WoE/Bryozoarch.js
@@ -1,0 +1,25 @@
+const Card = require('../../Card.js');
+
+class Bryozoarch extends Card {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'opponent',
+            match: (card) => card.type === 'action',
+            effect: ability.effects.blank()
+        });
+
+        this.reaction({
+            when: {
+                onCardPlayed: (event, context) =>
+                    event.card.type === 'action' && event.player === context.player.opponent
+            },
+            gameAction: ability.actions.destroy((context) => ({
+                target: context.player.creaturesInPlay[0]
+            }))
+        });
+    }
+}
+
+Bryozoarch.id = 'bryozoarch';
+
+module.exports = Bryozoarch;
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/Bryozoarch.js
```diff
@@ -0,0 +1,25 @@
+const Card = require('../../Card.js');
+
+class Bryozoarch extends Card {
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            targetController: 'opponent',
+            match: (card) => card.type === 'action',
+            effect: ability.effects.blank()
```
romanogit, line 8: I am not sure I like to use blank. I think we should have an effect that says `cannot resolve`

romanogit: For example, how it will interact with "High Priest Torvus" (Thank Lord this card is not in WoE) :)

server/game/cards/06-WoE/Bryozoarch.js
```diff
@@ -0,0 +1,25 @@
...
+        });
+
+        this.reaction({
```
romanogit, line 11: Should be an interrupt here

server/game/cards/06-WoE/Bryozoarch.js
strib: Does the new version look better?  I'm not sure I fully understand how all this works.

romanogit: I guess not, because I believe the interrupt you added you fizzle also the enhancements on the card played

strib: The test makes sure that the opponent gets the amber bonus.  Is the test wrong?

romanogit: ok, so probably the amber is somewhere else. It would be nice to add a test with High Priest Torvus on player1 field. It should not trigger its effect, but here I think the problem is in High Priest Torvus implementation. If it triggers, just `xit` the test case so we know about it in the future.

strib: Ok added the test case, and you're right, it fails.  `xit`'d it out.  Let me know if you want me to make an issue for it.

romanogit: no need, the xit is enough

PR #3173, opened 2023-06-06
WoE: implement Offering to Kiligog

strib:
Issue: #2907

### DIFF
--- a/test/server/cards/06-WoE/OfferingToKiligog.spec.js
+++ b/test/server/cards/06-WoE/OfferingToKiligog.spec.js
@@ -1,0 +1,89 @@
+describe('Offering to Kiligog', function () {
+    describe("Offering to Kiligog's ability", function () {
+        beforeEach(function () {
+            this.setupTest({
+                player1: {
+                    house: 'unfathomable',
+                    token: 'grumpus',
+                    inPlay: ['offering-to-kiligog', 'pelf', 'skullback-crab']
+                },
+                player2: {
+                    inPlay: ['troll']
+                }
+            });
+        });
+
+        it('should destroy a friendly creature and pick same one to put underneath', function () {
+            this.player1.useAction(this.offeringToKiligog, true);
+            this.player1.clickCard(this.pelf);
+            expect(this.player1).not.toBeAbleToSelect(this.skullbackCrab);
+            this.player1.clickCard(this.pelf);
+            expect(this.offeringToKiligog.childCards).toContain(this.pelf);
+            expect(this.player1).toHavePrompt('Choose a card to play, discard or use');
+        });
+
+        it('should destroy a friendly creature and pick a different once to put underneath', function () {
+            this.player1.fightWith(this.skullbackCrab, this.troll);
+            this.player1.useAction(this.offeringToKiligog, true);
+            this.player1.clickCard(this.pelf);
+            this.player1.clickCard(this.skullbackCrab);
+            expect(this.offeringToKiligog.childCards).toContain(this.skullbackCrab);
+            expect(this.pelf.location).toBe('discard');
+            expect(this.player1).toHavePrompt('Choose a card to play, discard or use');
+        });
+
+        it('should place cards out as tokens', function () {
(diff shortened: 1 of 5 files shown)


test/server/cards/06-WoE/OfferingToKiligog.spec.js
```diff
@@ -0,0 +1,89 @@
...
+            this.player1.clickCard(this.skullbackCrab);
+
+            this.player1.endTurn();
+            this.player2.clickPrompt('brobnar');
+            this.player2.endTurn();
+            this.player1.clickPrompt('unfathomable');
+
+            this.player1.clickCard(this.offeringToKiligog);
+            this.player1.clickPrompt("Use this card's Omni ability", 1);
+            //this.player1.clickPrompt('Right');
```
strib, line 56: @romanogit weird that I'm not getting a prompt in this test, but it's working fine in other tests that make multiple tokens.  Maybe needs another look.

romanogit: I believe that's because the play area is empty

romanogit: I think you will need to use sequentialForEach here, because, in theory, the user can choose the order. Take a look at General Sherman

strib: Hmm ok, same with Space Invaders too I guess.  I didn't realize this way wouldn't let you choose the order.

strib: Ok fixed up here and for Space Invaders.  Not sure why it's different from something like Giltspine School that also makes multiple tokens at a time with `makeTokenCreature()` but still gets the placement prompts, but anyway this seems to work.

romanogit: because for Giltspine School you do not choose the cards, you just play from the top of deck

romanogit: I think Space Invaders will just take the orders you chose to reveal the cards. It was correct.

Here we will need a sequentialMakeToken, like we have a sequentialPutIntoPlay (see GeneralSherman).

What you did does not force the selection of the card to put into play.

romanogit: I think you can revert Space Invaders

strib: But `Space Invaders` wasn't prompting for left/right for the second token, so it couldn't have been right before.

I don't understand this issue with what I have now, but I'll see about changing it a bit later tonight.

romanogit: Yes, that's weird. If we make multiple tokens, it should prompt.

romanogit: I think it is not prompting the first time because the play area is empty

strib: That's right, I don't expect it to prompt the first time.  But it should prompt for the second token (which is why there's only one `clickPrompt()` in the test.

romanogit: ok, so why no prompt before, right?

strib: Yeah exactly.  I thought that's what you were trying to help me fix, but I guess it's something else.

romanogit: let's say under OtK there are 3 cards, you must choose which card to put in play faced down. So you select the card, and you select the flank.

For Space Invaders, this is not needed, it will place in play following the order you revealed, which should be correct.

For Space Invaders, it should prompt as many time as needed for flank, though

romanogit:
I will merge an take a look at the implementation

strib:
Ok I got started on it in #3179 but it's not working (and needs some refactoring once it does work).

PR #3178, opened 2023-06-06
Woe friendship

dgbrokaw:
~~Not working yet. The result of the last `it` show that none of the creatures take damage. I think using applyDamage to redirect the divided damage is likely the problem.~~

EDIT: [this commit](https://github.com/keyteki/keyteki/pull/3178/commits/a5527ccd9d0822991d82c11adf27169f84f4f8da) add tokens instead of applying damage, and this allows the neighboring Friendship test to pass.


romanogit:
I will merge and try to think about Drecker and Shadowself interaction later... will work for the majority of the cases

PR #3179, opened 2023-06-06
WoE: rework Offering to Kiligog with sequential plays

strib:
Suggested by romanogit.

Issue: https://github.com/keyteki/keyteki/pull/3173

### DIFF
--- a/server/game/GameActions/SequentialMakeTokenCreatureAction.js
+++ b/server/game/GameActions/SequentialMakeTokenCreatureAction.js
@@ -1,0 +1,111 @@
+const SingleCardSelector = require('../CardSelectors/SingleCardSelector');
+const GameAction = require('./GameAction');
+
+class CardListSelector extends SingleCardSelector {
+    constructor(cardList, revealList = []) {
+        super({
+            cardType: [...new Set(cardList.map((c) => c.type))]
+        });
+        this.cardList = cardList;
+        this.revealList = revealList && revealList.length > 0 ? revealList : cardList;
+    }
+
+    // eslint-disable-next-line no-unused-vars
+    findPossibleCards(context) {
+        return this.revealList;
+    }
+
+    // eslint-disable-next-line no-unused-vars
+    canTarget(card, context) {
+        if (!card) {
+            return false;
+        }
+
+        return this.cardList.includes(card);
+    }
+}
+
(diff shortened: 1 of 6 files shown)


test/server/cards/06-WoE/SpaceInvaders.spec.js
```diff
@@ -43,7 +43,6 @@ describe('Space Invaders', function () {
             this.player1.clickCard(this.zorg);
             this.player1.clickCard(this.pelf);
             this.player1.clickPrompt('Done');
-            this.player1.clickPrompt('Right');
             expect(this.player1.player.creaturesInPlay.length).toBe(2);
             expect(this.zorg.location).toBe('play area');
             expect(this.zorg.name).toBe('Grumpus');
```
strib, line 48: The test fails here because the creatures have their real names, not token names, and I'm not why they aren't getting tokenized...

romanogit: I will check Space Invaders... on it

romanogit: @strib, hopefully https://github.com/keyteki/keyteki/pull/3180 solves the Space Invader's problem (and other cards as well)

alifeee:
Closes #3175

test/server/cards/06-WoE/SpaceInvaders.spec.js
strib: For some reason the sequential action kept getting its properties set to default in between its function calls, so we lost the `cardLocation`.  Passing it around explicitly fixed the problem.  Ready to review!

server/game/GameActions/SequentialMakeTokenCreatureAction.js
```diff
@@ -0,0 +1,111 @@
+const SingleCardSelector = require('../CardSelectors/SingleCardSelector');
+const GameAction = require('./GameAction');
+
+class CardListSelector extends SingleCardSelector {
```
romanogit, line 4: I believe we had a PR that turn this `CardListSelector` inner class into a file itself. I think we should reuse.

strib: Weird, I don't know how that change got lost.  Fixed!

PR #3188, opened 2023-06-07
WoE: implement Plunder

strib:
Issue: #2907

### DIFF
--- a/server/game/cards/06-WoE/Plunder.js
+++ b/server/game/cards/06-WoE/Plunder.js
@@ -1,0 +1,51 @@
+const _ = require('underscore');
+const Card = require('../../Card.js');
+
+class Plunder extends Card {
+    // Play: Reveal a random unrevealed card from your opponent's
+    // hand. Then, you may either repeat this effect or your opponent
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/Plunder.js
```diff
@@ -0,0 +1,51 @@
+const _ = require('underscore');
+const Card = require('../../Card.js');
+
+class Plunder extends Card {
+    // Play: Reveal a random unrevealed card from your opponent's
+    // hand. Then, you may either repeat this effect or your opponent
+    // discards the last revealed card.
+    setupCardAbilities(ability) {
+        this.revealedCards = [];
```
romanogit, line 9: when is this array cleared?

strib: Why does it need to be cleared?  I modeled this on "One Stood Against Many", which never clears its array.

strib: Ohhh wait, yes it does, I see it now.  I'll figure out a fix.

PR #3194, opened 2023-06-07
WoE: implement Sibyl Waimare

strib:
Issue: #2907

### DIFF
--- a/server/game/cards/06-WoE/SibylWaimare.js
+++ b/server/game/cards/06-WoE/SibylWaimare.js
@@ -14,11 +14,10 @@ class SibylWaimare extends Card {
                 target: context.player.opponent.deck[0]
             })),
             then: {
-                gameAction: ability.actions.sequentialForEach((context) => ({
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/SibylWaimare.js
```diff
@@ -0,0 +1,30 @@
...
+            when: {
+                onBeginRound: (_, context) =>
+                    context.player.opponent === this.game.activePlayer &&
+                    context.player.opponent.deck.length > 0
+            },
+            gameAction: ability.actions.discard((context) => ({
+                target: context.player.opponent.deck[0]
+            })),
+            then: {
+                gameAction: ability.actions.sequentialForEach((context) => ({
```
romanogit, line 17: Here should not be sequentialForEach. Exhaust must happen at the same time. Just use exhaust with target

strib: Thanks, didn't realize that could be multiple cards.  Fixed!

PR #3196, opened 2023-06-07
WoE: Implement Halo of Enlightenment

### LINKED ISSUE #3181: (180) Halo of Enlightenment: Play: Make a token creature. This creature cannot be attacked while its controller controls a token creature.

jtrussell:
See #2904
Fixes #3181


jtrussell:
There must be a more direct way of getting at the attached creature's controller's creatures... though when I say it like that it sounds like a mouthful 😂.

PR #3202, opened 2023-06-07
WoE: implement Sunk Cost

strib:
Issue: #2907

### DIFF
--- a/server/game/cards/06-WoE/SunkCost.js
+++ b/server/game/cards/06-WoE/SunkCost.js
@@ -13,7 +13,6 @@ class SunkCost extends Card {
             },
             effect: 'make a token creature if {1} randomly discards a card matching {2}',
             effectArgs: (context) => [context.player.opponent, context.house],
-            effectAlert: true,
             gameAction: ability.actions.discardAtRandom(),
             then: (preThenContext) => ({
                 gameAction: ability.actions.makeTokenCreature((context) => ({
--- a/server/game/cards/06-WoE/SunkCost.js
+++ b/server/game/cards/06-WoE/SunkCost.js
@@ -1,0 +1,32 @@
+const Card = require('../../Card.js');
+
+class SunkCost extends Card {
+    // Play: Choose a house. Your opponent discards a random card from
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/SunkCost.js
```diff
@@ -0,0 +1,32 @@
...
+    setupCardAbilities(ability) {
+        this.play({
+            condition: (context) => !!context.player.opponent,
+            target: {
+                mode: 'house',
+                houses: (context) => context.player.opponent.houses
+            },
+            effect: 'make a token creature if {1} randomly discards a card matching {2}',
+            effectArgs: (context) => [context.player.opponent, context.house],
+            effectAlert: true,
```
romanogit, line 16: I do not think `effectAlert` is needed here

server/game/cards/06-WoE/SunkCost.js
```diff
@@ -0,0 +1,32 @@
...
+            gameAction: ability.actions.discardAtRandom(),
+            then: (preThenContext) => ({
+                gameAction: ability.actions.conditional({
```
romanogit, line 19: Here you can avoid conditional using amount:
```
ability.actions.makeTokenCreature((context) => ({
   amount: context.preThenEvent.cards.length > 0 &&
                        context.preThenEvent.cards[0].hasHouse(preThenContext.house) ? 1 : 0
})
```

strib: Done, thanks.

server/game/cards/06-WoE/SunkCost.js
strib: I disagree -- it is the same type of effect as Allusions of Grandeur, which does have the `effectAlert` (I copied this from there).  But if you have a reason to think we should treat Sunk Cost differently I'm happy to change it.

strib: Oops sorry, I was confusing this one with Watch Your Step.  You're right, I removed it.

PR #3204, opened 2023-06-07
Add some messaging documentation

alifeee:
This was mainly so I (a new contributor) could get my head round a few of the messaging options.

It is not an exhausting list.

However, since it has helped me (a new contributor), I figured it could help other people.


alifeee:
Replaced by #3205

PR #3206, opened 2023-06-07
WoE: implement Knightapult

### LINKED ISSUE #3186: (216) Knightapult: Action: The next time a friendly creature enters play this turn, you may have it enter anywhere in your battleline, ready.

alifeee:
Part of #2904
Closes #3186


alifeee:
This gameAction doesn't work:

```js
                    ability.actions.cardLastingEffect((context) => ({
                        targetLocation: 'play area',
                        target: context.event.card,
                        effect: ability.effects.addKeyword({ deploy: 1 })
                    })),
```

I cannot find an example of a card which makes other cards enter with deploy.

We might have to make a new [effect](https://github.com/keyteki/keyteki/blob/woe/server/game/effects.js#LL37C1-L37C1)?

e.g., `entersPlayAnywhere`

alifeee:
Light Everlasting also plays a creature anywhere.

image

However, I think it works differently, since with Knightapult, the creature is already played? I'm not sure of the timing charts etc.

[Light Everlasting code](https://github.com/keyteki/keyteki/blob/master/server/game/cards/05-DT/LightEverlasting.js#L12)

alifeee:
implemented via Discord suggestion:

https://discord.com/channels/600996321763262464/1112040768895467582/1116115039619928064

alifeee:
now all tests work apart from:

`Knightapult action with token creatures should cause the next token to be deployable`

I guess token creatures work slightly differently?

alifeee:
# attempt 1 - does not work

add ` || event.card.type === 'token creature'` to [this line](https://github.com/alifeee/keyteki/blob/7419ac9a34717007199b1de48ce122c97c2f7351/server/game/cards/06-WoE/Knightapult.js#L10)

```
                until: {
                    onCardEntersPlay: (event) => event.card.type === 'creature'
                },
```

alifeee:
`ability.effects.entersPlayAnywhere` is used in `PlayCreatureAction` (`extends BasePlayAction extends BaseAbility`)

whereas `MakeTokenCreatureAction extends PlayerAction extends GameAction`

I'm lost in code here. I'm not sure where you get the choice of "Play Left/Right" when you play a token creature, so I'm not sure where entersPlayAnywhere needs to be added.

alifeee:
Latest commit broke it. Unsure why

This works:

```js
                until: {
                    onCardEntersPlay: (event) => event.card.type === 'creature'
                },
```

This is broken:

```js
                until: {
                    onCardEntersPlay: (event) =>
                        event.player === this.controller && event.card.type === 'creature'
                },
```

strib:
I made https://github.com/alifeee/keyteki/pull/3 to fix up the token case -- for tokens, the `ability` doesn't satisfy `isCardPlayed()` but we still want to deploy in that case.

I also changed it back to a "player" effect, rather than a "card" effect.  That seems more natural to me and I don't really understand how the card effect works in this case.  But if you want to change it back, that's cool.

Also you might want to add a test where you force your opponent to make a token creature on your turn after using Knightapult (say, kill their Kamalani).  I worry that this would use up the Knightapult effect on the enemy creature when it should not.

alifeee:
thanks! I will do some of that now

alifeee:
okay

1. changed `creaturesEnterPlayReady` back to `entersPlayReady`, as doing so still passes tests, and `entersPlayReady` is already an effect used by many cards.
2. added new tests

Failures:

1) Knightapult action, causing opponent to play a creature with Trojan Sauropod, should still give next friendly creature deploy ❌
2) Knightapult action, causing opponent to play a creature with Trojan Sauropod, should still give next friendly creature ready ❌
3) Knightapult action, causing opponent to play a token creature with Kamalani, should not give the new tokens deploy ❌
4) Knightapult action, causing opponent to play a token creature with Kamalani, should give the next friendly creature ready ❌

strib:
Ok I put up https://github.com/alifeee/keyteki/pull/5 .  I moved it back to a player effect again as the match has to be for a player, not a card.  For some reason it seems to be applying the effect to both players during the turn, hence the need for the `match`.  I don't really understand why.

Another fun test to add would be to somehow play one of your opponents creatures right after your knightapult -- lateral shift maybe?  Knightapult should work in that case, and then be turned off for the next creature.

server/game/GameActions/PutIntoPlayAction.js
```diff
@@ -48,9 +48,11 @@ class PutIntoPlayAction extends CardGameAction {
             let choices = ['Left', 'Right'];
 
             if (
-                context.ability &&
-                context.ability.isCardPlayed() &&
-                (this.deploy || card.hasKeyword('deploy')) &&
+                ((context.ability && context.ability.isCardPlayed()) ||
+                    player.anyEffect('creaturesEnterPlayAnywhere', context)) &&
```
alifeee, line 52: This feels specific. I had a look to see if I could replace it with some check of the event `"onMakeToken"` instead (to be more generic), but I could not figure it out.

For now it works.

alifeee:
@strib thank you again!

Again I changed back to the card action `entersPlayReady`, as it still passes all tests, and reduces the amount of card-specific game actions we have. As you see, I did this by splitting Knightapult's "remainder of turn" actions into two: one for ready; and one for enters play anywhere.

As for your test idea, I had a little think but I could not think of a simple way to implement it. Perhaps that means that it is a rare case that will not happen in-game (haha...). Thus, for now I propose we can skip it.

This card could now be merged 👍

romanogit:
I took you test cases, but I made the effect a "Card" effect, which simplified the implementation a lot. Basically, I merged the two effects in a single array.

PR #3208, opened 2023-06-07
Take control on to swap control

romanogit:


cryogen:
The API of this is a bit weird. Can you use constants instead of -1? or an enum or something?

PR #3209, opened 2023-06-08
WoE: implement Near-Future Lens

### LINKED ISSUE #3170: A12 - Near-Future Lens
Play with the top card of your deck revealed.

Omni: Play the top card of your deck as is it belonged to the active house and was in your hand.

strib:
Basic idea here is to print a message for the new top card every time
the top card of the deck changes, and to always have that top card
marked as not "facedown".  Currently unable to test it in the actual
UI but hopefully it works???

Fixes #3170

### DIFF
--- a/server/game/CardVisibility.js
+++ b/server/game/CardVisibility.js
@@ -28,7 +28,7 @@ class CardVisibility {
     }
 
     isEffectRule(card) {
-        if (card.getEffects('visbileIn').some((effect) => effect === card.location)) {
+        if (card.getEffects('visibleIn').some((effect) => effect === card.location)) {
             return true;
         }
 
--- a/server/game/CardVisibility.js
+++ b/server/game/CardVisibility.js
@@ -28,7 +28,15 @@ class CardVisibility {
     }
 
     isEffectRule(card) {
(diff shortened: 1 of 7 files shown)


alifeee:
This is what it looks like in the feed.

IMO it would be nice if each "[player] reveals [card] on top of their deck" said *why*

i.e., was something like "[player] uses Near-Future Lens to reveal [card]"

image

strib:
@alifeee Done, thanks for the suggestion!

alifeee:
:)

if you want to see the game logs, you can also set the environment variable `DEBUG_TEST=1`, and it will print them to the console.

I believe you can also use toHaveRecentChatMessage to make some sort of test for this

https://github.com/keyteki/keyteki/blob/6b2faaea05980b66755b85f80e01402600e7806d/test/helpers/integrationhelper.js#L152-L169

romanogit:
I do not like the print chat log idea here because it can keep a state that is very helpful.

strib:
It _has_ to be in the logs, I can't see a better choice.  You have to expose each card right before it is drawn.  The alternative is making both the opponent and the player click each time a card is drawn, to make sure they both saw it.

alifeee:
something could be done like with Archivist, right? A UI change?

PR #3211, opened 2023-06-08
add Hallafest

ryanch:
Adding card related to issue:
https://github.com/keyteki/keyteki/issues/3043

Changes:
* new card hallafest
* new test cases for card hallafest
* updated SearchAction event to return cards selected by user
* updated SearchAction and BaseCardSelector to optionally enforce unique card names

Validation done:
* Test case passed
* Lint passed


romanogit:
Nice

PR #3214, opened 2023-06-08
WoE: implement Magistrate Crispus

### LINKED ISSUE #3139: (244) Magistrate Crispus: At the end of your turn, each player takes control of each creature and artifact they own.

strib:
Fixes #3139

### DIFF
--- a/server/game/cards/06-WoE/MagistrateCrispus.js
+++ b/server/game/cards/06-WoE/MagistrateCrispus.js
@@ -1,0 +1,39 @@
+const Card = require('../../Card.js');
+
+class MagistrateCrispus extends Card {
+    // At the end of your turn, each player takes control of each
+    // creature and artifact they own.
+    setupCardAbilities(ability) {
+        this.interrupt({
+            when: {
+                onRoundEnded: (event, context) => context.player === this.game.activePlayer
+            },
+            gameAction: [
+                ability.actions.sequentialCardLastingEffect((context) => ({
+                    forEach: context.game.cardsInPlay.filter(
+                        (card) =>
+                            (card.type === 'creature' || card.type === 'artifact') &&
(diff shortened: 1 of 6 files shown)


server/game/cards/06-WoE/MagistrateCrispus.js
```diff
@@ -0,0 +1,39 @@
...
+class MagistrateCrispus extends Card {
+    // At the end of your turn, each player takes control of each
+    // creature and artifact they own.
+    setupCardAbilities(ability) {
+        this.interrupt({
+            when: {
+                onRoundEnded: (event, context) => context.player === this.game.activePlayer
+            },
+            gameAction: [
+                ability.actions.sequentialCardLastingEffect((context) => ({
```
romanogit, line 12: Active Player has to choose the order, so it needs to prompt first, list sequentialPutIntoPlay

strib: That's why I introduced `sequentialCardLastingEffect`, to prompt for each creature.  (I don't think `sequentialPutIntoPlay` is relevant here.)

Or are you talking about which player's cards get taken back first?  I didn't think that was worth a prompt but can you see a case where it matters?  I think technically you wouldn't pick by player, but card by card (you could mix cards from different players I think).

strib: Yes it prompts card by card (just like General Sherman) for each player.  First the cards you own, then the cards your opponent owns.

When I get some time I can look into making it pool all the card together, rather than player by player.

strib: Ok, it now lets the active player choose the placement order among all the cards from both players.

romanogit:
This seems very good! A difficult implementation.

PR #3221, opened 2023-06-10
Test/splash-attack-on-drecker

### LINKED ISSUE #3218: Test if a splash attack damage also hits Drecker
Test if a splash attack damage also hits Drecker

alifeee:
Closes #3218


romanogit:
Thank you! Excellent

Issue #2908, opened 2022-11-17
WoE mechanics to implement


romanogit:
- [x] Add Ekwidon house game
- [x] Add Ekwidon house UI
- [x] Make token from deck
- [x] Make token from selected cards (hand, for example)
- [x] Flip card down to make a token
- [x] Flip token up
- [x] Divide damage
- [x] Use creature as amber
- [x] Splash-attack
- [x] Archive restrictions
- [x] Take control of each upgrade 
- [x] Swap control
- [ ] #3280
- [x] Fetch from "new" API
- [x] Import Deck with token
- [x] Render token
- [x] Render token's versus card
- [x] Test case with token inPlay from start
- [x] Staff Up versus Ether Spider - choose redirect amber

romanogit:
The Make token from deck needs to be redone

strib:
I think we also need something for Staff Up (https://archonarcana.com/Staff_Up) to let the active player choose if Ether Spider is also in place.  Right now it looks like `gainAmber` uses the most recent effect and isn't built to let players choose.

PR #3222, opened 2023-06-10
Add/Big-Magnet

### LINKED ISSUE #3030: (114) Big Magnet: Play: Chose a friendly creature. Take control of each upgrade in play and move it to that creature.

alifeee:
make Big Magnet work! :)

Part of #2903
closes #3030

#2908 says that we need another effect ("take control of each upgrade"), but I have just used "actions.attach"

### DIFF
--- a/server/game/GameActions/AttachAction.js
+++ b/server/game/GameActions/AttachAction.js
@@ -39,9 +39,23 @@ class AttachAction extends CardGameAction {
     }
 
     getEvent(card, context) {
+        let eventName = 'onCardAttached';
+        if (!!this.upgrade.parent && this.upgrade.parent === card) {
+            // If the parent stays the same, we don't want to fire any
+            // attachment triggers, but we still want to run the event
+            // handler logic below to possibly change control of the
+            // upgrade.
+            eventName = 'onCardAttachedToSameParent';
+        }
         return super.createEvent(
-            'onCardAttached',
-            { card: this.upgrade, parent: card, player: context.player, context: context },
+            eventName,
+            {
+                card: this.upgrade,
+                parent: card,
+                player: context.player,
+                context: context,
+                oldParent: this.upgrade.parent
+            },
             (event) => {
                 if (event.card.location === 'play area') {
                     event.card.parent.removeAttachment(event.card);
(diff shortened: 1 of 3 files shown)


alifeee:
draft: will add tests for:

- Quadracorder - should switch key cost to opponent
- Disruption Field - same thing. should also move counters

romanogit:
>  but I have just used "actions.attach"

more or less, take control of an upgrade would affect, for example, quadracorder, imho.

alifeee:
tests added.

- Disruption Field ✅

- Quadracorder ❌

 Failing tests

- Big Magnet with Quadracorder played onto friendly creature by enemy, after big magnet played, opponent keys should cost 1 more ❌
- Big Magnet with Quadracorder played onto friendly creature by enemy, after big magnet played, friendly keys should be forgeable ❌

alifeee:
I am not sure where to continue from here. Would appreciate some help

romanogit:
First: Attaching an upgrade changes controller to the player attaching, which is great.

Your test for Quadracorder fails due to a very peculiar difference.
Disruption Field is attached to mindworm and you move it to Troll. 
Quadracorder is already attached to Troll and you want to change ownership.

strib:
Nice, I checked and all the tests work if you take out the `card != context.target` clause, making it re-attach the target's upgrades to itself.

I have no clue if you're allowed to "move" an upgrade to its own parent a second time, I'll check.

strib:
Ok, you can definitely re-attach the upgrade to the same creature and cause control to change, so removing `card !== context.target` seems correct.

But, I don't think we can fire `onCardAttached` events for cards that don't move, like for Blasters or for cards next to Dhrxgar.  So we might need a way to supress/interrupt/cancel those events.

strib:
Ok I think https://github.com/alifeee/keyteki/pull/4 should fix it up by not firing `onCardAttached` events when the parent stays the same (but still changing the controller of the upgrade).

alifeee:
> Your test for Quadracorder fails due to a very peculiar difference. Disruption Field is attached to mindworm and you move it to Troll. Quadracorder is already attached to Troll and you want to change ownership.

I almost feel like we should separate out that test into a different test, to check for that case.

alifeee:
okay, I think this is done, if the tests work?

unless anyone can think of another edge case test to make

romanogit:
Thanks @strib  and. @alifeee  - very good team work!!!!

I used your code and tests but created a MoveUpgrade action. All done in #3235

Issue #2911, opened 2022-11-18
WoE - Autogenerated cards that needs review and test cases


romanogit:
Mixes houses:

Implement test cases for:
- [ ] Airlock.js (Mars)
- [ ] Ambitrage.js (Ekwidon)
- [ ] AncientBattleground.js (Brobnar)
- [ ] B0T.js (Star Alliance)
- [ ] Braindart.js (Mars)
- [ ] BrikkNastee.js (Brobnar)
- [ ] CelestialGorm.js (Sanctum)
- [ ] ColonelMariana.js (Sanctum)
- [ ] Cultist.js (Unfathomable)
- [ ] EchoReflector.js (Star Alliance)
- [ ] Fish.js (Unfathomable)
- [ ] ForcedRetirement.js (Ekwidon)
- [ ] FreshMarks.js (Brobnar)
- [ ] GedHammer.jss (Brobnar)
- [ ] GemcoatVendor.js (Ekwidon)
- [ ] GenerousOffer.js (Ekwidon)
- [ ] HeavySubsidies.js (Sanctum)
- [ ] HymnToDuma.js (Sanctum)
- [ ] Kaboom.js (Mars)
- [ ] KelpingHands.js (Unfathomable)
- [ ] MightClub.js (Brobnar)
- [ ] PaxSauriana.js (Saurian)
- [ ] Prospector.js (Ekwidon)
- [ ] PullUpStakes.js (Ekwidon)
- [ ] Rebel.js (Mars)
- [ ] Scholar.js (Saurian)
- [ ] ShiGzisoBuggy.js (Ekwidon)
- [ ] ShiSnyasiBuggy.js (Ekwidon)
- [ ] ToxicudaVenom.js (Unfathomable)
- [ ] TransitoryPhilosopher.js (Ekwidon)
- [ ] Trooper.js (Saurian)
- [ ] YxlixMesmerist.js (Mars)

PR #3229, opened 2023-06-13
add fix Autogenerated cards

ryanch:
Updating cards and tests for:
https://github.com/keyteki/keyteki/issues/2911

Changes:
* fixed Gen Hammer (was only applying effect to itself)
* added tests for Fish, Forced Retirement, Fresh Marks, Ged Hammer

Validation done:
* Test cases passed
* Lint passed

### DIFF
--- a/server/game/cards/06-WoE/GedHammer.js
+++ b/server/game/cards/06-WoE/GedHammer.js
@@ -5,8 +5,24 @@ class GedHammer extends Card {
     setupCardAbilities(ability) {
         this.destroyed({
             gameAction: ability.actions.sequential([
-                ability.actions.ready(),
-                ability.actions.enrage()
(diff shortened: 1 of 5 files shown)


romanogit:
@ryanch : not sure why Fish test failed

alifeee:
This line fails, from `setupTest`

https://github.com/keyteki/keyteki/blob/96237d4b072a7f6a838c83bc33b4350e43965196/test/helpers/playerinteractionwrapper.js#L100-L102

if you replace `fish:flaxia` and `token:fish` with `diplomat:flaxia` and `token:diplomat`, it works.

So.. must be something going wrong with creating a fish token creature specifically.

alifeee:
`fish:flaxia` is not given the `'flipToken'` effect

Issue #3241, opened 2023-06-13
WoE: Tokenized creatures retain aspects when later played as a creature


dmikalova:
In a couple test games it looked like a creature card that was first turned into a token, then destroyed, and then after shuffling the deck and being played, the creature still had some aspects of the token. Examples:

- Faust card was tokenized into a prospector. Prospector is destroyed, discard is shuffled into deck, and then opponent plays faust as faust. When faust was destroyed, the opponent drew a card as if faust had the prospector destroyed effect.
- Fuguru was tokenized into a raider, destroyed, discard is shuffled into deck, and then opponent plays fuguru as fuguru. I played genetic blast against fuguru, and the damage was dealt not just to fuguru, but all raiders as well as if fuguru still had the raider name.

alifeee:
This issue should be reopened - #3287

dmikalova:
I just had a game where a rowdy skald retained a researcher ability

> dmikalova uses Rowdy Skald to archive Ironyx Vatminder

![image](https://github.com/keyteki/keyteki/assets/324548/ee97e9c9-57a6-4d22-9d11-a4c6df0b94e4)

PR #3256, opened 2023-06-15
WoE: token test and fix for token-making prompt image

### LINKED ISSUE #3079: ActivePromptTitle of a token is showing versus card name
ActivePromptTitle of a token is showing versus card name

strib:
Fixes #3079
Adds xit test for #3241


strib:
@romanogit I couldn't figure out the destroyed abilities issue but I added an `xit` test for it.  I tried adding an `until:` to the lasting effect that creatures a token but it didn't seem to do anything.

strib:
The token destroy test now works, given the recent Mimic Gel test and fixes.

PR #3258, opened 2023-06-15
WoE: give Rampaging Brutodon its play ability

strib:
Not sure how I missed this before...


romanogit:
I think there is another way to check for a play ability.

strib:
> I think there is another way to check for a play ability.

I tried `isCardPlayed` but that's for cards being played I think, not play reactions.  If you find anything else, would love to change this since it's pretty clunky.

romanogit:
Does SciOfficerMorpheus help?

romanogit:
Look how it uses ability.properties.name:

```
event.card.abilities.reactions.some(
                        (ability) =>
                            ability.properties.name === 'Play' &&
```

strib:
Thanks, fixed!

PR #3275, opened 2023-06-18
Woe side pane

dgbrokaw:
Adds each player's token to the side pane.

image


dgbrokaw:
Might not need the new player stat

dgbrokaw:
The img key is wrong...they'll be the same if both players have the same token.

cryogen:
keys are only needed if the component is in an array, so you can just remove the keys entirely

dgbrokaw:
Cool, will do.

PR #3287, opened 2023-06-21
Test/strange-shell-gives-artifact-action

alifeee:
When strange shell returns itself to hand, it should lose the "Action" ability. But artifacts still have it.

Related: #3241


romanogit:
"Was token" issue fixed

Issue #3340, opened 2023-06-22
Outnegotiate counting all types of pips for effect


DylanGranger:
![image](https://github.com/keyteki/keyteki/assets/34193643/f1644cbd-f7ca-4b55-95fa-930e33ed531c)

Outnegotiate targeted a Lethologica with its regular amber pip alongside a damage and draw pip, and incorrectly stole 3 amber.

Game log:
Fabulosing brings 嚴酷的月亮戰略家​尼莫 to The Crucible
Fabulousing brings Ypsu the Terribly Avaricious to The Crucible
Fabulosing won the flip and is first player
 Fabulousing has connected to the game server
 Fabulosing has connected to the game server
Fabulosing is shuffling their deck
Fabulosing draws 6 card  s to their maximum of 6
Fabulousing is shuffling their deck
Fabulousing draws 6 card  s to their maximum of 6
Fabulosing draws 1 card 
Key phase - Fabulosing
Fabulosing does not forge a key. They have 0 amber. The current cost is 6 Æmber 
House phase - Fabulosing
Fabulosing chooses untamed  as their active house this turn
Main phase - Fabulosing
Fabulosing plays Commune
Fabulosing uses Commune to lose all their Æmber  and gain 4 Æmber 
Fabulosing played Commune which has Omega, ending this step
Ready phase - Fabulosing
Draw phase - Fabulosing
End of turn 1
Fabulosing: 4 Æmber  (0 keys) Fabulousing: 0 Æmber  (0 keys)
TURN 1 - Fabulousing
Key phase - Fabulousing
Fabulousing does not forge a key. They have 0 amber. The current cost is 6 Æmber 
House phase - Fabulousing
Fabulousing chooses saurian  as their active house this turn
Main phase - Fabulousing
Fabulousing discards Crushing Charge
Fabulousing discards Crushing Charge
Fabulousing plays Imperator Drusilla
Ready phase - Fabulousing
Fabulousing readies their cards 
Draw phase - Fabulousing
Fabulousing draws 3 card  s to their maximum of 6
End of turn 1
Fabulosing: 4 Æmber  (0 keys) Fabulousing: 0 Æmber  (0 keys)
TURN 2 - Fabulosing
Key phase - Fabulosing
Fabulosing does not forge a key. They have 4 amber. The current cost is 6 Æmber 
House phase - Fabulosing
Fabulosing chooses untamed  as their active house this turn
Main phase - Fabulosing
Fabulosing plays Niffle Grounds
[...]

PR #3343, opened 2023-06-22
WoE: Outnegotiate should only steal for amber bonus icons

strib:
Issue: #3340


server/game/GameActions/MakeTokenCreatureAction.js
```diff
@@ -6,6 +6,7 @@ class MakeTokenCreatureAction extends PlayerAction {
         this.deployIndex = undefined;
         this.cards = null;
         this.cardLocation = 'deck';
+        this.alwaysSucceed = false;
```
romanogit, line 9: what is this for? I think you submitted TheVisibleHandle change to Outnegotiate issue. 

Keep this issue to fix only Outnegotiate.

strib: Oh wow, don't know how that happened, sorry. Fixed!

Issue #3339, opened 2023-06-22
Nauxtillian Splash-Attack Damage Not Going Thru Elusive on Blorb


zbarmstrong:
These two decks:

https://www.keyforgegame.com/deck-details/8cff1d99-f2c3-433a-9e11-0b27f0b58c10

https://www.keyforgegame.com/deck-details/718a37f5-7da8-4275-bf2c-440fb3bdca77

Screenshot saved in this PDF:

[Untitled document.pdf](https://github.com/keyteki/keyteki/files/11835110/Untitled.document.pdf)

Green Aeronaut's Action: ability was used to give the friendly Nauxtillian Splash-Attack 3. The Nautxillian was used to fight a Blorb; the attacked blorb's neighbors were not destroyed by Splash-Attack as they should have been. The attacked blorb survived as it should have.

zbarmstrong:
[splash-attack error 2.pdf](https://github.com/keyteki/keyteki/files/11839949/splash-attack.error.2.pdf)

Also happened with 2x Vulka fighting Blorbs. Document with deck links and screenshot attached.

PR #3345, opened 2023-06-22
Splash attack should fire even when fighting an elusive creature

strib:
Issue: #3339


strib:
(Please review this one carefully, I am scared of changing this code.)

strib:
@romanogit updated with your suggestion of a 0-damage event, seems to work but let me know if there are other tests I should add.

PR #3348, opened 2023-06-22
Don't default to facedown=true in RevealAction

### LINKED ISSUE #3316: Cards flip to back and some/all not selectable (poss. after Psychic Bug play with Restingruntus in play...?)
I was in a game (WoE deck vs CotA deck). Restringuntus was in play. CotA player played Psychic Bug. At some point after that, cards on WoE player's deck flipped to back side and were not playable or viewable. Eventually was able to select cards after Restringuntus left play but still not see them in the hand view. See the two attached screenshots.
image
image

strib:
Fixes #3316


strib:
I _think_ this fixes #3316 without affecting Near-Future Lens.  It's not showing the top card in the deck anyway, just in the log, so this shouldn't matter and it's clearly wrong for every other reveal action.

PR #3354, opened 2023-06-22
WoE: Space Invaders should always work, no deck needed

### LINKED ISSUE #3352: Space Invaders Won't Create Tokens from hand if Deck is empty
In bellow example I play space invaders with 3 creatures in hand and no cards in deck. The creatures in hand are not selectable 

image

strib:
Fixes #3352

### DIFF
--- a/server/game/cards/06-WoE/SpaceInvaders.js
+++ b/server/game/cards/06-WoE/SpaceInvaders.js
@@ -16,8 +16,7 @@ class SpaceInvaders extends Card {
                     target: context.player,
                     amount: context.target.length,
                     cards: context.target,
-                    cardLocation: 'hand',
-                    alwaysSucceed: true
+                    cardLocation: 'hand'
                 }))
             },
             effect: 'reveal and tokenize {1}{2} {0}',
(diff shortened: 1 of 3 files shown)


server/game/cards/06-WoE/SpaceInvaders.js
```diff
@@ -16,7 +16,8 @@ class SpaceInvaders extends Card {
                     target: context.player,
                     amount: context.target.length,
                     cards: context.target,
-                    cardLocation: 'hand'
+                    cardLocation: 'hand',
+                    alwaysSucceed: true
```
romanogit, line 20: I think something is wrong here.

strib: `MakeTokenCreature` checks that the deck is not empty, and that check isn't needed for this card.  `alwaysSucceed` was the easiest way to get around that but I'm open to other suggestions for sure.

romanogit: I believe it is the same root cause of https://github.com/keyteki/keyteki/issues/1840

romanogit: disregard the previous comment, I was thinking about TheVisibleHand.
This one probably needs a different GameAction. Let me think about it

romanogit:
Used your test case in https://github.com/keyteki/keyteki/pull/3363

PR #3361, opened 2023-06-23
The visible hand implemented without the "always" flag

romanogit:
@strib, @cryogen 
I implemented the change from "The Visible Hand" without the introduced flag.

I am thinking about Space Invaders


cryogen:
LGTM

PR #3378, opened 2023-06-26
WoE: Symposium should handle tokens that get destroyed on use

strib:
Reported on discord.

### DIFF
--- a/server/game/cards/06-WoE/Symposium.js
+++ b/server/game/cards/06-WoE/Symposium.js
@@ -5,20 +5,25 @@ class Symposium extends Card {
     // token creature, you may exalt, ready and use another friendly
     // creature.
     setupCardAbilities(ability) {
+        this.wasToken = false;
+
         this.play({
             condition: (context) => context.game.creaturesInPlay.length > 0,
             target: {
                 controller: 'self',
                 cardType: 'creature',
                 gameAction: ability.actions.sequential([
-                    ability.actions.exalt(),
+                    ability.actions.exalt((context) => {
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/Symposium.js
```diff
@@ -5,20 +5,25 @@ class Symposium extends Card {
...
+
         this.play({
             condition: (context) => context.game.creaturesInPlay.length > 0,
             target: {
                 controller: 'self',
                 cardType: 'creature',
                 gameAction: ability.actions.sequential([
-                    ability.actions.exalt(),
+                    ability.actions.exalt((context) => {
+                        this.wasToken = context.target.isToken();
```
strib, line 17: I suspect there is a better way of doing this, very open to suggestions!

romanogit: Probably we should use clone and Card's clone should consider tokenCard

strib: That sounds good.  But I tried looking at `context.preThenEvent` in this case, and `card === null` (so there is no clone made).  And I couldn't find any other child events or anything.  In `preThenContext` I can only find a clone of Symposium.  Any ideas about where to find the clone?

romanogit: I fixed createSnapshot to fix the cloned type, but I think I might also need to add a clone for istoken

romanogit:
Used Cyberclone approach in #3496

PR #3523, opened 2023-10-17
Fix tokens and entersPlayReady with Rocketeer Tryska

### LINKED ISSUE #3429: Rocketeer Tryska + Tokens
I controlled Rocketeer Tryska thanks to Talent Scout. I then took the tide with chains and played Visible Hand. Neither of my tokens entered play ready when they were played on either side of her.

romanogit:
Fix #3429

### DIFF
--- a/server/game/Effects/Values/ConditionValue.js
+++ b/server/game/Effects/Values/ConditionValue.js
@@ -1,13 +1,14 @@
 const EffectValue = require('./EffectValue');
 
 class ConditionValue extends EffectValue {
-    constructor(condition) {
+    constructor(condition, target = null) {
         super(condition);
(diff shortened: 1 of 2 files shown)


server/game/Effects/Values/ConditionValue.js
```diff
@@ -1,13 +1,14 @@
...
+    constructor(condition, target = null) {
         super(condition);
         this.condition = condition || (() => true);
+        this.target = target;
     }
 
-    getValue() {
-        return this;
+    getValue(target) {
+        return new ConditionValue(this.condition, target);
```
romanogit, line 11: This can be expensive as we create a new ConditionValue for every target. Maybe just store the target?

PR #3524, opened 2023-10-17
Fix Edict of Nerotaurus when token has an action as flipped card

### LINKED ISSUE #3320: WOE: Edict of Nerotaurus does not allow Reap -> Action -> Reap
Edict of Nerotaurus does not allow Reap -> Action -> Reap
```
After a creature reaps, the next creature used this turn cannot reap.
After a creature fights, the next creature used this turn cannot fight.
```
I thought using a creature for an Action counted as using a creature. The 3rd creature could only fight, not reap. This happened with the Trader token using the Action, so it is also possible that the Action killing the creature might have caused the bug if this otherwise works correctly. Or maybe I misunderstand the rules.

romanogit:
Fixes #3320

### DIFF
--- a/server/game/Events/Event.js
+++ b/server/game/Events/Event.js
@@ -93,10 +93,6 @@ class Event {
             this.cancel();
             return;
         }
-
-        if (this.card) {
-            this.clone = this.card.createSnapshot();
-        }
     }
 
     executeHandler() {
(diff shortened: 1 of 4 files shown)


server/game/Events/Event.js
```diff
@@ -93,10 +93,6 @@ class Event {
             this.cancel();
             return;
         }
-
-        if (this.card) {
```
romanogit, line 97: I am still not comfortable with this change.... There can be a situation, for example, simultaneous resolution, that this change could have broken.
I need to think about a situation and write a TC.

PR #3542, opened 2023-11-25
Add Vietnamese locale

AVAVT:
This is request from Keyforge Vietnam to add Vietnamese translations to the site.


cryogen:
Thanks!

PR #3551, opened 2023-12-23
GR: implement Curse of Fertility

strib:


server/game/cards/07-GR/CurseOfFertility.js
```diff
@@ -0,0 +1,31 @@
...
+            location: 'any',
+            targetController: 'any',
+            effect: [ability.effects.entersPlayUnderOpponentsControl()]
+        });
+
+        this.interrupt({
+            when: {
+                onRoundEnded: (_, context) =>
+                    context.player === this.game.activePlayer &&
+                    !!context.player.opponent &&
```
cryogen, line 19: The !! isn't really necessary in this context. It's only really needed if you need an actual boolean rather than truthiness. I'll merge it as is but just FYI

PR #3553, opened 2023-12-25
Local setup updates

dmikalova:
I updated things that tripped me up when doing my local setup:

- .node-version file so tools like fnm and nvm know which version to use
- Setting the docker build to use the platform linux/amd64 - node installed failed on a Mac M1 without this
- Updated the README
- Added default test users - this saves time for me when wiping the db volume, but if unwanted I'm happy to remove.


.node-version
```diff
@@ -0,0 +1 @@
+v16.20.2
```
cryogen, line 1: I'd rather this target a major version rather than a specific one, but looking into it I'm not sure you can?

server/db/schema/99 - Data.sql
```diff
@@ -62,6 +62,17 @@ INSERT INTO public."Users" ("Id", "Password", "Registered", "Username", "Email",
     (1, '$2b$10$T7eqHoi26C3ADmTDbGOYseTbsrPdCoNFkMKmgh21T4Y6i9NVylgxG', NOW(), 'admin', 'admin@example.com', 'Brobnar', 'normal', False, True, False, True,
      False, '127.0.0.1');
 
+-- Add default test users (NOT FOR PRODUCTION)
+
+INSERT INTO public."Users" ("Id", "Password", "Registered", "Username", "Email", "Settings_Background", "Settings_CardSize",
+    "Settings_OrderAbilities", "Settings_ConfirmOneClick", "Settings_UseHalfSizedCards", "Verified", "Disabled", "RegisterIp") VALUES
+    (2, '$2b$10$T7eqHoi26C3ADmTDbGOYseTbsrPdCoNFkMKmgh21T4Y6i9NVylgxG', NOW(), 'test0', 'test0@gmail.com', 'none', 'x-large', True, True, False, True,
```
cryogen, line 69: I'd rather these default to 'normal' card sizes, since that's the site default

server/db/schema/99 - Data.sql
```diff
@@ -62,6 +62,17 @@ INSERT INTO public."Users" ("Id", "Password", "Registered", "Username", "Email",
...
+     False, '127.0.0.1');
+INSERT INTO public."Users" ("Id", "Password", "Registered", "Username", "Email", "Settings_Background", "Settings_CardSize",
+    "Settings_OrderAbilities", "Settings_ConfirmOneClick", "Settings_UseHalfSizedCards", "Verified", "Disabled", "RegisterIp") VALUES
+    (3, '$2b$10$T7eqHoi26C3ADmTDbGOYseTbsrPdCoNFkMKmgh21T4Y6i9NVylgxG', NOW(), 'test1', 'test1@gmail.com', 'Dis', 'x-large', True, True, False, True,
```
cryogen, line 73: As above

cryogen (changes requested):

.node-version
dmikalova: I changed to v16 and it seems to work for me with fnm

cryogen (approved):

PR #3554, opened 2023-12-25
GR: implement Flying Broomstick

strib:


server/game/Card.js
```diff
@@ -650,6 +650,13 @@ class Card extends EffectSource {
         return !!this.tokens[type];
     }
 
+    sumTokens() {
+        if (!this.tokens) {
+            return 0;
+        }
+        return Object.values(this.tokens).reduce((a, b) => a + b, 0);
```
cryogen, line 657: Should this not just be Object.values(this.tokens).length?

cryogen: Ahha, of course!

PR #3556, opened 2023-12-26
GR: implement Sabira the Medium, Till the Earth and Spoo-key Charge

strib:


server/game/player.js
```diff
@@ -202,6 +202,9 @@ class Player extends GameObject {
             });
             this.addTopCardOfDeckVisibleMessage();
         }
+        if (shuffledDiscardIntoDeck) {
+            this.game.raiseEvent('onDiscardShuffledIntoDeck', { player: this });
```
cryogen, line 206: I don't think this should be a separate event and should probably be a parameter of the shuffle event

cryogen:
I think it would be better to "detect" when you're returning the discard pile to deck, rather than having a new flag

cryogen (changes requested):
I think it would be better to "detect" if the target returned to deck is the discard pile, rather than having a flag for it

strib:
> I think it would be better to "detect" if the target returned to deck is the discard pile, rather than having a flag for it

@cryogen I agree but I couldn't figure out how to do that.  Do you have any ideas?  I thought about just checking if the discard pile is empty but that doesn't work for things that just shuffle the deck when the discard happens to be empty.  Is there a way to see the immediate moveCard events that happened as a predecessor to this shuffle?

server/game/player.js
strib: @cryogen I had two issues with doing it this way:
1. The original shuffle event is emitted _before_ the actual event happens, for some reason.  This trigger needs to happen after.
2. The other event uses `emitEvent`, which seems to make it impossible to listen for because it does some weird thing where it appends `:reaction` to the event name.  `raiseEvent` seems to take care of that.

All that said, nothing seems to use the `onDeckShuffled` aside from tests, at least on the server side, so maybe we could just replace that event completely.

strib:
Here's another wrinkle: An [unofficial ruling](https://discord.com/channels/698635177248948316/753022406255181865/1192313493064532108) on the ABR discord server says that even shuffling a deck after you should have been shuffling a discard pile into the deck, but that discard pile was empty, should still trigger Sabira.  I'm not sure how that can be detected without an explicit flag, but maybe there is a way.

cryogen (approved):

PR #3566, opened 2023-12-27
GR: implement Blernean Hydra and Xenos Darkshadow

strib:


server/game/Effects/DynamicEffect.js
```diff
@@ -15,7 +17,7 @@ class DynamicEffect extends StaticEffect {
     recalculate(target) {
         let oldValue = this.getValue(target);
         this.values[target.uuid] = this.calculate(target, this.context);
-        return oldValue !== this.values[target.uuid];
+        return !_.isEqual(oldValue, this.values[target.uuid]);
```
cryogen, line 20: I'd prefer to reduce underscore usage rather than increasing it, what is this trying to achieve?

strib: This is the first case where the value being used for dynamic recalculation is a dictionary rather than a simple value, and the normal `===` doesn't work for it.  I messed around with trying to figure out whether it's a dictionary and just do a more in-depth comparison in that case but it seems underscore already does that so I didn't want to reinvent.  But let me know if you have other suggestions!

strib: I see, thanks. If that's more important than having less code, I could introduce a new kind of effect builder, like `flexibleProperties` or something, that handles this.

cryogen: I think I understand now. Yes please that would be preferable.

strib: @cryogen I added `flexibleProperties` (also needed for a new revenant.  Please take another look!

cryogen: LGTM, I'll ask @romanogit so have a look if he has capacity

server/game/Effects/DynamicPropertiesEffect.js
```diff
@@ -0,0 +1,45 @@
+const StaticEffect = require('./StaticEffect');
+
+class DynamicPropertiesEffect extends StaticEffect {
```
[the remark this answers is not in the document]
strib: Done, and fixed a logic error in my new code as well, oops.  Please take another look.

PR #3577, opened 2023-12-29
GR: implement Curse of Confusion

strib:


server/game/cards/07-GR/CurseOfConfusion.js
```diff
@@ -0,0 +1,33 @@
+const Card = require('../../Card.js');
+
+class CurseOfConfusion extends Card {
+    // Treachery. (This card enters play under your opponent’s control.)
+    //
+    // At the end of your turn, exhaust each friendly creature of the
+    // active house.
+    setupCardAbilities(ability) {
+        this.persistentEffect({
```
cryogen, line 9: This should be a keyword effect no?

strib: We could implement it that way if you prefer.  Unfortunately the JSON data doesn't appear to contain it as an actual keyword, so we'd still need to have code like this for each card that uses it in order to add the keyword.  That would be a little clearer but seemed a bit harder to get right.  Happy to try it though if you prefer.

cryogen: It does now :)

PR #3579, opened 2023-12-29
GR: implement Strange Torpedo

strib:


server/game/GameActions/DealDamageAction.js
```diff
@@ -156,6 +156,7 @@ class DealDamageAction extends CardGameAction {
             }
 
             let amountApplied = damageDealtEvent.amount - armorUsed;
+            damageDealtEvent.amountApplied = amountApplied;
```
strib, line 159: I couldn't figure out a way to see the `onDamageApplied` event in the `then:` ability of `Strange Torpedo`.  Please let` me know if I'm missing something.

romanogit: event.damageDealtEvent.amountApplied?

strib: The events I'm seeing in the `then:` ability of Strange Torpedo are all `onDamageDealt` events; they don't have a `damageDealtEvent` field.  And as far as I can tell they don't have any child or sub events showing the amount applied.

(To be clear, the code in this PR works just fine; I was just asking if there was a way to do it without adding this new `amountApplied` field here by somehow accessing the `onDamageApplied.amount` field.  But maybe there is no way.)

PR #3603, opened 2023-12-31
GR: implement Jackpot Greene

strib:


server/game/cards/07-GR/JackpotGreene.js
```diff
@@ -0,0 +1,32 @@
...
+                gameAction: ability.actions.discard((context) => ({
+                    target: context.source.controller.deck[0]
+                })),
+                then: {
+                    condition: (context) =>
+                        context.preThenEvents &&
+                        context.preThenEvents.length > 0 &&
+                        context.preThenEvents[0].card.hasHouse('staralliance'),
+                    gameAction: ability.actions.gainAmber({
+                        target: this.controller
```
strib, line 22: I know we're supposed to avoid using `this` but I can't find another way to get the controller of Jackpot Greene inside this reap ability.  Let me know if I'm missing something.

romanogit: `context.source.controller` ?

strib: Unfortunately not.  After the opponent creature reaps, the `context.source` is that creature, not Jackpot Greene anymore.  Not sure if that's a bug in the engine or not but that's how it works in practice. (In fact, in the `discard` action above, I rely on that to discard from the correct deck.)

romanogit: You could try building `then` dynamically :

`then: (preThenContext) => {`

and use `preThenContext.source.controller`

Check the code for examples.

strib: That didn't work; the preTheContext source is also the creature.

But I decided to just turn this whole thing into a reap reaction, rather than a reap ability (which is more correct to the card text anyway), and now it works just fine without using `this`.  Please take another look!

PR #3642, opened 2024-01-03
GR: implement 3LL-I0T

strib:
This card is the first card that forces a specific parent for an upgrade on
play (as opposed to re-attaching an existing upgrade).  I was able
to make this work by allowing a parent to be passed into
`PlayUpgradeAction` from a new `GameAction` but it's not my
favorite code in the world and I'm happy to change it however
anyone else suggests.

### DIFF
--- a/server/game/BaseActions/PlayUpgradeAction.js
+++ b/server/game/BaseActions/PlayUpgradeAction.js
@@ -4,7 +4,7 @@ const CardLastingEffectAction = require('../GameActions/CardLastingEffectAction'
 const Effects = require('../effects');
 
 class PlayUpgradeAction extends BasePlayAction {
-    constructor(card) {
+    constructor(card, parent) {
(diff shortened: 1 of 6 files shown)


cryogen:
Conflicts

server/game/BaseActions/PlayUpgradeAction.js
```diff
@@ -4,13 +4,21 @@ const CardLastingEffectAction = require('../GameActions/CardLastingEffectAction'
 const Effects = require('../effects');
 
 class PlayUpgradeAction extends BasePlayAction {
-    constructor(card) {
+    constructor(card, parent) {
         super(card, {
             activePromptTitle: 'Choose a creature to attach this upgrade to',
             cardType: 'creature',
```
romanogit, line 10: Can you use `target: parent` here? Just curious.

PR #3648, opened 2024-01-04
GR: implement Curse of Spontaneity

strib:


strib:
Updated to use the keyword.

PR #3652, opened 2024-01-04
GR: implement Anachronometer

strib:


server/game/cards/07-GR/Anachronometer.js
```diff
@@ -0,0 +1,47 @@
...
+                        Math.min(context.player.opponent.discard.length, context.player.deck.length)
+                    )
+                }))
+            })
+        ])
+    };
+}
+
+class Anachronometer extends Card {
+    // This creature gains: "After Fight/After Reap: Shuffle your
```
romanogit, line 30: Interesting, is this the first we have a double gain ability when using upgrades?

strib: You mean for both fight and reap?  That's somewhat common, like https://archonarcana.com/Observ-u-Max .

PR #3657, opened 2024-01-05
GR: implement Skreel Swarm

strib:


server/game/cards/07-GR/SkreelSwarm.js
```diff
@@ -0,0 +1,28 @@
...
+            gameAction: ability.actions.dealDamage((context) => ({
+                amount: 1,
+                target: context.game.creaturesInPlay.filter(
+                    (card) => card.controller !== context.player
+                )
+            }))
+        });
+
+        this.scrap({
+            gameAction: ability.actions.archive((context) => ({
```
romanogit, line 19: it is an optional ability, is says "you may", right?

romanogit (changes requested):

server/game/cards/07-GR/SkreelSwarm.js
strib: Oops, great catch.  Done.

PR #3658, opened 2024-01-05
GR: implement Soul Exchange

strib:


server/game/cards/07-GR/SoulExchange.js
```diff
@@ -0,0 +1,56 @@
+const Card = require('../../Card.js');
+
+function firstCreatureInDiscard(context) {
```
romanogit, line 3: Can you rewrite like this to avoid the double filter repetition?

```
function firstCreatureInDiscard(context) {
    const player = !context.select || context.select === 'Mine'
        ? context.player : context.player.opponent;
    return player.discard.find((c) => c.type === 'creature')
```

strib: Done, thanks.

PR #3661, opened 2024-01-07
GR: implement Bonded Auctioneer

strib:


server/game/cards/07-GR/BondedAuctioneer.js
```diff
@@ -0,0 +1,37 @@
...
+            }
+        });
+
+        this.scrap({
+            effect: "return {0} to its owner's hand",
+            effectArgs: (context) => [context.target],
+            target: {
+                cardType: 'artifact',
+                location: 'play area',
+                gameAction: [ability.actions.returnToHand()]
```
romanogit, line 29: I believe you can avoid the array and the `effect` property on line 24.

server/game/cards/07-GR/BondedAuctioneer.js
```diff
@@ -0,0 +1,37 @@
...
+    //
+    // Scrap: Return an artifact to its owner’s hand.
+    setupCardAbilities(ability) {
+        this.reap({
+            effect: 'destroy {0} and have {1} gain 1 amber',
+            effectArgs: (context) => [context.target.controller],
+            target: {
+                cardType: 'artifact',
+                location: 'play area',
+                gameAction: [ability.actions.destroy()]
```
romanogit, line 14: here, you can avoid the array as well.

server/game/cards/07-GR/BondedAuctioneer.js
strib: Thanks, done.

server/game/cards/07-GR/BondedAuctioneer.js
strib: Done.

PR #3665, opened 2024-01-08
GR: implement Market Fluctuation

strib:


server/game/cards/07-GR/MarketFluctuation.js
```diff
@@ -0,0 +1,27 @@
+const Card = require('../../Card.js');
+
+class MarketFluctuation extends Card {
+    // Play: Each haunted player draws 3 cards. Each player that is
+    // not haunted discards 3 random cards from their hand.
+    setupCardAbilities(ability) {
+        this.play({
+            gameAction: ability.actions.draw((context) => ({
```
romanogit, line 8: I think you could use `ability.actions.sequential` instead of `then` with `allwaysTriggers`. Anyway, it has the same result.

strib: Done, thanks.

PR #3670, opened 2024-01-08
GR: implement Technorachnophobia

strib:


server/game/cards/07-GR/Technorachnophobia.js
```diff
@@ -0,0 +1,34 @@
...
+        this.play({
+            gameAction: ability.actions.discard((context) => ({
+                target: context.player.deck.slice(0, Math.min(context.player.deck.length, 10))
+            })),
+            then: {
+                condition: (context) =>
+                    context.preThenEvents &&
+                    context.preThenEvents
+                        .flatMap((event) => event.card.getHouses())
+                        .filter((h, i, a) => a.indexOf(h) === i) // get the list of unique houses
```
romanogit, line 16: I wonder if we can do it in less lines:

context.preThenEvents
  .flatMap((event) => event.card.getHouses())
  .reduce((obj, str) => (obj[str] = (obj[str] || 0) + 1, obj[str] >= 5) ? true : obj, {}) === true

ChatGPT (TM)

strib: The code didn't work but I tweaked it a little to make it work. I'm not sure it's more readable this way but I'm fine either way.  Please take another look.

PR #3671, opened 2024-01-08
GR: implement De-escalation

strib:


server/game/cards/07-GR/DeEscalation.js
```diff
@@ -0,0 +1,34 @@
...
+        this.play({
+            effect: 'destroy all creatures{1}',
+            effectArgs: (context) => [
+                context.player.opponent
+                    ? ' and have ' +
+                      context.player.opponent.name +
+                      ' archive the top 3 cards of their deck'
+                    : ''
+            ],
+            gameAction: [
```
romanogit, line 16: this one needs to be sequential. If opponent has a Rad Penny, for example, they must be able to archive it.

romanogit (changes requested):

server/game/cards/07-GR/DeEscalation.js
strib: Good catch!  Fixed, thanks.

server/game/cards/07-GR/DeEscalation.js
```diff
@@ -0,0 +1,34 @@
...
+            effectArgs: (context) => [
+                context.player.opponent
+                    ? ' and have ' +
+                      context.player.opponent.name +
+                      ' archive the top 3 cards of their deck'
+                    : ''
+            ],
+            gameAction: ability.actions.sequential([
+                ability.actions.destroy((context) => ({ target: context.game.creaturesInPlay })),
+                ability.actions.conditional((context) => ({
```
romanogit, line 18: I do not think you need to add conditional here.

```javascript
                ability.actions.destroy((context) => ({ target: context.game.creaturesInPlay })),
                ability.actions.archive({
                   target: context.player.opponent ? context.player.opponent.deck.slice(0, 3) : []
                })
])
```

server/game/cards/07-GR/DeEscalation.js
```diff
@@ -0,0 +1,34 @@
...
+                    condition: context.player.opponent && context.player.opponent.deck.length > 0,
+                    trueGameAction: ability.actions.archive({
+                        target: context.player.opponent.deck.slice(
+                            0,
+                            Math.min(context.player.deck.length, 3)
```
romanogit, line 23: Should be context.player.opponent.deck.length

No need to worry as slice() returns less than 3 if array is small.
```
> [1,2].slice(0,3)
[ 1, 2 ]
> 
```

romanogit (changes requested):

romanogit (changes requested):
I suggested some simplifications.

server/game/cards/07-GR/DeEscalation.js
strib: Done, thanks.

server/game/cards/07-GR/DeEscalation.js
strib: Done, thanks.

romanogit (approved):

PR #3690, opened 2024-01-11
GR: implement Returning Champion

strib:


server/game/cards/07-GR/ReturningChampion.js
```diff
@@ -0,0 +1,40 @@
...
+                target: context.source,
+                amount: 3
+            }))
+        });
+
+        this.fight({
+            condition: (context) =>
+                context.source.controller.isHaunted() && context.source.hasToken('damage'),
+            target: {
+                activePromptTitle: 'Choose a creature to deal damage to',
```
romanogit, line 20: I think "deal damage to" can be confusing. Maybe just say "move damage to"

server/game/cards/07-GR/ReturningChampion.js
```diff
@@ -0,0 +1,40 @@
...
+                cardCondition: (card, context) => card !== context.source,
+                cardType: 'creature',
+                gameAction: [
+                    ability.actions.addDamageToken((context) => ({
+                        target: context.target,
+                        amount: context.source.tokens.damage
+                    })),
+                    ability.actions.heal((context) => ({
```
romanogit, line 28: I would not use `heal` here as healing could have intercepts in the future. 
Maybe create (if not existing) a `removeDamageToken`

server/game/cards/07-GR/ReturningChampion.js
strib: Done.

server/game/cards/07-GR/ReturningChampion.js
strib: Done.

PR #3691, opened 2024-01-11
GR: implement Rorag Screamer

strib:


test/server/cards/07-GR/RoragScreamer.spec.js
```diff
@@ -0,0 +1,51 @@
...
+            this.player1.fightWith(this.roragScreamer, this.dewFaerie);
+            expect(this.roragScreamer.exhausted).toBe(false);
+            this.player1.fightWith(this.roragScreamer, this.dewFaerie);
+            expect(this.roragScreamer.exhausted).toBe(false);
+            this.player1.fightWith(this.roragScreamer, this.cornicenOctavia);
+            expect(this.roragScreamer.location).toBe('discard');
+        });
+
+        it('can make a creature lose elusive for the turn', function () {
+            this.player1.clickCard(this.roragScreamer);
```
romanogit, line 31: as a suggestion, add a `scrap` method, this will make development simpler.

test/server/cards/07-GR/RoragScreamer.spec.js
```diff
@@ -0,0 +1,51 @@
...
+                },
+                player2: {
+                    inPlay: ['dew-faerie', 'cornicen-octavia']
+                }
+            });
+        });
+
+        it('readies after fighting', function () {
+            this.player1.playCreature(this.roragScreamer);
+            this.player1.endTurn();
```
romanogit, line 18: you can do `this.roragScreamer.exhausted = false` to avoid going through so much code.

romanogit (changes requested):

test/server/cards/07-GR/RoragScreamer.spec.js
strib: Good idea, added and updated all existing tests,

test/server/cards/07-GR/RoragScreamer.spec.js
strib: I knew about this but I'm going through the turn stuff on purpose because it feels more realistic to me and I don't want to accidentally miss anything important that happens when readying.  I prefer this way but if you feel strongly about it I can change it.

test/helpers/playerinteractionwrapper.js
```diff
@@ -634,6 +634,11 @@ class PlayerInteractionWrapper {
         return card;
     }
 
+    scrap(card) {
```
romanogit, line 637: A minor thing: I would have called it "discard". Scrap is the effect from discarding a card. I think "discard" reflects the action the player is doing better.

strib: I agree, but you're the one who suggested calling it `scrap` in your original comment!  Oh well.  I might make an MR later cleaning some of this testing stuff up.

romanogit: oh man 🤦 sorry abou that 😞

PR #3695, opened 2024-01-12
GR: implement Fight to the End

strib:


romanogit:
There is a conflict here to be resolved.

strib:
Thanks, conflict fixed.

PR #3710, opened 2024-01-13
GR: implement Soul Bomb

strib:


server/game/cards/07-GR/SoulBomb.js
```diff
@@ -0,0 +1,20 @@
+const Card = require('../../Card.js');
+
+class SoulBomb extends Card {
+    // Omni: If you are haunted, deal 4 to each creature. Destroy Soul Bomb.
+    setupCardAbilities(ability) {
+        this.omni({
+            gameAction: [
```
romanogit, line 7: this needs to be sequential. Probably won't make much difference, unless you have crazy combinations like:

you deal 4 damage to your jargogle which plays barehanded which is placed under it and you eventually put `Soul Bomb` at the top of your deck instead of destroying it.

Zero chance of happening, but just to show why sequential matters ;)

strib: Thanks!  Done.

PR #3720, opened 2024-01-14
GR: implement Dark Memento

strib:


server/game/cards/07-GR/DarkMemento.js
```diff
@@ -0,0 +1,65 @@
...
+        this.reaction({
+            when: {
+                onBeginRound: (_, context) =>
+                    context.player === this.game.activePlayer && !context.player.isHaunted()
+            },
+            condition: (context) => context.player.deck.length > 0,
+            gameAction: ability.actions.discard((context) => ({
+                target: context.player.deck[0]
+            })),
+            then: (preThenContext) => ({
```
romanogit, line 19: Do we need to use this recursion? 
Can't we discard as much as needed, like:
```
 gameAction: ability.actions.discard((context) => ({
    target: context.player.deck.slice(0, 0, 10 - content.player.discard.length)
}))
```            
wdyt?

romanogit (changes requested):

server/game/cards/07-GR/DarkMemento.js
strib: I was worried there could be interruptions on each individual discard that could end up canceling the rest, or moving the discarded card to a different zone (like Bottomless Depths does for cards discarded from your hand).  If you're not worried about that happening in future cards we can simplify.

strib: Also, the "10" doesn't work because of that other card that changes how haunted is calculated.

romanogit: > Also, the "10" doesn't work because of that other card that changes how haunted is calculated.

ouch ouch ouch :( So true.

PR #3725, opened 2024-01-14
GR: implement Kaspara

strib:


server/game/cards/07-GR/Kaspara.js
```diff
@@ -0,0 +1,32 @@
...
+    setupCardAbilities(ability) {
+        this.persistentEffect({
+            effect: ability.effects.modifyPower(
+                (card) =>
+                    card.controller.amber +
+                    (card.controller.opponent ? card.controller.opponent.amber : 0)
+            )
+        });
+
+        this.persistentEffect({
```
romanogit, line 16: Each friendly create needs to gain a. "Play" effect as well.

romanogit (changes requested):

server/game/cards/07-GR/Kaspara.js
strib: Good catch, done, thanks.

PR #3730, opened 2024-01-15
GR: implement Scooboo

strib:


server/game/Card.js
```diff
@@ -606,9 +607,9 @@ class Card extends EffectSource {
 
         _.each(this.getPersistentEffects(true), (effect) => {
             if (effect.location !== 'any') {
-                if (to === 'play area' && from !== 'play area') {
```
romanogit, line 609: This change is in a neurological part of the code

strib: Understood, but I needed a way to trigger effect recalculation on a move from discard to hand, where the play area is not involved at all.  Happy to do it a different way if you have any recommendations.

romanogit:
Can you split this PR ? One for Scooboo and another for Singing Scythe?

strib:
@romanogit yes they both apply persistent effects while in the discard, so they both need the code change in `CardEffect.js`.  I can just duplicate that change across two different PRs if that's your preference.  Can't do it right this minute but will try to get to it a bit later.

strib:
I was wrong about the changes to `CardEffect.js` -- turns out those weren't needed for either card.  But some of the `Card.js` changes were needed for both.  I duplicated them and put up #3823 for Singing Scythe.  Please take a look!

PR #3732, opened 2024-01-15
GR: implement the clock suite

strib:
Tick-Tock, Chronometer, Ancestral Timekeeper and Synchronizer.

### DIFF
--- a/server/game/cards/07-GR/Synchronizer.js
+++ b/server/game/cards/07-GR/Synchronizer.js
@@ -1,0 +1,41 @@
+const Card = require('../../Card.js');
+
+class Synchronizer extends Card {
+    // Each friendly Clock creature gains, “Destroyed: Move half the time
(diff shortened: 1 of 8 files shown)


server/game/cards/07-GR/Synchronizer.js
```diff
@@ -0,0 +1,41 @@
...
+                controller: 'self',
+                cardType: 'creature',
+                cardCondition: (card) => card.hasTrait('clock'),
+                gameAction: [
+                    ability.actions.addTimeCounter((context) => ({
+                        amount: context.source.tokens['time']
+                    })),
+                    ability.actions.removeTimeCounter((context) => ({
+                        target: context.source,
+                        amount: context.source.tokens['time']
```
romanogit, line 31: minor: you could use `all: true` here

PR #3757, opened 2024-01-17
GR: implement In Here Somewhere...

strib:


server/game/cards/07-GR/InHereSomewhere.js
```diff
@@ -0,0 +1,39 @@
...
+class InHereSomewhere extends Card {
+    // Play: If you are haunted, archive 2 cards from your discard pile.
+    // Otherwise, discard the top 5 cards of your deck.
+    setupCardAbilities(ability) {
+        this.play({
+            gameAction: ability.actions.conditional({
+                condition: (context) => !context.player.isHaunted(),
+                trueGameAction: ability.actions.discard((context) => ({
+                    target: context.player.deck.slice(0, 5)
+                }))
```
romanogit, line 12: can you use a `falseGameAction: archive` with a `promptForSelect` ?

strib: Good idea, I didn't know that would work.  Actually, it didn't work without a fix to `ConditionalGameAction` but it should work now, thanks!

PR #3760, opened 2024-01-18
GR: implement Reaver

strib:


server/game/cards/07-GR/Reaver.js
```diff
@@ -0,0 +1,33 @@
+const Card = require('../../Card.js');
+
+class Reaver extends Card {
+    // After Reap: Move 1 A from a friendly creature to your pool. If you do,
+    // discard a card.
+    setupCardAbilities(ability) {
+        this.reap({
+            target: {
+                activePromptTitle: 'Choose a captured amber to move to your pool.',
+                cardCondition: (card) => card.hasToken('amber'),
```
romanogit, line 10: that's not needed. In theory, I could select a card without amber and move 0 amber.

strib: It's the same wording as Selwyn the Fence from CotA, so I copied the code from there.  But you're right that later cards like Chant of Hubris and Consul Primus let you pick creatures with no amber so I guess that's a bug with Selwyn?

strib: Ok I updated it.

PR #3763, opened 2024-01-18
GR: implement Webstur

strib:


server/game/GameActions/ChooseGameAction.js
```diff
@@ -45,8 +45,10 @@ class ChooseGameAction extends GameAction {
         super.preEventHandler(context);
         let activePromptTitle = this.activePromptTitle;
         let choices = Object.keys(this.choices);
-        choices = choices.filter((key) =>
-            this.choices[key].some((action) => action.hasLegalTarget(context))
+        choices = choices.filter(
+            (key) =>
+                this.choices[key].some((action) => action.hasLegalTarget(context)) ||
+                this.choices[key].length === 0
```
romanogit, line 51: this could have implications... I will analyze more carefully.

strib: This seemed safe since we don't use `chooseAction` anywhere else as far as I can tell.  The other option I thought of was making a `NullAction`/`NoopAction` but that seemed lame.

romanogit: I did not understand why you had to add this check.

strib: @romanogit it is to allow choices that don't have any actions in their handler.  I couldn't find any other example of a choice that did nothing when you clicked on it.  Perhaps there was a better way to implement an action that allows a choice like that?

server/game/cards/07-GR/Webstur.js
```diff
@@ -0,0 +1,36 @@
+const Card = require('../../Card.js');
+
+class Webstur extends Card {
+    // After Fight: For each damage on Webstur, you may discard the top card
+    // of each player's deck.
+    setupCardAbilities(ability) {
+        this.fight({
+            gameAction: ability.actions.sequentialForEach((context) => ({
+                num: context.source.tokens.damage,
```
romanogit, line 9: Let's say Webstur has 3 damage, but we only want to discard one card, we have to click No twice, right?

I wonder if it is better to provide option:`Discard` and `Stop`, instead of no.

strib: > Let's say Webstur has 3 damage, but we only want to discard one card, we have to click No twice, right?

Correct.  That is abiding by a literal reading of the card ability.

> I wonder if it is better to provide option:Discard and Stop, instead of no.

Sure, that would be a valid way to do it too I think.  If you prefer that way, let me know and I'll try it.

romanogit: Not sure we would implement the "Stop" part, but I think it is better.

strib: @romanogit please take another look.  I tried using a 'may' ability instead and tracking the number of iterations in a variable.  When the user chooses "No", it stops.

romanogit: You rock! 🌟

PR #3788, opened 2024-01-20
GR: implement Immortal Greking

strib:


server/game/cards/07-GR/ImmortalGreking.js
```diff
@@ -0,0 +1,47 @@
...
+        this.play({
+            target: {
+                cardType: 'creature',
+                controller: 'opponent',
+                gameAction: ability.actions.sequential([
+                    ability.actions.cardLastingEffect((context) => ({
+                        duration: 'lastingEffect',
+                        effect: [
+                            ability.effects.takeControl(context.source.controller),
+                            ability.effects.takeControlOnRight()
```
strib, line 19: I know putting it on the right first and then moving it in a separate action isn't perfect, but it didn't seem worth complicating the `takeControl` logic for deploying.  I can give it a try if you all feel strongly about it though.

PR #3843, opened 2024-02-07
Handle upgrades, child cards, and leave play events on placeUnder()

strib:
Specifically for Replay Pod and Plan 10, cards in play should lose their upgrades, lose their child cards, and triggers any "leaves play" abilities.

### DIFF
--- a/server/game/Card.js
+++ b/server/game/Card.js
@@ -585,6 +585,21 @@ class Card extends EffectSource {
         this.endRound();
     }
 
+    clearDependentCards() {
+        for (let upgrade of this.upgrades) {
+            upgrade.onLeavesPlay();
+            upgrade.owner.moveCard(upgrade, 'discard');
+        }
+
+        for (let child of this.childCards) {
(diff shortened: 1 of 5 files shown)


server/game/Card.js
```diff
@@ -584,6 +584,21 @@ class Card extends EffectSource {
...
+            upgrade.onLeavesPlay();
+            upgrade.owner.moveCard(upgrade, 'discard');
+        }
+
+        for (let child of this.childCards) {
+            child.onLeavesPlay();
+            child.owner.moveCard(child, 'discard');
+        }
+
+        this.purgedCards.forEach((c) => (c.purgedBy = null));
```
cryogen, line 598: much prefer if this for matched the ones above :)

(appreciate you'#re just copy/pasting but I've seen it now :D )

cryogen (changes requested):

server/game/Card.js
strib: Done.

PR #3995, opened 2024-09-11
database migration for aember skies & corresponding data inserts

dgbrokaw:
Are we spelling it Aember Skies or Aember Skies in the database? ---> I checked the json, it's Æmber Skies


cryogen:
It's fine to use the weird character in text

PR #4006, opened 2024-09-12
AS: Harmonic Pack (Mars) incomplete, 3rd test fails

dgbrokaw:
I don't understand why the third test fails. I thought `alwaysTriggers` would do it.

### DIFF
--- a/server/game/cards/08-AS/HarmonicPack.js
+++ b/server/game/cards/08-AS/HarmonicPack.js
@@ -1,0 +1,41 @@
+const Card = require('../../Card.js');
+
+class HarmonicPack extends Card {
+    setupCardAbilities(ability) {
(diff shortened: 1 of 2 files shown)


strib:
You have a typo there: `alwaysTrigger` instead of `alwaysTriggers`.

Also, it would be useful if you put the text of the ability in a comment above `setupCardAbilities()`, so it's quick to look up what the card is supposed to do.  Thanks!

dgbrokaw:
Awesome, thanks! I have the card text for all the other ones, but was getting a weird error about unusual whitespace for this one, so I took it out just to get the commit out there.

strib:
I made a new PR based on this one that finally does it, I think: #4157 .

PR #4171, opened 2024-09-23
AS: implement Tangrant (Mars)

dgbrokaw:
I think this is literally Mars Witch of the Wilds in terms of effects. So, I was lazy and didn't add tests.


cryogen:
Please can we have tests? Even if they're a copy of the card you've copied the implementation of. The test coverage is pretty good and I'd hate to see it start to reduce

dgbrokaw:
OK, I will do it soon

Issue #4190, opened 2024-10-03
Reap the Wild Wind reveals twice


JareCoder:
![image](https://github.com/user-attachments/assets/c3b800f2-d281-4b17-80c5-2c5684a838b2)

Reap the Wild Wind runs twice as seen in the image. First in the "Uses" statement and then twice after. Only the last 2 revealed cards are used for the amber gain.

strib:
Well at least it's symmetric for both players.  Fixed in #4191 I think. Thanks!

PR #4191, opened 2024-10-03
AS: fix Reap the Wild Wind so it doesn't log random cards

strib:
I guess the default log message re-evaluates the randomness of the revealed cards when constructing the log message. Who knew?

Closes: #4190


cryogen:
It's because the effects are called in multiple passes, so it gets a new one each time

PR #4422, opened 2025-02-15
Add support for Discovery (with icon)

strib:
There are no new cards in this set.


strib:
Shout out to duk for the icon!

Issue #3947, opened 2024-05-26
Saur About That awards aember before resolving "Destroyed:" abilities


Tutenkharnage:
Opponent has Keyfrog and Myliobe on board, no keys, and 5 aember in pool. No other creatures are in play. No other cards that would give aember are in play. I played Saury About That and destroyed Keyfrog. Saury About That's "Play:" ability reads:

_**Play:** Destroy a creature. Its controller gains 1 aember._

Keyfrog's "Destroyed:" ability should have triggered after it was destroyed, but before the "gains 1 aember" sentence. Instead, the opponent forged a key, as noted in the game log:

_**Main phase - Tutenkharnage**
Tutenkharnage plays Saury About That
Tutenkharnage uses Saury About That to destroy Keyfrog and have Testenkharnage gaining 1 Æmber 
Testenkharnage uses Keyfrog to forge a key
Testenkharnage forges the blue key, paying 6 Æmber_

PR #4624, opened 2025-09-10
fix: update saury about that

dmikalova:
Fixes: https://github.com/keyteki/keyteki/issues/3947

### DIFF
--- a/server/game/cards/03-WC/SauryAboutThat.js
+++ b/server/game/cards/03-WC/SauryAboutThat.js
@@ -15,7 +15,7 @@ class SauryAboutThat extends Card {
                 ])
             },
             then: (preThenContext) => ({
-                message: '{3} gains 1 amber',
+                message: '{3} gains 1 amber from {1}',
                 messageArgs: () => [preThenContext.target.controller]
             })
         });


server/game/cards/03-WC/SauryAboutThat.js
```diff
@@ -4,17 +4,20 @@ class SauryAboutThat extends Card {
...
+                gameAction: ability.actions.sequential([
                     ability.actions.destroy(),
                     ability.actions.gainAmber((context) => ({
                         target: context.target ? context.target.controller : []
                     }))
-                ]
-            }
+                ])
+            },
+            then: (preThenContext) => ({
```
strib, line 17: I think you need an `alwaysTriggers: true` here, or else it won't fire if the creature was warded.  (Maybe add a test for that?)

dmikalova: Running the test the ward one does log the aember gain - also I suspect an alwaysTrigger would then log aember gain when there are no creatures in play

> 2025-09-10 21:28:06 player1 plays Saury About That
2025-09-10 21:28:06 player1 uses Saury About That to destroy Lamindra
2025-09-10 21:28:06 player2 uses Lamindra to remove its ward token
2025-09-10 21:28:06 player2 gains 1 amber from Saury About That

Issue #4385, opened 2024-11-16
"Corner the Market" and "The Old Tinker"


UltraUltraUltra:
After reaping with "The Old Tinker", because my opponent played "Corner the Market" in their last turn, I didn't draw a card, but draw part of the "The Old Tinker"'s reap effect is independent of discarding a card.

![image](https://github.com/user-attachments/assets/2d144189-2baa-4936-87e9-10a9a41d8f66)

Issue #3952, opened 2024-06-07
Corner the Market allows you to archive cards discarded from the opponent's hand.


UltimoFox:
I discarded Abyssal Sight from my opponent's hand with Subtle Maul, and I was prompted to choose whether or not to archive it by Corner the Market, which was played against me on the previous turn. This should not happen, as CTM only works on cards discarded from MY OWN hand.

Screenshot: 
![image](https://github.com/keyteki/keyteki/assets/47079472/8cb21312-2c05-4441-8515-42e7b8287c87)

Game Log:
 UltimoFox has connected to the game server
 delphoxician has connected to the game server
delphoxician chooses to randomize the first player
delphoxician won the flip and is first player
UltimoFox is shuffling their deck
UltimoFox draws 6 card  s to their maximum of 6
delphoxician is shuffling their deck
delphoxician draws 6 card  s to their maximum of 6
delphoxician draws 1 card 
Key phase - delphoxician
delphoxician does not forge a key. They have 0 amber. The current cost is 6 Æmber 
House phase - delphoxician
delphoxician chooses unfathomable  as their active house this turn
Main phase - delphoxician
delphoxician plays Fathom Reaver
delphoxician uses Fathom Reaver to make a token creature
Ready phase - delphoxician
delphoxician readies their cards 
Draw phase - delphoxician
End of turn 1
UltimoFox: 0 Æmber  (0 keys) delphoxician: 0 Æmber  (0 keys)
TURN 1 - UltimoFox
Key phase - UltimoFox
UltimoFox does not forge a key. They have 0 amber. The current cost is 6 Æmber 
House phase - UltimoFox
UltimoFox chooses dis  as their active house this turn
Main phase - UltimoFox
UltimoFox plays Yurk
UltimoFox uses Yurk to discard Regrowth
UltimoFox plays Tentacus
Ready phase - UltimoFox
UltimoFox readies their cards 
Draw phase - UltimoFox
UltimoFox draws 3 card  s to their maximum of 5
End of turn 1
UltimoFox: 0 Æmber  (0 keys) delphoxician: 0 Æmber  (0 keys)
TURN 2 - delphoxician
Key phase - delphoxician
delphoxician does not forge a key. They have 0 amber. The current cost is 6 Æmber 
House phase - delphoxician
delphoxician chooses brobnar  as their active house this turn
Main phase - delphoxician
delphoxician plays Ancient Battleground
delphoxician plays Crogg the Clumsy
[...]

PR #4642, opened 2025-10-02
fix: Corner the market interrupting after discard effects

### LINKED ISSUE #3558: Corner the Market + Auto-encoder
Corner the Market now says:
Play: During your opponent’s next turn, they [cannot](https://archonarcana.com/Cannot_vs_Must) play cards, and each time they discard a card from their hand, they may archive that card from their discard pile.

This is no longer a replacement effect, and when using CTM you should also archive the top of your deck if you have auto-encoder out. Currently it does not archive the top card of deck for you.

dmikalova:
- fix: Corner the market interrupting after discard effects - ie the old tinker or auto-encoder
- fix: CTM prompting if you make opponent discard on your turn
- Changed cancel discard to false, added a player check, and updated tests for CTM.

Should fix #3558 and #4385 and #3952

### DIFF
--- a/server/game/cards/06-WoE/CornerTheMarket.js
+++ b/server/game/cards/06-WoE/CornerTheMarket.js
@@ -23,32 +23,20 @@ class CornerTheMarket extends Card {
                             event.location === 'hand' && event.card.controller === context.player
                     },
                     triggeredAbilityType: 'reaction',
-                    gameAction: ability.actions.sequential([
-                        ability.actions.archive((context) => ({
-                            reveal: true,
-                            target: context.event.card,
-                            promptWithHandlerMenu: {
-                                optional: true,
-                                activePromptTitle: 'Archive?',
-                                cards: [context.event.card],
-                                choices: ['Discard'],
-                                handlers: [
-                                    () => {
-                                        context.event.doNotCancelDiscard = true;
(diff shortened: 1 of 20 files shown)


server/game/cards/06-WoE/CornerTheMarket.js
```diff
@@ -42,16 +43,12 @@ class CornerTheMarket extends Card {
...
                             }))
                         })
                     ]),
-                    message: '{0} uses {1} to allow archival of {2} instead of discard',
-                    messageArgs: (context) => [
-                        context.player.opponent,
-                        context.source,
-                        context.event.card
-                    ]
+                    message: '{0} uses {1} to archive {2} from the discard pile',
```
strib: I don't like this change because it says they archive a card even when they choose to discard it. Ideally we'd only print a message when they do choose to archive it.

dmikalova: Hmm good catch, but I'm having trouble implementing this like deusillus. As it is, its before the event so I can't check the card.location. If I move it to a then: then I can't seem to find the card that was discarded. Would the solution be similar to Deusillus - just save the choice to a var?

strib (approved):

PR #4645, opened 2025-10-05
fix: Badger should trigger after playing Badger

### LINKED ISSUE #4644: Badger does not deal damage for itself
When badger is played it should deal 3 damage to an enemy creature, but it does not.

dmikalova:
fix: Badger should trigger after playing Badger - fixes #4644

### DIFF
--- a/server/game/cards/13-CC/Badger.js
+++ b/server/game/cards/13-CC/Badger.js
@@ -11,7 +11,6 @@ class Badger extends Card {
                 when: {
                     onCardPlayed: (event) =>
                         event.player === context.player &&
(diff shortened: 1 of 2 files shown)


server/game/cards/13-CC/Badger.js
```diff
@@ -11,7 +11,7 @@ class Badger extends Card {
                 when: {
                     onCardPlayed: (event) =>
                         event.player === context.player &&
-                        event.card !== context.source &&
+                        // event.card !== context.source &&
```
strib: Can nuke this line.

strib (approved):

PR #4648, opened 2025-10-05
fix: nextRoundEffects wait for opponent's next turn

dmikalova:
- fix: nextRoundEffects wait for opponent's next turn

By default nextRoundEffects would always go off during the next turn, even if it was your turn with cards like Tachyon Manifold or Ancestral Timekeeper. This PR updates the default behavior to wait until the opponent's next turn as found by Copilot.

I added tests for all cards that use nextRoundEffect, and checked that the current tests pass, but I'm not the expert on this codebase so let me know if you think there's a better way to do this.

### DIFF
--- a/server/game/effectengine.js
+++ b/server/game/effectengine.js
@@ -114,12 +114,32 @@ class EffectEngine {
 
         _.each(this.nextRoundEffects, (effect) => {
             if (effect.roundDuration > 1) {
-                effect.nextRound = false;
-                effect.roundDuration -= 1;
-                this.add(effect);
+                // Check if we should wait for opponent turn
+                let shouldActivate = true;
+                if (effect.waitForOpponentTurn && effect.effectController) {
+                    // Only activate if it's now the opponent's turn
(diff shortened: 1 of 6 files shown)


server/game/effectengine.js
```diff
@@ -114,12 +114,32 @@ class EffectEngine {
...
+                    this.add(effect);
+                } else {
+                    // The effect stays in nextRoundEffects for another round
+                }
             }
         });
-        this.nextRoundEffects = [];
+
+        // Only remove effects that were activated
+        this.nextRoundEffects = this.nextRoundEffects.filter((effect) => {
```
strib: This can be simpler, something like:

```
this.nextRoundEffects = this.nextRoundEffects.filter((effect) =>
    effect.roundDuration > 1 &&
    effect.waitForOpponentTurn &&
    effect.effectController &&
    this.game.activePlayer === effect.effectController
```

strib (changes requested):
Looks like the tests got themselves broke.

PR #4655, opened 2025-10-12
fix: Refactor next round system to handle multiple turns from the same player

### LINKED ISSUE #3888: Sensor Cheif Garcia & Extra Turn Card
garcia

Sensor Cheif Garcia: Play/After Fight/After Reap: Keys cost +2A **during your opponent’s next turn.**
Ancestral Timekeeper: (~~~) If you do, take another turn after this one.

In TCO, when a player take another turn, key cost increases if one played sensor chief garcia at previous turn. Because garcia's text said "during your opponent's next turn", I believe it is error.

dmikalova:
nextRoundEffects currently uses a turn counting system where 1 means end of current turn, 2 means end of next turn, and 3 means end of the turn after that, which assumes player, opponent, player turn orders. However with Tachyon Pulse and Ancient Timekeeper that is no longer true.

I have updated the effectengine `onRoundEnded` process to instead use string descriptors that correspond to the game's semantics of turns and removed the roundDuration counters:

```
Expires at the end of the current turn:
- untilEndOfRound
- forRemainderOfTurn

Starts and ends during opponent's next turn:
- duringOpponentNextTurn

I've removed all instaces of 'From now until the end of the opponent's next turn', bc all instances were either really duringOpponentNextTurn or untilNextTurn:
- nextRoundEffect

Expires when the current player's next turn starts:
- untilNextTurn 

Expires at the end of the current player's next turn:
untilEndOfMyNextTurn
```
Fixes #3888

### DIFF
--- a/server/game/effectengine.js
+++ b/server/game/effectengine.js
@@ -108,6 +108,7 @@ class EffectEngine {
             this.unapplyAndRemove(
                 (effect) =>
                     effect.duration === 'untilEndOfMyNextTurnInitiated' &&
+                    // At this point active player has switched, so check for the opposite player
                     effect.effectController !== this.game.activePlayer
             ) || effectsRemoved;
 
@@ -132,7 +133,7 @@ class EffectEngine {
             }
         });
 
-        // Add 'during your opponent's next turn' effects when switching player's
+        // Add 'during your opponent's next turn' effects when switching players
         this.duringOpponentNextTurnEffects = this.duringOpponentNextTurnEffects.filter((effect) => {
             if (
                 effect.duration === 'duringOpponentNextTurn' &&


server/game/effectengine.js
```diff
@@ -103,23 +95,59 @@ class EffectEngine {
...
-        _.each(this.effects, (effect) => {
-            if (effect.roundDuration > 1) {
-                effect.roundDuration -= 1;
+
+        // Remove 'until end of your next turn' effects for the player whose turn is ending
+        effectsRemoved =
+            this.unapplyAndRemove(
+                (effect) =>
+                    effect.duration === 'untilEndOfMyNextTurnInitiated' &&
+                    effect.effectController !== this.game.activePlayer
```
strib, line 112: Is this `!==` because `this.game.activePlayer` has already switched to the opponent by the time `onRoundEnded()` has fired?  If so, please add a comment here so it doesn't confuse someone again like it did me.

server/game/effectengine.js
```diff
@@ -103,23 +95,59 @@ class EffectEngine {
...
+            if (effect.duration === 'untilNextTurn') {
+                effect.duration = 'untilNextTurnInitiated';
             }
         });
 
-        _.each(this.nextRoundEffects, (effect) => {
-            if (effect.roundDuration > 1) {
-                effect.nextRound = false;
-                effect.roundDuration -= 1;
+        // Add 'during your opponent's next turn' effects when switching player's
```
strib: "player's" -> "players"

Issue #3513, opened 2023-10-08
Flea Market with 0 aember in pool


Tutenkharnage:
Flea Market reads:

_**Action:** Look at a random card in your opponent's hand. You may give your opponent 1 aember. If you do, play that card as if it were yours._

TCO won't show a random card on play if you have 0 aember in your pool, but having 0 aember in your pool just prevents you from paying your opponent to play the card yourself; the card clearly allows you to look at a random card in your hand, no matter what.

PR #4661, opened 2025-10-20
WOE: Ekwidon-FleaMarket

DKeatonDC:
Making changes related to issue:
https://github.com/keyteki/keyteki/issues/3513

Changes:

-     modified card Flea Market needed
-     modified test case, checking message to confirm it is sent

Validation done:

-     Test case passed
-     Lint passed


strib:
Thanks for looking at this!  However, the test change here passes without the card change.  `alwaysTriggers` only has an effect in `then` abilities as far as I know. I think maybe #3513 has already been fixed or was never a bug?

DKeatonDC:
@strib -- made these changes as well

PR #4678, opened 2025-11-18
WOE Sactum-Hymn to Duma & Brobnar-Ragnorak Prep

DKeatonDC:
Adds message describing capture.

### DIFF
--- a/test/server/cards/06-WoE/HymnToDuma.spec.js
+++ b/test/server/cards/06-WoE/HymnToDuma.spec.js
@@ -64,7 +64,7 @@ describe('HymnToDuma,', function () {
                     expect(this.flaxia.tokens.amber).toBe(2);
                     expect(this.player2.amber).toBe(1);
                     expect(this).toHaveRecentChatMessage(
-                        'player1 uses Hymn to Duma to capture 2 amber placing it on Flaxia'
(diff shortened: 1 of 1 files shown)


server/game/cards/06-WoE/HymnToDuma.js
```diff
@@ -15,7 +15,9 @@ class HymnToDuma extends Card {
                         cardType: 'creature',
                         controller: 'self',
                         gameAction: ability.actions.capture({ amount: 2 })
-                    }
+                    },
+                    message: '{0} uses {3} to capture 2 amber placing it on {2}',
```
dmikalova: If the opponent doesn't have 2 aember this will be inaccurate

dmikalova (changes requested):

server/game/cards/06-WoE/HymnToDuma.js
```diff
@@ -15,7 +15,12 @@ class HymnToDuma extends Card {
                         cardType: 'creature',
                         controller: 'self',
                         gameAction: ability.actions.capture({ amount: 2 })
-                    }
+                    },
+                    message: '{0} uses {3} to capture {4} amber placing it on {2}',
```
strib: I think it would read better with a `,` after "amber".

strib (approved):

PR #4694, opened 2025-12-11
Prevents ordered discard of a card already discarded

### LINKED ISSUE #4663: Scrap effects stacking issue
I bumped into this issue : 

Image

I had Cryptic Collapse and 2x Hazard Zerp in hand.
I played Cryptic Collapse which activated both Hazard Zerp scrap effects.
Then the scrap effect of one of my Hazard Zerp activated the scrap effect of the second Hazard Zerp.
This resulted in 3 times 3 damage dealt while it should have been 2 times 3 damage.

fionawhim:
Fixes #4663

Handles a bug around Hazard Zerp discarding a card that was already chosen for discard due to Cryptic Collapse, causing a “Scrap:” effect to fire twice for the same card.

Also makes `onOrderedDiscard` store an array of its sequential `onDiscard` events so that Cryptic Collapse can check and see which events actually succeeded, since it’s only those that it should take credit for, rather than all cards that were originally intended to be discarded.

Adds a `DiscardCardsAction.collectDiscardedCards` helper to unravel the possible events that come from discard actions.

Fixes similar discard-counting bugs in:

* Helmsman Spears
* Kaupe (Evil Twin)
* Shopping Spree
* Trade Secrets
* Plummet


test/server/cards/03-WC/HelmsmanSpears.spec.js
```diff
@@ -5,25 +5,64 @@ describe("Helmsman Spears'", function () {
...
             expect(this.player1).toBeAbleToSelect(this.soulkeeper);
             expect(this.player1).toBeAbleToSelect(this.tautauVapors);
+            expect(this.player1).toBeAbleToSelect(this.hazardZerp);
             this.player1.clickCard(this.soulkeeper);
             this.player1.clickPrompt('Done');
             expect(this.soulkeeper.location).toBe('discard');
             expect(this.tautauVapors.location).toBe('hand');
+            expect(this.player1.hand.length).toBe(3);
+        });
+        it('should allow discarding cards with scrapt and drawing of the equal amount when reaping', function () {
```
strib: "scrapt" -> "scrap"

test/server/cards/03-WC/HelmsmanSpears.spec.js
```diff
@@ -5,25 +5,64 @@ describe("Helmsman Spears'", function () {
...
+            // Damage Helmsman Spears (sorry guy, you’re the only creature in play.)
+            this.player1.clickCard(this.helmsmanSpears);
+
+            expect(this.player1).toHavePrompt('Choose a card to play, discard or use');
+            expect(this.soulkeeper.location).toBe('discard');
+            expect(this.tautauVapors.location).toBe('discard');
+            expect(this.hazardZerp.location).toBe('discard');
+
+            // We should only have 2 cards in hand because Helmsman Spears only
+            // discarded 2 cards. Tautau Vapors was discarded by Helmsman
```
strib: Do you mean it was discarded by Hazard Zerp?

test/server/cards/12-PV/CrypticCollapse.spec.js
```diff
@@ -37,5 +40,97 @@ describe('Cryptic Collapse', function () {
...
+
+            // First we lay out the discard order for the Cryptic Collapse
+            expect(this.player1).toHavePrompt('Cryptic Collapse');
+            expect(this.player1).toHavePrompt('Select next card to discard');
+            expect(this.player1).toBeAbleToSelect(this.hazardZerp1);
+            expect(this.player1).toBeAbleToSelect(this.hazardZerp2);
+            this.player1.clickCard(this.hazardZerp1);
+
+            expect(this.player1).toHavePrompt('Cryptic Collapse');
+            expect(this.player1).toHavePrompt('Select next card to discard');
```
strib, line 65: Hmm this doesn't seem right to me.  You should have to completely resolve hazardZerp1's effect before choosing the next card to discard, I think (by the rulebook).  What if the scrap effect caused you to draw a card (Brillix Ponder)?

(I am open to this being a partial measure that gets us closer to the right solution for some common cases. Just want to make sure I understand if this is intentional or not.)

test/server/cards/03-WC/HelmsmanSpears.spec.js
fionawhim: lol fixed

test/server/cards/03-WC/HelmsmanSpears.spec.js
fionawhim: Yes, thank you for the catch.

test/server/cards/12-PV/CrypticCollapse.spec.js
fionawhim: Yep, definitely agree on correctness. This would more or less fall under the 3rd option from [this issue comment](https://github.com/keyteki/keyteki/issues/4663#issuecomment-3632857949), making sure to bear in mind your note about not demanding ordering unless necessary.

The current way multiple-discards-when-scrapping is written is to target all of the discards, then order all of the targets, then enqueue `onCardDiscarded` events for them, with my PR really only achieving the fix of “not double-discarding a card once we get to resolving the `onCardDiscarded`s.”

I’m happy to tackle the more by-the-rules implementation of executing each discard as it’s chosen… might be better in a follow-up PR, since there’s more complexity there and also it would be a UX change to not order everything up front. (And in the meantime the Hazard Zerp issue is fixed.)

(Also, could you clarify for me, if you know: if I discard Brillix Ponder as my _last_ card when doing Cryptic Collapse, does the drawn card get discarded as well? In other words is the “Scrap:” effect still nesting under a “discard your entire hand” when you’re resolving it? Or once you’ve discarded Brillix Ponder would that satisfy “your hand is discarded” and that effect is over before “Scrap:” is resolved?)

strib: Ok thanks, fixing this in a followup makes sense to me.  Thank you for working on this!

> if I discard Brillix Ponder as my last card when doing Cryptic Collapse, does the drawn card get discarded as well?

I don't think we have an official ruling here.  What I've seen in the discord rules discussions (e.g. [here](https://discord.com/channels/563737400275107920/588586239997902858/1411837430222295080)) is that "discard your hand" is usually interpreted as discarding one at a time until you have 0 cards in hand (so the drawn card would get discarded).

fionawhim: Sounds good! I can use that Brillix Ponder behavior as a baseline and try to work this out

Thanks for all of the reviews!

PR #4711, opened 2025-12-29
fix: Nirbor Flamewing not getting onTurnStart reaction when destroyed as a token (#3483)

### LINKED ISSUE #3483: [WOE] - Double Nirbor does not proc at the start of the turn
Hi, 

I use a double nirbor deck and for mysterious reasons, sometimes it does not proc. Sometimes it proc 1 time and sometimes the combo resolves nicely. (it is supposed to proc until you run out of tokens) 

image

Regards - Augustin

fionawhim:
Fixes #3483

This solves issue https://github.com/keyteki/keyteki/issues/3483 for Nirbor Flamewing, where when it hit the discard
as a token creature, its onTurnStart reaction to resurrect itself wouldn’t be
registered.

By swapping the order of `updateEffects` and `updateAbilityEvents`, we
allow for the `CopyCard` effect to be expired first via `updateEffects`, so
that when `updateAbilityEvents` runs it sees the card’s actual set of
reactions, not the stale ones from the copied card.

Adds regression test to Nirbor Flamewing.

Also:
- Fixes typo in CopyCard.js

### DIFF
--- a/server/game/Card.js
+++ b/server/game/Card.js
@@ -785,8 +785,19 @@ class Card extends EffectSource {
         }
 
         if (originalLocation !== targetLocation) {
-            this.updateAbilityEvents(originalLocation, targetLocation);
+            // We want to update effects first, then unregister/re-register
+            // events. This ordering is important for _e.g._ token creatures,
+            // since we need to remove the tokenizing `CopyCard` effect first
+            // (which will happen when updateEffects clears out all lasting
+            // effects), then register any reactions for the card in its new
+            // location.
+            //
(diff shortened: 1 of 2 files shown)


fionawhim:
Converting this to draft because actually I think now there could be problems going in the _other_ direction. If a `CopyCard` effect caused a card to gain reactions, `updateAbilityEvents` would no longer correctly _unregister_ from those events because they wouldn’t come up in `this.getReactions()`.

I think this needs to split `updateAbilityEvents` into “unregister” and “register” halves, where first you unregister, then you update effects, then you register.

PR #4712, opened 2025-12-30
fix: 2 bugs around flipping cards

### LINKED ISSUE #3957: Temporal purge returns aember
When temporal purge is used to flip a token creature, and the card that is flipped is an action, any aember on that card should return to the common supply instead of returning as captured aember because captured aember only returns when a creature is removed from play.

Currently TCO returns the aember.

fionawhim:
Fixes #3957

This fixes 2 bugs with flipping cards:

- Prevents token creatures given Gĕzdrutyŏ the Arcane’s ability from flipping face up, based on [this Sanctumonious Discord discussion](https://discord.com/channels/563737400275107920/588586239997902858/1455295375454896311) that came to the consensus that “flip facedown” meant “physically the card face down” rather than “flip to the other side.”
- Flips cards and returns them to their printed type before discarding them when they’re not creatures. Makes it so that æmber captured on a token creature that’s an action on the other side goes to the common supply when Temporal Purge is played.

(Includes a new test on De-Animator to ensure that its current behavior, which considers the current type of the card — artifact — rather than the printed type — creature — for æmber when it’s destroyed.)

### DIFF
--- a/server/game/Card.js
+++ b/server/game/Card.js
@@ -785,8 +785,19 @@ class Card extends EffectSource {
         }
 
         if (originalLocation !== targetLocation) {
-            this.updateAbilityEvents(originalLocation, targetLocation);
+            // We want to update effects first, then unregister/re-register
+            // events. This ordering is important for _e.g._ token creatures,
+            // since we need to remove the tokenizing `CopyCard` effect first
+            // (which will happen when updateEffects clears out all lasting
+            // effects), then register any reactions for the card in its new
+            // location.
+            //
+            // If the order is reversed, updateAbilityEvents would try to
+            // register any of the _token’s_ reactions for the new location,
+            // even if the tokenness is about to go away.
             this.updateEffects(originalLocation, targetLocation);
+            this.updateAbilityEvents(originalLocation, targetLocation);
+
             this.game.emitEvent(EVENTS.onCardMoved, {
                 card: this,
                 originalLocation: originalLocation,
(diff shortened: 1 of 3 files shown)


server/game/GameActions/FlipAction.js
```diff
@@ -11,6 +16,7 @@ class FlipAction extends CardGameAction {
     canAffect(card, context) {
         return (
             card.location === 'play area' &&
+            (this.direction === 'any' || (this.direction === 'face-down' && !card.isToken())) &&
```
dmikalova: Might it make sense to also add `face-up` since temporal purge references? I don't think there's any situation at the moment where its relevant, but just for completeness

test/server/cards/08-AS/DeAnimator.spec.js
```diff
@@ -145,5 +145,20 @@ describe('De-Animator', function () {
             expect(this.botBookton.location).toBe('discard');
             expect(this.player1).toHavePrompt('Choose a card to play, discard or use');
         });
+
+        it('should send æmber on destroyed mineralized creatures to the common supply', function () {
```
dmikalova: nit: mineralized doesn't exactly mean anything in the game - creatures can have a mineralize counter on them, and if de-animator is out then the creatures with a mineralize counter are artifacts. But if de-animator is not out then they're still creatures.

Might be more accurate to say "destroyed creatures as artifacts" or something similar.

dmikalova (approved):
Looks good, just some small nits

server/game/GameActions/FlipAction.js
fionawhim: Makes sense!

test/server/cards/08-AS/DeAnimator.spec.js
fionawhim: Done.

PR #4714, opened 2025-12-31
WOE Saurian-Senator

DKeatonDC:
Adds a test for Senator.

### DIFF
--- a/server/game/cards/06-WoE/RenewedLife.js
+++ b/server/game/cards/06-WoE/RenewedLife.js
@@ -14,7 +14,9 @@ class RenewedLife extends Card {
             then: {
                 gameAction: ability.actions.makeTokenCreature((context) => ({
                     amount: context.preThenEvent.amount
-                }))
+                })),
+                message:'{0} uses {1} to make {2} token creatures',
+                messageArgs: (context) => [context.player, context.source. context.preThenEvent.amount] 
             }
         });
     }
--- a/server/game/cards/06-WoE/RenewedLife.js
+++ b/server/game/cards/06-WoE/RenewedLife.js
@@ -14,9 +14,7 @@ class RenewedLife extends Card {
             then: {
                 gameAction: ability.actions.makeTokenCreature((context) => ({
                     amount: context.preThenEvent.amount
-                })),
-                message:'{0} uses {1} to make {2} token creatures',
-                messageArgs: (context) => [context.player, context.source. context.preThenEvent.amount] 
+                }))
             }
         });
     }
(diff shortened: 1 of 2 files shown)


server/game/cards/06-WoE/RenewedLife.js
```diff
@@ -14,7 +14,9 @@ class RenewedLife extends Card {
             then: {
                 gameAction: ability.actions.makeTokenCreature((context) => ({
                     amount: context.preThenEvent.amount
-                }))
+                })),
+                message:'{0} uses {1} to make {2} token creatures',
```
fionawhim: Dunno if you use VSCode, but you can install the Prettier extension and enable Format on Save and it should take care of these automatically for you.

dmikalova: Would also say something like '{0} uses {1} to fully heal {2} and make {3} token creatures'

test/server/cards/06-WoE/Senator.spec.js
```diff
@@ -3,79 +3,33 @@ describe('Senator', function () {
...
                 player2: {
                     amber: 4,
-                    inPlay: ['lamindra']
+                    inPlay: ['troll'],
+                    hand: ['poltergeist']
                 }
             });
         });
 
-        describe('after action', function () {
```
dmikalova, line 18: Is there an impetus for removing all these tests rather than just adding on to them?

dmikalova (changes requested):

server/game/cards/06-WoE/RenewedLife.js
DKeatonDC: This is something in-progress (that I still can't figure out) that accidentally got added to my PR.  I'll revert it.

test/server/cards/06-WoE/Senator.spec.js
DKeatonDC: Oh I was being lazy and starting by working from a test for a similar card.  The tachyon manifold test did not seem as vital, but I can work on getting it right it this weekend.

DKeatonDC: @dmikalova -- I added the test back--feel free to take a look.

server/game/cards/06-WoE/RenewedLife.js
DKeatonDC: @fionawhim -- not sure if you saw this, below, but accidentally included some in progress work for another test in my pr.  It's been reverted and made its own pr.

Issue #4708, opened 2025-12-29
Edict of Nerotaurus does not leave play correctly


dmikalova:
After a creature fights or reaps, Edict of Nerotaurus should be applied for the rest of the turn, even if it leaves play - similar to Snag's Mirror. Currently if it leaves play its applied effects go away.

While Edict is not in play, it is still tracking fight/reaps, so when it comes into play its effect immediately applied to prevent reap/fights. It should only start tracking after it comes into play.

I tried implementing this behavior using the current system - I could add conditions to onUseCard to reset the state after Nerotaurus is removed from play, but that doesn't fix both situations. I tried using a reaction similar to Snag's Mirror, however it fails the Umbra dying to Scylla test.

Two other situations - while Blossom Drake is out Edict should not count any reap/fights, and if something were to reap/fight and play an Edict - for example Bot Bookton or Jargogle - then Edict would still be within the after reap/fight window, would see the Bot Bookton reap/Jargogle Fight, and then block the next use.

fionawhim:
How about something like:

1. Change the booleans to `nextCreatureUsedCannotReap` and `nextCreatureUsedCannotFight` just for a bit of clarity.
2. Update the `persistentEffect`s to be `location: 'any'` so that they’ll stick around even if Edict leaves play.
3. Always set the booleans to false in the handlers when cards are used, but only ever set them to `true` if `this.location === 'play area' && !this.isBlank()`.

(I’m assuming here that Blossom Drake is considered the same as the card leaving play… that it wouldn’t keep counting but it would still prevent the next fight/reap from happening.)

One thing I’m wondering about though is whether Edict counts a fight or reap that fizzled due to _e.g._ lethal assault damage. And if it doesn’t count as a fight, would it count as a _use_ and clear the booleans (_i.e._ you could do reap -> fizzled attack -> reap as well as reap -> fizzled attack -> attack)? I think you should be able to check the `cancelled` bit on `event.reapEvent` or `event.fightEvent` in the `onUseCard` handler if you need to be sensitive to this case.

PR #4719, opened 2026-01-02
fix: edict of nerotaurus coming in and out of play

dmikalova:
fix: Rewrote Edict of Nerotaurus to handle coming in and out of play, being blanked, and coming in mid-reap with multiple effects, fixing #4708 
feat: added tests similar tests for Snag's Mirror and Etan's Jar being blanked

Tried updating the original code, which became more and more convoluted with each corner case. Tried using Snag's Mirror reaction method, which set me on the right track, but needed to use Etan's Jar's lastingEffect instead of persistentEffect to handle being blanked.

#4708

### DIFF
--- a/server/game/cards/01-Core/SoulSnatcher.js
+++ b/server/game/cards/01-Core/SoulSnatcher.js
@@ -1,8 +1,7 @@
 const Card = require('../../Card.js');
 
 class SoulSnatcher extends Card {
-    // Each time a creature is destroyed,
-    // its owner gains 1A.
+    // After a creature is destroyed, its owner gains 1A.
     setupCardAbilities(ability) {
         this.reaction({
             when: {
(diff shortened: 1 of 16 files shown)


server/game/cards/06-WoE/EdictOfNerotaurus.js
```diff
@@ -1,59 +1,38 @@
...
+            gameAction: ability.actions.lastingEffect(() => ({
+                until: {
+                    onUseCard: (event) => event.card.type === 'creature',
+                    onTurnEnd: () => true
+                },
                 targetController: 'any',
-                condition: () => this.lastCreatureFought,
                 effect: ability.effects.cardCannot('fight')
-            })
+            }))
```
fionawhim, line 33: Oh nice! This is really elegant, and feels like a perfect translation of the card text.

dmikalova: Thanks! Yeah it really is exactly what I would have hoped for

PR #4720, opened 2026-01-02
fix: Ihaka of the Depths reacts to draws 1 at a time

### LINKED ISSUE #4397: Ihaka of the Depths causes me to draw all cards first, then trigger for however many cards drawn
No game logs available, sorry. 

Deck: https://decksofkeyforge.com/decks/98e7e206-ea1d-4b4f-949e-eb9dc392165f

Each time Ihaka triggers with this deck during my draw cards step, TCO log shows that you draw cards to refill hand first, then trigger Ihaka X times, where X = cards drawn that step. If you have 1 card remaining in deck, you should be able to stack cards on bottom each time to draw them immediately, but instead TCO causes you to shuffle first then place the cards on bottom.

Ruling confirmed by blinkingline here: https://discord.com/channels/669626891447173130/669627064856346624/1314219362885832734

dmikalova:
fix: Ihaka of the Depths reacts to draws 1 at a time - fixes #4397

### DIFF
--- a/server/game/cards/01-Core/SoulSnatcher.js
+++ b/server/game/cards/01-Core/SoulSnatcher.js
@@ -1,8 +1,7 @@
 const Card = require('../../Card.js');
 
 class SoulSnatcher extends Card {
-    // Each time a creature is destroyed,
-    // its owner gains 1A.
+    // After a creature is destroyed, its owner gains 1A.
     setupCardAbilities(ability) {
         this.reaction({
             when: {
--- a/server/game/cards/01-Core/Tolas.js
+++ b/server/game/cards/01-Core/Tolas.js
@@ -2,18 +2,17 @@ const Card = require('../../Card.js');
 
 class Tolas extends Card {
     // Elusive. (The first time this creature is attacked each turn, no damage is dealt.)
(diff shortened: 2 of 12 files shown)


server/game/cards/07-GR/IhakaOfTheDepths.js
```diff
@@ -4,13 +4,14 @@ class IhakaOfTheDepths extends Card {
...
+        });
+
         this.reaction({
             when: {
                 onCardPlaced: (event, context) =>
-                    event.drawn &&
-                    context.source.controller === event.card.controller &&
-                    context.source.controller === context.game.activePlayer &&
-                    context.source.controller.discard.length > 0
+                    event.drawn && context.source.controller === event.card.controller
```
strib: I don't follow why you removed the two conditions here.  Don't we still need to restrict it to the controller's turn, and also only react when there's a card in the discard pile?  (I get that you can try to use Ihaka on an empty discard pile and it will just fizzle, but is there a reason to force a click in that case?)

dmikalova: I'm not sure either, agreed, and reverted.

Issue #3959, opened 2024-08-27
Add "USERNAME mulliganed" to the text log for log parsing tools


dimfish:
I gather from the discord stats thread that there is a game-level db. I'm interested in parsing the game's text log to analyze house choice, aember over turns, stuff like that. It would be nice if the mulligan got recorded for this tracking.

(I'm catsmdogs on TCO and discord)

PR #4721, opened 2026-01-02
feat: add mulligans to logs

dmikalova:
feat: add mulligans to logs for #3959

### DIFF
--- a/server/game/cards/01-Core/SoulSnatcher.js
+++ b/server/game/cards/01-Core/SoulSnatcher.js
@@ -1,8 +1,7 @@
 const Card = require('../../Card.js');
 
 class SoulSnatcher extends Card {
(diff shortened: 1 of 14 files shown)


server/game/gamesteps/setup/mulliganprompt.js
```diff
@@ -43,6 +45,22 @@ class MulliganPrompt extends AllPlayerPrompt {
 
         return false;
     }
+
+    onCompleted() {
```
strib, line 56: This is definitely a good improvement.  But do you think it would be possible to print the first player mulligan/keep decision as soon as it happens?  That would give us a better way to simulate IRL rules.

dmikalova: Didn't even think of that - implemented!

PR #4722, opened 2026-01-02
fix: Tolas and Neffru for owner

### LINKED ISSUE #3949: Tolas gives aember to wrong player when creature with treachery is destroyed
First off, my apologies for not saving the log for the game where this happened.

My opponent had played an Infiltrator and it was in my battleline. I played Tolas and then used Three Fates to kill the Infiltrator. When the on destroyed trigger is resolved, the Infiltrator has been moved to my opponent's discard so `context.event.card.controller.opponent` gives the aember to me instead of my opponent.

This probably also happens with other effects that change who controls a creature.

dmikalova:
fix: Tolas and Neffru for owner - fixes #3949

- Updated card comments for soul snatcher, tolas, and neffru to the new template
- Once a creature is destroyed and goes to discard, the controller reverts to the owner. This meant that Tolas was giving incorrent aember for creatures that changed control. Neffru also had this issue, but was implemented incorrectly to consider controller so both errors negated eachother.
- Updated message text for tolas and neffru to specify which creature destruction gained the aember

### DIFF
--- a/server/game/cards/04-MM/OptioGorkus.js
+++ b/server/game/cards/04-MM/OptioGorkus.js
@@ -2,7 +2,7 @@ const Card = require('../../Card.js');
 
 class OptioGorkus extends Card {
     // Elusive.
-    // Each of Optio Gorkuss neighbors gains, Destroyed: Move each A on this creature to a neighboring Optio Gorkus.
+    // Each of Optio Gorkus's neighbors gains, "Destroyed: Move each A on this creature to a neighboring Optio Gorkus."
     setupCardAbilities(ability) {
         this.persistentEffect({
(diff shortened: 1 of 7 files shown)


test/server/cards/02-AoA/Neffru.spec.js
```diff
@@ -54,15 +56,20 @@ describe('Neffru', function () {
         });
 
         it('should cause the owner of a destroyed creature to gain an amber when it attacks and kills it', function () {
```
strib: I think this description is wrong now, since "it attacks" refers to Neffru attacking I think?  And/or the test is testing something different now.  Maybe better just to add a new test instead?

PR #4747, opened 2026-01-17
fix: gigantic creatures missing keywords

### LINKED ISSUE #4406: M+: Sirs Colossus does not correctly show to be a taunt creature
I have a deck with two copies of Sirs Colossus. When there is one copy on board, it does not show to be a taunt creature. When both copies are on board, one shows as taunt and one does not. 

If I play a Securi-Droid on Sirs Colossus, it moves ahead of other creatures to indicate it has taunt. 

The deck in question: https://www.keyforgegame.com/deck-details/2f8100ff-e2e9-4656-9bfb-7412cc8d4722

I do not know if it is a function of there being two copies of it in my deck.

dmikalova:
fix: Add keywords to both gigantic halves when in play - eg Sirs Colossus should have taunt if the second half is played.

Added a related test for Deusillus to make sure traits are also being applied correctly.

Fixes #4406


test/server/cards/10-MoMu/SirsColossus.spec.js
```diff
@@ -80,5 +83,51 @@ describe('Sirs Colossus', function () {
...
+
+            this.player1.play(this.theSpiritSWay);
+            expect(this.sirsColossusTop.getTraits()).toEqual([]);
+            expect(this.sirsColossusTop.hasKeyword('taunt')).toBe(false);
+            expect(this.sirsColossusBottom.getTraits()).toEqual(['knight', 'spirit']);
+            expect(this.sirsColossusBottom.hasKeyword('taunt')).toBe(true);
+            expect(this.player1).isReadyToTakeAction();
+        });
+
+        it('should share keywords and traits when top is played', function () {
```
strib: Typo here: "top" -> "bottom"

strib (approved):

PR #4748, opened 2026-01-18
fix: several issues with gigantic creatures

### LINKED ISSUE #4423: Plan 10 and Gigantic
When I play Plan 10 with a Gigantic on board, Plan 10 count the Gigantic as one card so at the end of either play turn, if you chose that gigantic, TCO currently will return both cards back to your hand. Which is wrong I believe, as Gigantic counts as two cards when out of play and Plan 10 specificly says it will return just one card at the end of the turn.

### LINKED ISSUE #4426: Elite Disruptzord doesn't block gigantic top half
When my opponent has an Elite Disruptzord out I can still play the top half of a gigantic such as Wretched Anathema which is 10 power. Disruptzord should block either half of the gigantic from being played because it is played as one card.

### LINKED ISSUE #4440: Shell of a Ghost and Gigantic Creatures
During one of my games; I played Shell of a Ghost.

> Play: Destroy each creature that is not on a [flank](https://archonarcana.com/Flank). Put a creature from any discard pile into play under your [control](https://archonarcana.com/Control). Gain 2 chains.

I went ahead and picked out a gigantic creature (which both halves were in the discard) and it somehow was able to put the gigantic into play as if it was a regular creature card.

dmikalova:
This PR fixes several situations for when and how TCO handles both halves of Gigantic creatures. This stems from the fact that the top half is a creature with the bonus icons, but does not have the power, armor, traits, keywords, or abilities that the bottom half has. If a player plays a gigantic creature from hand by selecting the top half, then there were situations where it would not properly consider the other half of the gigantic creature.

- fix: Exhume - updated to new card text, added tests to allow selecting gigantic half and then fizzling
- fix: Elite Disruptzord/De-Doss - now blocks the top half from being played by using the combined gigantic creature's power. fixes #4426 
- fix: Overlord Greking should put any creatures killed by splash damage into play. However, each instance of this ability only has an allowance to play 1 card, so it will always fizzle for gigantics after attempting to play them. Similarly, Francus/Hapsis fighting a neighbor and killing shadow self should proc Francus, which the fix for this also fixes and the test cases updated to the correct ruling.
- fix: Plan 10 was returning both halves of a gigantic, and should only return one at a time. fixes #4423 
- fix: Shell of a Ghost was playing both halves of a gigantic, when it should be selectable but if selected it will fizzle due to only having 1 allowance. fixes #4440 
- fix: Brutal Consequences should only purge 1 half of a gigantic creature since it has already moved to the discard and been separated.
- fix: Francus and Hapsis tests were wrong. Added several other related test cases.


server/game/GameActions/PurgeAction.js
```diff
@@ -21,6 +21,58 @@ class PurgeAction extends CardGameAction {
         return this.targetType.includes(target.type);
     }
 
+    preEventHandler(context) {
```
dmikalova, line 24: This seems reasonable to me but maybe can be simplified?

server/game/GameActions/PutIntoPlayAction.js
```diff
@@ -33,9 +33,25 @@ class PutIntoPlayAction extends CardGameAction {
         return true;
     }
 
+    // Check if both halves of a gigantic creature can be put into play
```
dmikalova: tbh I don't know why these conditions work, but tests fail if any piece is removed.

server/game/GameActions/PutIntoPlayAction.js
```diff
@@ -205,22 +221,35 @@ class PutIntoPlayAction extends CardGameAction {
...
                     if (!part && card.parent) {
-                        // parts are placed togehter under another card and can be put into play together
                         part = card.parent.childCards.find((part) => card.compositeId === part.id);
                     }
 
-                    if (part) {
-                        card.controller.removeCardFromPile(part);
-                        card.composedPart = part;
+                    // Play from discard or other pile
+                    if (!part && !this.myControl && !this.beingPlayed) {
```
dmikalova: Again not sure why these specific conditions work

server/game/gamesteps/handlermenuprompt.js
```diff
@@ -112,6 +112,27 @@ class HandlerMenuPrompt extends UiPrompt {
         return { menuTitle: this.properties.waitingPromptTitle || 'Waiting for opponent' };
     }
 
+    onCardClicked(player, card) {
```
dmikalova, line 115: This handles Brutal Consequences where a gigantic is destroyed, goes to discard, and then the player can choose to purge one of the halves.

server/game/player.js
```diff
@@ -582,16 +582,13 @@ class Player extends GameObject {
             cloneOverride: origCard,
             from: location,
             to: targetLocation,
-            drawn: options.drawn
+            drawn: options.drawn,
```
dmikalova, line 599: This label is used by event handlers - eg purge - to be able to identify both halves of the gigantic after it leaves play

test/server/cards/01-Core/Francus.spec.js
```diff
@@ -45,12 +45,13 @@ describe('Francus', function () {
             this.player1.endTurn();
         });
 
-        it('should not capture 1 amber when destroyed creature was not involved in the fight', function () {
+        it("should capture 1 amber when fighting shadow self's neighbor", function () {
```
dmikalova, line 48: My understanding of the rules is that this test was wrong - Francus fight a neighbor of Shadow Self and killing Shadow Self should count Shadow Self as dying from fighting Francus.

test/server/cards/01-Core/ShadowSelf.spec.js
```diff
@@ -160,7 +160,7 @@ describe('Shadow Self', function () {
             this.player2.clickPrompt('logos');
             this.player2.fightWith(this.hapsis, this.badPenny);
             expect(this.hapsis.tokens.damage).toBe(1);
-            expect(this.hapsis.warded).toBe(false);
+            expect(this.hapsis.warded).toBe(true);
```
dmikalova, line 163: Same as Francus - this test was wrong

server/game/cards/02-AoA/Exhume.js
```diff
@@ -1,7 +1,7 @@
 const Card = require('../../Card.js');
 
 class Exhume extends Card {
-    // Play: Choose a creature in your discard pile. You may play that creature as if it belonged to the active house and was in your hand.
+    // Play: You may play a creature from your discard pile.
```
dmikalova, line 4: Exhume's wording changed for VM23 and its relevant to gigantic creatures

server/game/Card.js
```diff
@@ -958,8 +969,22 @@ class Card extends EffectSource {
         return this.getPower();
     }
 
-    getPower(printed = false) {
-        const printedPower = this.getBottomCard().printedPower;
+    getPower({ printed = false, restriction = false } = {}) {
```
dmikalova: Changed this to a destructured {} format so there's less random true/false all over the place - not sure if this is a good idea to keep doing going forward, or if its out of place.

server/game/GameActions/PutIntoPlayAction.js
dmikalova: Overlord Opus says:

> Why beingPlayed and myControl work (accidentally)
Overlord Greking: Uses myControl: true → triggers the condition
Normal play from hand: Uses both beingPlayed: true AND myControl: true, but location === 'hand' so it passes
Saurian Egg: Uses bare putIntoPlay() with neither flag → doesn't trigger the condition
Exhume: Uses playCard() which goes through PlayCreatureAction → sets beingPlayed: true and myControl: true → triggers the condition

Not sure if we want to leave as is, or put more effort into having the engine handle this in a less coincidental way.

server/game/Card.js
```diff
@@ -972,8 +972,22 @@ class Card extends EffectSource {
...
+        let printedPower = this.getBottomCard().printedPower;
+
+        // Return power needed for play restrictions
+        if (restriction) {
```
strib: Does it ever make sense to call with `printed == false` and `restriction == true`?  If so, then this is probably bugged in that case since it's not checking for effects.  If not, then maybe check that condition explicitly and throw an error?

server/game/Card.js
strib: I'm good with that.  Except the name `restriction` is pretty ambiguous.  Would `powerBeforePlay` make sense?

But depending on the answer to the below question, maybe it just makes sense to have it as a separate function, if it never needs to drop down into the effect checks.

server/game/GameActions/PurgeAction.js
```diff
@@ -21,6 +21,58 @@ class PurgeAction extends CardGameAction {
...
+            this.setTarget([context.event.card, context.event.giganticOtherHalf]);
+            return;
+        }
+
+        // Handle gigantic creatures that have been separated in discard
+        // If a single gigantic card is targeted but is now in discard without its composedPart,
+        // and the other half is also in discard, prompt for which half to purge.
+        // This handles the case where a card effect purges a creature that was destroyed
+        // and separated into two cards in the discard pile.
+        if (this.target.length === 1) {
```
strib, line 42: Would this allow the player to choose the bottom half of a gigantic that's in the discard when Hamstrung discards the top half?  Seems like it needs to distinguish the case where a card was targeted when originally in play.

server/game/GameActions/PutIntoPlayAction.js
strib: Yeah this seems pretty sketchy.  I think I'd rather just have an `allowGigantic` or `canPlayMultipleParts` or `numPlayAllowances` boolean on the action that gets plumbed through, at least then we can reason about it.

server/game/gamesteps/handlermenuprompt.js
strib: I don't follow these changes.  Originally,  the inherited `onCardClicked` just returns `false` and does nothing else.  Why are we now calling `cardHandler` when its in `properties.cards`?  (As far as I can tell, the return value is never actually used for anything.)

strib (commented):
This one is a bit tough to understand.  Left some suggestions and questions.

server/game/GameActions/PurgeAction.js
dmikalova: This does allow that that - working on a fix.

server/game/Card.js
dmikalova: I've changed this to a separate `powerForPlayRestriction` that gigantics implement - I think it came out a lot cleaner and resolves this.

server/game/gamesteps/handlermenuprompt.js
dmikalova: I was able to find a simpler logic that passes tests. This is needed bc Brutal Consequences destroys the gigantic, and then the gigantic separates in the discard, and without this Brutal Consequences will then purge both halves bc it is targeting the composed card, rather than spending its 1 purge allowance on the half a player chooses.

server/game/GameActions/PurgeAction.js
dmikalova: Fixed

server/game/Card.js
dmikalova: Fixed

strib (approved):

PR #4749, opened 2026-01-18
Update docs for AGENTS.md

dmikalova:
Started updating AGENTS.md and that lead me to go through and update all of the docs to be both more human readable and agent readable.


.prettierrc.js
```diff
@@ -3,5 +3,13 @@ module.exports = {
     printWidth: 100,
     singleQuote: true,
     jsxSingleQuote: true,
-    trailingComma: 'none'
+    trailingComma: 'none',
+    overrides: [
+        {
+            files: '*.md',
+            options: {
+                tabWidth: 2
```
dmikalova: This makes it so markdown lists (`- text`) don't use 4 space indentation (`-   text`) with prettier

AGENTS.md
```diff
@@ -0,0 +1,144 @@
...
+
+- Use `grep_search` to find cards with similar abilities (e.g., search "steal", "dealDamage", "destroyed")
+- Check the same set directory for patterns used in that expansion
+- Look at cards from Core set (`CotA/`) for foundational patterns
+- Search by ability type: `this.play(`, `this.reap(`, `this.persistentEffect(`
+
+### Testing Cards
+
+- Tests should be added to a new `spec.js` file for the card under `test/server/cards/<Set>/<CardName>.spec.js` that corresponds to the card being tested.
+  - Example: [BadOmen.spec.js](test/server/cards/PV/BadOmen.spec.js).
```
strib: Should be `12-PV` still, right?  I know you're thinking of changing it, but that hasn't landed yet.

AGENTS.md
```diff
@@ -0,0 +1,144 @@
...
+
+Ask the user before proceeding when:
+
+- Card text is ambiguous or has multiple valid interpretations
+- The desired behavior conflicts with how similar cards work
+- A fix could be in the card implementation OR the game engine
+- Multiple test failures suggest a broader architectural issue
+- The card references mechanics or keywords not yet implemented
+
+## Commands you can use
```
strib, line 47: We can instruct the agent on how to start the webserver, what port it is running on, etc. Some will be able to use browser use to actually test front-end work, give you screenshots, etc.  But we can add that in a future update if you prefer.

AGENTS.md
```diff
@@ -0,0 +1,144 @@
...
+## Code Style Conventions
+
+- Card classes extend `Card` and implement `setupCardAbilities(ability)`
+- Card ID is set as static property: `CardName.id = 'card-id';`
+- Export with `module.exports = CardName;`
+- Use arrow functions for conditions and callbacks
+- Comment the card text above `setupCardAbilities`
+- Always end files with a newline
+
+## Card Implementation Guide
```
strib, line 80: All this documentation is great but I'm a bit nervous about it getting out of date if an agent makes core changes to mechanics.  Maybe add a section about updating the docs if anything relevant is changed outside of `cards/`?

strib (approved):
This looks great to me overall, just added some nits.  I think it would also be good to get @cryogen to stamp this one.

AGENTS.md
dmikalova: Added a section for that - also pulled in some changes to run nodemon since I found that very useful for quickly reloading UI changes.

strib (approved):

PR #4754, opened 2026-01-21
fix: allow ordering multiple after play abilities

### LINKED ISSUE #4753: Teliga Urchin with no aember should be orderable
Teliga Urchin with no aember should be orderable, but since there is no aember it doesn't meet requirements and the urchin ability gets discarded.

dmikalova:
- fix: ordering of multiple after play abilities and the relevant tests - fixes #4753


test/server/cards/03-WC/FranesBlaster.spec.js
```diff
@@ -213,6 +213,7 @@ describe('Frane’s Blaster', function () {
             this.player1.reap(this.firstOfficerFrane1);
             this.player1.clickCard(this.firstOfficerFrane1);
 
+            this.player1.clickPrompt(this.franeSBlaster.name);
```
strib: Do you understand why there needs to be a click here?  It already clicked Frane itself above to choose to do Frane first.  Also this is not a play or a destroyed ability, so this PR shouldn't change it.  Same with Wall's below.

strib (commented):
Very nice! Just one comment I'm not sure about.

test/server/cards/03-WC/FranesBlaster.spec.js
dmikalova: I've updated the tests to reflect the flow more. The removal of the `this.meetsRequirements(context, []) === 'condition' &&` is what causes the flow to change - however this condition doesn't seem necessary since it doesn't make any changes other than asking to order frane's blaster - which is what should happen.

You'll notice the tests for Walls' are a bit different than Frane bc frane has something to capture, whereas walls does not have something to pull from discard. The tests prompt to order and work properly, however while walking through the UI it asks you to click Walls for both its ability and its blaster ability, and then picks blaster first and then resolves Walls' ability. This is the same behaviour as before, so I made a [separate ticket](https://github.com/keyteki/keyteki/issues/4767) to address it.

Lemme know if this is satisfactory, or if you want me to add the meetsRequirements back in.

strib (approved):

test/server/cards/03-WC/FranesBlaster.spec.js
strib: Thanks!  Much easier to follow this way.

PR #4762, opened 2026-01-23
fix: flipping cards without a token creature defined

### LINKED ISSUE #4750: Magda the rat causes a server error if flipped
Image

- Play Magda to steal and establish a lasting effect for when it leaves play
- Use creed of nurture to discard gezdrutyo the arcane and give Magda Gezdrutyo's action to flip
- Use Magda's Action to flip it over
- TCO now reports a server error
- The rules do not consider flipped cards to have entered or left play - instead Magda's ability should resolve once the game object (the flipped card) leaves play - [source](https://discord.com/channels/563737400275107920/588586239997902858/1462502762540568657)
- Check similar cards like turnkey

dmikalova:
- fix: flipping cards without a token creature defined - fixes #4750
- fix: several typos for "occurred" - noticed this from the Magda the Rat error.

If a creature with an on leaves play effect like magda or turnkey is in play, that effect should resolve not when flipped, but when the card itself leaves play.

If gezdrutyo or mimic gel gezdrutyo flip, they should flip into their owner's token creature reference card, and if none then go to discard, but are no longer considered a creature so wards don't protect them and amber is return to the common supply.

Similarly if Keyfrog flips and the owner does not have a token creature reference card, its destroyed effect should not trigger as it moves to discard.


test/server/cards/06-WoE/GezdrutyoTheArcane.spec.js
```diff
@@ -98,4 +98,63 @@ describe('Gĕzdrutyŏ the Arcane', function () {
...
+                    inPlay: ['gĕzdrutyŏ-the-arcane']
+                },
+                player2: {
+                    inPlay: ['iron-heidy'],
+                    hand: ['requisition-writ']
+                }
+            });
+        });
+
+        it("should flip into owner's token creature when opponent does not have token creature", function () {
```
strib, line 147: This one seems a bit wild to me.  Is this one official?

dmikalova: Yeah, as far as we can interpret the FAQ. The conclusion we came to in [Sanctumonious](https://discord.com/channels/563737400275107920/588586239997902858/1464365668530130964) is that "if you flip a card facedown, it becomes the little example token creature that is printed there. If there isn't one there then it leaves play to the discard and discarding all tokens/amber".
image

strib: Insanity.

PR #4764, opened 2026-01-24
fix: tryska's neighbors enters should enter play ready

dmikalova:
- fix: tryska's neighbors enters should enter play ready


server/game/Effects/Values/ConditionValue.js
```diff
@@ -10,14 +10,14 @@ class ConditionValue extends EffectValue {
         return this;
     }
 
-    checkCondition(abilityContext, effectContext) {
+    checkCondition(target, abilityContext, effectContext) {
```
dmikalova, line 13: Tryska's neighbors were not entering play ready from token creatures or gebuk because the check condition was checking Gebuk or the action that made the token creature, instead of the target of Gebuk or action being put into play.

server/game/cards/05-DT/WasteNot.js
```diff
@@ -7,16 +7,19 @@ class WasteNot extends Card {
...
-                    ability.actions.draw((context) => ({
-                        target: context.player,
-                        amount: context.target ? Math.ceil(context.target.power * 0.5) : 0
-                    }))
-                ]
+                gameAction: ability.actions.destroy()
             },
-            effect: 'destroy {0} and draw {1} cards',
-            effectArgs: (context) => (context.target ? Math.ceil(context.target.power * 0.5) : 0)
+            effect: 'destroy {0}',
```
dmikalova, line 12: Separate the destroy from the draw with a then so destroy abilities can interrupt.

strib (approved):

PR #4770, opened 2026-01-24
feat: Add damage() getter

dmikalova:
feat: Add damage() getter - instead of doing `expect(this.mother.tokens.damage).toBe(undefined);` you can now do `expect(this.mother.damage).toBe(0);`


strib (commented):
Why is this just for tests?  Any reason not to update all the `tokens.damage` uses under `server/game/cards`?

strib (approved):

PR #4771, opened 2026-01-24
feat: Add powerCounters() getter

dmikalova:
feat: Add powerCounters() getter - can now do `expect(this.mugwump.powerCounters).toBe(0);` instead of `expect(this.mugwump.tokens.power).toBeUndefined();`


strib (commented):
Same question here, why not update the few cards that user `tokens.power`?

PR #4784, opened 2026-01-26
fix: Lord Golgotha and Smite

### LINKED ISSUE #1946: Lord Golgotha and Smite seem to not work correctly together (maybe related to taunt)
See turn 2 in the attached log

My opponent had a board of, from left to right: daughter, prescriptive grammarbot, daemo-bot and infomorph. All undamaged

I played a creature with a damage pip to deal one damage to grammarbot, then I played lord golgotha followed by Smite and chose to attack the infomorph on their far right.

What should happen is that the before fight ability of golgotha should kill the daemo-bot, the infomorph should die in the fight and the the 2 damage from smite should kill the grammarbot

But what happened was that only the infomorph and daemo-bot took damage and died. No smite damage was dealt to the grammarbot.
[smite golgotha.txt](https://github.com/keyteki/keyteki/files/5164946/smite.golgotha.txt)

dmikalova:
fix: Lord Golgotha and Smite - fixes #1946 

Tried to simplify smite, but it was surprisingly difficult and a seemingly unique card. Due to the ready and fight part, it doesn't seem like the fight target is available in the preThenContext.

### DIFF
--- a/server/game/cards/02-AoA/Smite.js
+++ b/server/game/cards/02-AoA/Smite.js
@@ -1,77 +1,45 @@
 const Card = require('../../Card.js');
-const { EVENTS } = require('../../Events/types.js');
-const FightGameAction = require('../../GameActions/FightGameAction');
-const AbilityResolver = require('../../gamesteps/abilityresolver');
-const SimpleStep = require('../../gamesteps/simplestep');
-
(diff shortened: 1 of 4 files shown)


strib (approved):
I like this cleanup but am scared of it.  But let's try!

PR #4790, opened 2026-01-26
feat: switch to vitest

dmikalova:
feat: switch to vitest, saving about 40% time on running the full test suite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
     "scripts": {
         "mongod": "mongod --dbpath=/data --port 27017\n",
         "start": "node .",
-        "dev:db": "docker-compose up -d redis postgres",
         "dev:lobby": "NODE_APP_INSTANCE=node nodemon --inspect=0.0.0.0:9229 .",
         "dev:gamenode": "NODE_APP_INSTANCE=node nodemon --inspect=0.0.0.0:9229 server/gamenode",
         "fetchdata": "node server/scripts/fetchdata",


test/helpers/vitest-reporter.js
```diff
@@ -0,0 +1,190 @@
+/**
```
dmikalova, line 1: All of the default reporters output a line for each file, except the dot reporter which has known issues with the amount of files and tests we're running. This instead just tallies the amounts as it goes and looks like this:

```
 npm run test

> keyteki@2.0.0 test
> vitest run


 RUN  v4.0.18 /Users/david.mikalova/Code/github.com/dmikalova/keyteki


 Test Files  80 passed (80/2313 3%)
      Tests  11702 passed (11702)
   Start at  15:38:04
   Duration  5.31s
```

cryogen:
> > Please try to keep changes focused and don't make unrelated changes. Can you remove the non vitest stuff from this PR?
> 
> I'm trying to, but its also hard to know exactly which changes you're referring to 😅 Here's an explanation of each change:
> 
>     * There were a couple of files that got hit by prettier while iterating, thought I removed these but prettier auto fmts on commit. Should be removed now
> 
>     * Some setup with docker, I was using docker to make sure everything was building with vitest but I can see how that's different
> 
>     * test/helpers/vitest-reporter.js - This isn't strictly needed in this PR, but it makes the test output a lot better to parse with the amount of tests we have
> 
>     * There are latent bugs in the tests that switching to vitest's greater strictness exposed:
>       
>       * test/helpers/playerinteractionwrapper.js - tests that were setting to a deck were doing it wrong before
>       * test/server/cards/02-AoA/ScowlyCaper.spec.js - This was an error that's now exposed
>       * vitest uses it.skip instead of xit:
>         
>         * test/server/cards/02-AoA/YzphyzKnowdrone.spec.js
>         * test/server/cards/06-WoE/Bryozoarch.spec.js
>         * test/server/gamesteps/selectcardprompt.spec.js
>         * test/server/tide.spec.js
>       * test/server/cards/03-WC/Etaromme.spec.js - this test was not set up correctly and now fails
>       * test/server/cards/07-GR/PsychicHaze.spec.js - this test was not set up correctly and now fails
>       * test/.eslintrc.js - updated for vitest. Specifically the no-restricted-syntax rule is to recreate the jasmine no-focused-tests rule, since the version of eslint and vitest that natively supports doesn't support the version of eslint being used
>       * test/tsconfig.json - Adds vitest types to the test folder
>       * .eslintrc.js - updated for vitest
>       * test/helpers/integrationhelper.js - updates all the custom matchers to vitest style
>       * I think the rest of the cards are are self-evident updating to vitest
> 
> 
> Lemme know if there's anything more I can break down.

You got the ones I was bothered, about, thanks!

PR #4802, opened 2026-01-27
test: Watch Your Step with empty opponent's deck

dmikalova:
test: Watch Your Step with empty opponent's deck - related to #3421


test/server/cards/06-WoE/WatchYourStep.spec.js
```diff
@@ -44,6 +44,24 @@ describe('Watch Your Step', function () {
             expect(this.player1.player.creaturesInPlay[0].exhausted).toBe(false);
             expect(this.player1.player.creaturesInPlay[1].exhausted).toBe(false);
         });
+
+        it("should not make tokens if the opponent does follow instructions and opponent's deck is empty", function () {
```
strib: Is this supposed to say it _should_ make tokens if the opponent deck is empty (which is irrelevant to the ability, I think).

dmikalova: You're right - not should have been in "does not follow"

strib (approved):

PR #4818, opened 2026-01-29
fix: Game server on node 22

dmikalova:
fix: Game server on node 22

Did what I needed to do to get the game lobby and node building and running. Ran through a full game with no issues.

Vite complained about the locales in public so I moved them to client:
```
Assets in public directory cannot be imported from JavaScript.
If you intend to import that asset, put the file in the src directory, and use /src/locales/es.json instead of /public/locales/es.json.
If you intend to use the URL of that asset, use /locales/es.json?url.
```


client/locales
dmikalova, line 1: This seemed to be a random exact copy of the zhannt.json locale??

PR #4819, opened 2026-01-29
feat: 25% faster tests

dmikalova:
feat: 25% faster tests by profiling - from ~23s to ~16s

The changes break down to either short-circuiting several expensive recursive checks, and disabling an expensive check serialization option when the full suite running locally. Each cuts about 3 seconds.


server/game/cards/index.js
```diff
@@ -1,44 +1,106 @@
...
 const fs = require('fs');
 const path = require('path');
 
-function getDirectories(srcpath) {
-    let fullPath = path.join(__dirname, srcpath);
-    return fs.readdirSync(fullPath).filter(function (file) {
-        return fs.statSync(path.join(fullPath, file)).isDirectory();
-    });
-}
+// Map cardId -> filePath (built by scanning files without requiring them)
```
dmikalova, line 4: The changes in this file lazy load cards for tests

server/game/Effects/Effect.js
```diff
@@ -91,10 +92,27 @@ class Effect {
             return true;
         }
 
-        let effectOnSource = this.source.persistentEffects.some(
-            (effect) => effect.ref && effect.ref.includes(this)
-        );
-        return (this.location === 'any' || !this.source.facedown) && effectOnSource;
+        // Check if source is blanked and this is a printed ability (not a keyword effect)
```
dmikalova, line 95: The changes in this file returns early where possible before doing a full array traversal when needed

server/game/Card.js
```diff
@@ -932,7 +940,39 @@ class Card extends EffectSource {
...
+                let copyEffect = this.mostRecentEffect('copyCard');
+                let baseKeywords = copyEffect
+                    ? copyEffect.printedKeywords
+                    : this.getBottomCard().printedKeywords;
+                return baseKeywords[keyword] || 0;
+            }
+            return 0;
+        }
+
+        // Single pass over effects for both removeKeyword and addKeyword
```
dmikalova, line 954: Minimizing recursive check

strib (commented):
I think I'd like @cryogen to weigh in on this one, I'm scared of how integral these changes seem to the engine.

server/game/Card.js
```diff
@@ -915,15 +919,15 @@ class Card extends EffectSource {
     }
 
     isBlank() {
-        return this.anyEffect('blank');
+        return this.effects.some((effect) => effect.type === 'blank');
```
cryogen, line 922: why change these?

server/game/cards/index.js
cryogen: This doesnt just lazy load for tests, it does it on the live server. This is an over optimisation that isn't needed and could be harmful

server/game/effectengine.js
```diff
@@ -83,7 +83,8 @@ class EffectEngine {
         );
         if (loops === 10) {
             throw new Error('EffectEngine.checkEffects looped 10 times');
-        } else {
+        } else if (stateChanged || this.newEffect) {
```
cryogen, line 86: This makes me nervous

cryogen (changes requested):

PR #4854, opened 2026-02-04
fix: Aemberling choice

### LINKED ISSUE #3502: Forging blindly with aemberling
When there is the prompt to know if we want to forge with an aemberling, we dont know witch aemberling the prompt is talking about. we might not want to destroy some token that have upgrade for example

dmikalova:
fix: Aemberling choice - fixes #3502


server/game/player.js
```diff
@@ -738,19 +738,70 @@ class Player extends GameObject {
     }
 
     getAmberSources() {
```
dmikalova, line 740: These changes consolidate the 3 amber source workflows into one, allowing:

- the order of spend to be chosen by player. This has no real effect on the game, but helps convenience their decisions
- Cards that can be spent, like Amberling, can be selected, rather than prompting through one at a time, left to right.
- Amber is not spent until all decisions are made and key is about to be forged - eg if you have Amberling and callipygian ideal, the Aemberling should not be discarded if it is chosen first
- If you can forge a key, you must, and that continues to be enforced. You also cannot overspend

strib (approved):

PR #4863, opened 2026-02-06
fix: draw messaging

### LINKED ISSUE #4862: Logs should show order of draw and shuffle
Comment from player: would like to have: when a player shuffle between draw

for example, a player has only 1 card in hand, so They draw 5. THey draw 2 and that is the end of the deck, so they shuffle before drawing 3 more. I want to see that detail.

Also chains:

I'd like the "draws up to" line to state the correct number of cards. For example, with Daughter on board and 1 chain, here's what I see at the end of my turn:

Tutenkharnage draws 2 cards to their maximum of 7
Tutenkharnage's chains are reduced by 1 to 0

But I think this would be clearer:

Tutenkharnage draws 2 cards to their maximum of 6
Tutenkharnage's chains are reduced by 1 to 0

Because I'm drawing to a max of 6 cards, not 7. As stated in the rulebook, having 1–6 chains means that you "refill your hand to 1 fewer card."

dmikalova:
fix: draw messaging - fixes #4862 

- Updated the messaging to follow the vernacular used in the rulebook
- Refill hand size accounts for chains
- When drawing multiple cards and you have an empty deck, it will log the draws, then shuffle, then further draws in order

Draw with no deck:
```
2026-02-06 10:18:49 player1 attempts to draw with an empty deck, so they shuffle their discard pile to reset their deck
2026-02-06 10:18:49 player1 draws 3 cards to refill their hand to 6 cards
```

Draw with 1 card in deck:
```
2026-02-06 10:20:54 player1 plays Recklessness
2026-02-06 10:20:54 player1 gains an amber due to Recklessness's bonus icon
2026-02-06 10:20:54 player1 uses Recklessness to do several things
2026-02-06 10:20:54 player1 randomly discards Mass Buyout
2026-02-06 10:20:54 player2 randomly discards Stealth Mode
2026-02-06 10:20:54 player2 randomly discards Timetraveller
2026-02-06 10:20:54 player2 randomly discards Rogue Operation
2026-02-06 10:20:54 player1 draws 1 card
2026-02-06 10:20:54 player1 attempts to draw with an empty deck, so they shuffle their discard pile to reset their deck
2026-02-06 10:20:54 player1 draws 2 cards
2026-02-06 10:20:54 player2 draws 3 cards
```

Draw with daughter and 7 chains:
```
2026-02-06 10:31:44 player1 draws 5 cards to refill their hand to 5 cards
2026-02-06 10:31:44 player1 sheds 1 chain to 6 chains
```


server/game/player.js
```diff
@@ -163,30 +163,48 @@ class Player extends GameObject {
...
 
     /**
-     * Called when one of the players decks runs out of cards, removing 5 honor and shuffling the discard pile back into the deck
+     * Called when a player goes to draw with an empty deck. Shuffles the discard pile back into the deck and displays a message in chat
      */
-    deckRanOutOfCards() {
-        this.game.addMessage("{0}'s deck has run out of cards, so they shuffle", this);
+    drawWithEmptyDeck() {
+        this.game.addMessage(
+            '{0} goes to draw with an empty deck, so they shuffle their discard pile to reset their deck',
```
strib: "goes to draw" seems weird to me.  "attempts to draw"?

strib (approved):

server/game/player.js
dmikalova: Funnily enough I originally had attempts to draw but switched to goes when reviewing the rules but also think I got the jelly brain from looking at the words too much. Agree that attempts is less confusing so changed.

strib (approved):

PR #4865, opened 2026-02-06
feat: Add message tests

### LINKED ISSUE #4858: Implement way to check for messaging changes
When working with the core message system, its difficult to tell how changes will effect all the card's messages.

With tests being so fast, it should be possible to checkout master, run all of the tests for every card with debug logs, save that to a tmp file, and then do the same for the current branch and diff them. This would help surface changes to message, assuming the tests run in a deterministic order.

dmikalova:
feat: Add message tests - resolves #4858 

Adds logs for all the standards actions in the game. The tests work by checking every single log message and comparing that to what is given.

The intention is that as log messages are updated, this will catch unexpected changes. For example a recent change causes playing an artifact to get logged twice which went unnoticed. 

Another pattern that happens when changing log messages is that a change may make the logs for one card good, and then another card changes, and when fixing that a third card changes. When fixing that the first card then reverts. At some point in this cycle it becomes easy to either forget to validate or miss a change. This allows for updating the first card - then updating the related messages. Then fixing the second card that changed without reverting the first. etc.

I expect this to be more necessary going forward as I dig into the weeds for fixing logs, as well as updating logs to be more in line with KeyForge vernacular.

Updated the test timeout bc it was making the tests flakey in situations where my computer was busy.


strib (commented):
Could you also add agent guidance to create new tests like this whenever adding a new type of mechanic or action?  And also not to change existing tests unless it is very sure the change is intentional based on user prompting.  (Not sure how well that will work but it's worth a try.)

dmikalova:
> Could you also add agent guidance to create new tests like this whenever adding a new type of mechanic or action? And also not to change existing tests unless it is very sure the change is intentional based on user prompting. (Not sure how well that will work but it's worth a try.)

Added - actually split out the root agents into implementation and tests to have the context in the right spots

strib (approved):

PR #4963, opened 2026-04-29
Pack/dm mars

dmikalova:
> [!NOTE]
> **Medium Risk**
> Mostly additive card scripting plus extensive tests, but it exercises many core engine actions (discard/forge, take control, key cost, conditional multi-step resolution), so mis-specified targeting/then-context could cause gameplay regressions.
> 
> **Overview**
> Adds implementations for a set of `14-DM` Mars cards, covering new *play/action/reap/fight/persistent* abilities such as conditional effects based on `isOverwhelmed()`, key-cost modification while exhausted (Entrench), control changes, bottom-decking creatures, power-counter manipulation, and multi-step ready/use sequencing.
> 
> Adds comprehensive unit tests for each new card, extends `GrammyTaps` tests to cover Entrench behavior during `ReadyPhase`, and updates `HelmsmanSpears`/`KeyDrain` specs with skipped “desired behavior” tests plus a passing test that documents the current engine limitation where cards drawn mid-resolution by Scrap cannot be selected by `mode: 'unlimited'` discard prompts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5e1b8c97467188d31a0d5e6ad0691eaf1ab2cc3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/game/cards/14-DM/AgentBuuff.js
```diff
@@ -0,0 +1,32 @@
+const Card = require('../../Card.js');
+
+class AgentBuuff extends Card {
+    // After Reap: If you are overwhelmed, give a friendly creature three +1 power counters.
+    // Otherwise, give each friendly creature a +1 power counter.
+    setupCardAbilities(ability) {
+        this.reap({
+            alwaysTriggers: true,
+            target: {
+                activePromptTitle: 'Choose a friendly creature',
```
strib: We don't need a special prompt for any other card that asks you to pick a friendly creature, so why this one?  Also, doesn't this make you pick a creature even in the non-overwhelmed case?

server/game/cards/14-DM/ArbiterVyynx.js
```diff
@@ -0,0 +1,31 @@
+const Card = require('../../Card.js');
+
+class ArbiterVyynx extends Card {
+    // After Reap: Destroy a friendly non-Mars creature. If you do, take control of an enemy creature.
+    setupCardAbilities(ability) {
+        this.reap({
+            target: {
+                activePromptTitle: 'Choose a friendly creature',
```
strib: Same with these prompts.

server/game/cards/14-DM/DigitalManipulation.js
```diff
@@ -0,0 +1,31 @@
...
+                target: context.player.opponent.deck.slice(0, 1)
+            })),
+            then: () => ({
+                alwaysTriggers: true,
+                condition: (context) =>
+                    context.preThenEvent &&
+                    context.preThenEvent.card &&
+                    !context.preThenEvent.card.hasHouse('mars'),
+                target: {
+                    activePromptTitle: 'Choose an enemy creature',
```
strib: Same.

server/game/cards/14-DM/EdificeForMars.js
```diff
@@ -0,0 +1,27 @@
+const Card = require('../../Card.js');
+
+class EdificeForMars extends Card {
+    // Play: Destroy a non-Mars artifact. If you do, you may ready a friendly card.
+    setupCardAbilities(ability) {
+        this.play({
+            target: {
+                activePromptTitle: 'Choose an artifact',
```
strib: Ok I'm going to stop commenting on these.

server/game/cards/14-DM/KeyDrain.js
```diff
@@ -0,0 +1,51 @@
...
+            handler: (context) => {
+                context.discardCount = 0;
+
+                const promptDiscard = () => {
+                    if (context.player.hand.length === 0) {
+                        forge();
+                        return;
+                    }
+
+                    context.game.promptForSelect(context.player, {
```
strib: This seems overcomplicated.  Why not copy Helmsman Spears?

server/game/cards/14-DM/ProductiveTrash.js
```diff
@@ -0,0 +1,50 @@
+const Card = require('../../Card.js');
+
+class ProductiveTrash extends Card {
+    // Play: You may discard a non-Mars card. If you do, a friendly creature captures 1A for each bonus icon on the discarded card.
+    setupCardAbilities(ability) {
+        this.play({
+            effect: 'discard a non-Mars card and capture amber for each bonus icon',
+            handler: (context) => {
+                const promptCapture = (remaining) => {
```
strib: Also seems way too complicated -- should just be like Mindfire except you get to choose the target card to discard?

strib (approved):

server/game/cards/14-DM/AgentBuuff.js
dmikalova: Removing the strings.

Based on the tests this doesn't seem to make you choose when non-overwhelmed.

server/game/cards/14-DM/KeyDrain.js
dmikalova: This way handles a one by one discard, which could be a brillix ponder which draws a card that should then be discardable ([discussion](https://discord.com/channels/563737400275107920/588586239997902858/1498893941724938360), although its not consensus)

As such, helmsman spears is actually not behaving correctly in this situation.

server/game/cards/14-DM/ProductiveTrash.js
dmikalova: I don't think its as simple as mindfire - you can optionally choose to discard a card, and then you have to choose where each capture goes. That's 2-7 choices being made (discard?, discard which, capture 0-5 times) whereas mindfire just fires off.

server/game/cards/14-DM/KeyDrain.js
strib: I thought you changed `DiscardAction` to handle all those corner cases already, but I admit I haven't been following all the nitty gritty of these discussions. Should we just fix it there instead?

server/game/cards/14-DM/ProductiveTrash.js
strib: I didn't mean exactly like Mindfire, but a discard with an optional target with a `sequentialForEach` capture (like `BringLow`) in the `then` ability should do it I would think.  Maybe I'm missing something.

server/game/cards/14-DM/KeyDrain.js
dmikalova: I know I updated a bunch of things, for example infernal terran with a discard your whole hand does handle this, but I did not update the selection prompt that spears' uses.

After fixing spears to look like key drain, I realized it would be better to just fix the engine. However its a big change that would affect 7+ cards, so I'd rather split that out. As such I made a [ticket](https://github.com/keyteki/keyteki/issues/4970) and I'm going to implement key drain like spears for now with a note in the tests.

server/game/cards/14-DM/ProductiveTrash.js
dmikalova: That worked!

strib (approved):

PR #5030, opened 2026-05-10
fix: auto-skip redundant single-target prompt for Com. Officer Hings

### LINKED ISSUE #2553: Com. Officer Hings and Com. Officer Gross - Reap prompt
When you reap with Com. Officer Hings or Com. Officer Gross and you have both of them in play, you're prompted to select the other card for seemingly no reason.  Is this a bug?

image

dmikalova:
When reaping with Com. Officer Hings (or similar cards) and there is exactly one legal target, the player was shown a prompt with no real choice. This change skips the prompt automatically in that case unless the player has 'orderForcedAbilities' enabled in their settings.

Fixes #2553

---

> [!NOTE]
> **Medium Risk**
> Changes core targeting resolution flow and can affect many card abilities’ prompt behavior, though it is limited to non-optional `SingleCardSelector` targets and can be disabled via `orderForcedAbilities`.
> 
> **Overview**
> Removes a redundant prompt by **auto-selecting the only legal target** for non-optional, non-`pretarget` abilities that use a `SingleCardSelector`, unless the player has `optionSettings.orderForcedAbilities` enabled.
> 
> Adds tests for `Com. Officer Hings` to verify the target auto-resolves when only one `Com. Officer Gross` is in play, and still prompts when multiple legal targets exist.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 575049c0e484423e144830d8ae22b7e4e7e1a82a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


server/game/AbilityTargets/AbilityTargetCard.js
```diff
@@ -50,6 +51,28 @@ class AbilityTargetCard extends AbilityTarget {
...
 
+        // Auto-resolve when this is a non-optional, non-pretarget single-card target
+        // and there is exactly one legal target. This skips a redundant prompt where
+        // the player has no real choice. Players who prefer to click through every
+        // forced choice can opt out via the `orderForcedAbilities` setting.
+        if (
+            context.stage !== 'pretarget' &&
+            this.selector instanceof SingleCardSelector &&
+            !Optional.EvalOptional(context, this.properties.optional) &&
+            !context.game.activePlayer?.optionSettings?.orderForcedAbilities
```
strib: Curious why this is needed.  If there's only one choice, why not always pick it even if they have ordering enabled?  Without this, does something get messed up if there are other abilities in the same window?

strib (approved):
Approved but curious about the question I asked before I merge.

server/game/AbilityTargets/AbilityTargetCard.js
dmikalova: I just tried this out, and without the orderForcedAbilities there over 300 tests that fail, mostly due to expecting prompts.

Removing it also causes real issues with some tests like wasteful regret - I'm not sure exactly what's going on, but something with the timing window isn't going as expected.

I made a note to investigate further in a future PR if that works?

strib: Sure, thanks.  Sad this isn't more useful to us ability-orderers out there though.

PR #5049, opened 2026-05-13
fix: rematch with new decks race condition

dmikalova:
Fixes #4938

Two race conditions when using Rematch with new decks in sealed:

1. Server: On rematch, player sockets were not joined to the new game channel, so Player B never received the gamestate event. Fix: call socket.joinChannel(newGame.id) for each player.
2. Client: The game socket cleargamestate event could arrive after the lobby socket gamestate for the new pending game, wiping out the new game state. Fix: skip clearGameState dispatch if a pending game already exists.

---

> [!NOTE]
> **Medium Risk**
> Touches rematch/socket lobby synchronization; logic is narrow but wrong timing could still leave stale or missing game UI state.
> 
> **Overview**
> Fixes rematch-with-new-decks races where the second player could miss lobby updates or lose the new pending game in Redux.
> 
> On the **server**, `onGameRematchWithNewDecks` now calls `socket.joinChannel(newGame.id)` for non-owner players after they join—matching normal join/rematch flows so they receive `gamestate` on the new pending game channel.
> 
> On the **client**, the game socket `cleargamestate` handler no longer always dispatches `clearGameState`. It clears only when there is no `lobby.currentGame` or when `currentGame.started` is still true (the finished game). If a pending rematch (`started: false`) already landed via the lobby socket, the late clear is skipped so the new game isn’t wiped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 839c8c1a9d159c4f5f270c1aec10c8c967eafafb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryogen:
Was this just done with ai or in response to an actual confirmed reproducible issue? I have my doubts

dmikalova:
> Was this just done with ai or in response to an actual confirmed reproducible issue? I have my doubts

The fix was done with AI in response to a notorious issue where like half the time rematch does not work and one person gets dropped to the lobby and can't join the new game the other person is in. Its a race condition so its not consistently reproducible, but I have experienced this a lot with friends, and have heard of people experiencing it a lot, and from my tests this seems to fix it.

client/redux/middleware/socket-middleware.js
```diff
@@ -271,7 +271,13 @@ export const socketMiddleware = (store) => (next) => (action) => {
         });
 
         gameSocket.on('cleargamestate', () => {
-            store.dispatch(lobbyActions.clearGameState());
+            const currentState = store.getState();
+            // Don't clear game state if the lobby has already set up a new pending game
+            // (e.g., from a rematch with new decks). The game socket's cleargamestate
+            // can race with the lobby socket's gamestate for the new pending game.
+            if (!currentState.lobby.currentGame || currentState.lobby.currentGame.started) {
```
strib, line 293: Does the second condition imply that, since the current game has already started, it must refer to the game we want to clear (rather than the new game we're racing against)? Maybe make the comment a little more clear.

Also, for my education since I don't know the client code well: is there any danger of a leak here if a new game has started and we don't clear the previous one?

dmikalova: Yeah, the currentGame.started is the game that has already finished which should be cleared for the new game.

Here's the response on risk of leak:

Leak risk
Practically none.

- The slot that gets "kept" is currentGame, which holds a small game-summary object. In the only case we skip, it's already pointing at the new pending game (that's the whole reason we're skipping), so there's no stale reference.
- rootState (the heavy board data) isn't touched by clearGameState at all. It gets overwritten when the next game's gamestate arrives, and fully cleared on gameSocketClosed / gameSocketDisconnected.
- The worst case is: user leaves the page right after this skip, before the next gamestate arrives. In that case currentGame references the new pending game until the tab is closed — bounded, GC'd on unload, not a leak.

strib (approved):

PR #5072, opened 2026-05-18
fix: resolve popup loop and DnD invariant crash on card remount

dmikalova:
## Summary

Fixes a client-side popup feedback loop surfaced via Sentry (issue 7246736959).

### 1. Popup update loop in `CardPileLink`

`onPopupChange` was called unconditionally inside `updatePopupVisibility` on every render cycle, causing repeated `showDrawDeck` game messages to fire in a loop. Fixed by:
- Moving `onPopupChange` into a `useEffect` that tracks previous `showPopup` via `useRef`, firing only when the value actually changes.
- Guarding the selectable-cards effect to skip `updatePopupVisibility` when the desired visibility already matches current state.

### 2. Removed draw-deck popup server wiring

The `onDrawPopupChange` → `handleDrawPopupChange` → `sendGameMessage('showDrawDeck', ...)` chain in `GameBoard` / `PlayerStats` / `DrawDeck` was the source of server message spam. Removed the entire prop-threading chain since open/close of the draw deck popup no longer needs to notify the server on every toggle.

---

> [!NOTE]
> **Low Risk**
> Localized React state and prop cleanup on the game board; no auth, data, or server protocol changes beyond ceasing draw-deck popup sync.
> 
> **Overview**
> Stops a **client popup feedback loop** by changing how `CardPileLink` notifies parents and by dropping draw-deck server sync on every toggle.
> 
> In **`CardPileLink`**, `onPopupChange` no longer runs inside `updatePopupVisibility` on every call. It runs in a **`useEffect`** only when `showPopup` actually changes (tracked with a ref). The auto-open effect for selectable cards / prompt targets now updates visibility **only when** the desired state differs from the current `showPopup`, avoiding redundant state updates.
> 
> The **`showDrawDeck` socket path** is removed: `GameBoard` no longer sends `showDrawDeck` on draw popup changes, and the `onDrawPopupChange` / draw-specific `onPopupChange` wiring through **`PlayerStats`** and **`DrawDeck`** is deleted. Draw deck popups stay local to the UI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4fa55bc301b704f5623ee556dd6ca5e9900505a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryogen:
I disagree with the bot on what the issue was, as looking at the code that couldnt have happened that way. Would be interested in repro steps for this to see if there isnt a simpler fix

PR #5086, opened 2026-05-21
fix: ability targeting prompt layout with many target cards

dmikalova:
When there are many target cards, they overflow the targeting prompt container. This wraps target cards in a flex row that uses a CSS custom property (`--target-count`) to compute negative margins so cards overlap cleanly when they don't fit side-by-side.

**Changes:**
- `AbilityTargeting.jsx`: wrap target cards in `.ability-target-row`, make container full-width
- `tailwind.css`: add `.ability-target-row` component with overlap calculation

Previously with too many targets this would just have the cards side by side and go off out of bounds.
image

Now they squish:
image

---

> [!NOTE]
> **Low Risk**
> UI layout and prompt metadata only; no auth, payments, or core game rules beyond which card summaries are sent to the client.
> 
> **Overview**
> Fixes **ability targeting prompts** when many cards are shown: the client now uses a **full-width row** and **overlapping target thumbnails** (negative spacing via `calc` on `marginLeft`) so targets stay inside the prompt instead of overflowing.
> 
> On the server, **targeting prompt payloads** now treat **`context.subject`** like an explicit target—when building controls for forced triggered abilities and when inferring fallback targets in handler menus and card-select prompts—so multi-subject / subject-based abilities show the right cards in the UI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3666b69148fa78148186624142aafe7d73591388. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cryogen:
I'm sure they used to got his, the move to tailwind probably broke it!

client/styles/tailwind.css
```diff
@@ -2131,4 +2131,18 @@ pre {
     .blocklist td {
         vertical-align: middle !important;
     }
+    
+    /* Ability targeting prompt: spread up to --target-count cards (each w-16)
+     across the remaining space, overlapping as needed when they don't fit. */
+    .ability-target-row {
```
cryogen: Given this is only used in one place, can this be replaced with direct tailwind classes?

dmikalova: Updated - ended up in a new PR
https://github.com/keyteki/keyteki/pull/5123

PR #5092, opened 2026-05-21
fix: block list quick join and live lobby refresh

### LINKED ISSUE #2766: Block list doesn't work with quick join
I have a user in my block list, and while I can't see their chat or games in the list, if I use quick join I will still join their games. Using exact capitalization doesn't help either.

dmikalova:
## Summary

Fixes two related block list issues:

1. **Quick join ignored block list** — Quick join now calls `game.isVisibleFor(socket.user)` so it skips games hosted by (or containing) blocked users.

2. **Lobby didn't refresh after block list change** — Adding/removing a block now emits `onBlocklistChanged`, which triggers a live refresh of the user list and game list for the affected socket, so changes take effect immediately without a page reload.

Fixes #2766

---

> [!NOTE]
> **Medium Risk**
> Touches rematch/socket Redux timing and bonus-icon resolution paths used during play; changes are localized but affect live game and lobby handoff behavior.
> 
> **Overview**
> This PR bundles several **gameplay**, **client**, and **lobby** fixes beyond the block-list work described in the PR text (those changes are not in the attached diff).
> 
> **Bonus icons:** House-name enhancements (e.g. `untamed` on a card) are no longer treated as resolvable bonus icons. `Card.getResolvableBonusIcons()` filters them out, and `ResolveBonusIconsAction` only loops over that list—fixing bogus “resolve this untamed icon?” prompts when effects like Amphora Captura replace icons.
> 
> **Ditch the Loot:** Play now requires at least two creatures in play; with only one creature, amber stays put instead of breaking the move flow. Specs cover friendly/enemy transfers and the single-creature edge case.
> 
> **Rematch / client state:** `cleargamestate` on the game socket only clears `lobby.currentGame` when the slot still holds the finished game (`started: true`) or is empty—so a pending rematch that arrived first on the lobby socket is not wiped. Rematch-with-new-decks joins non-owner players to the new game channel via `socket.joinChannel`. Chat keyword rendering avoids a trailing space after the last token.
> 
> **Tests:** End-turn prophecy and tide specs are consolidated under `extra-cards/`; tide drops the unenforced Dark Tidings deck gate and a skipped test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb8cedb1dd8162ea464635834a43f15d6a7b4fb1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


strib (approved):

server/lobby.js
```diff
@@ -1203,6 +1204,15 @@ class Lobby {
         }
 
         updatedUser.blockList = user.blockList;
+
+        // Re-send the lobby state so the user's game/user list (which is
+        // filtered by their block list) reflects the change without
+        // requiring a page refresh.
+        const socket = this.socketsByName[user.username];
+        if (socket) {
+            this.sendUserListFilteredWithBlockList(socket, this.getUserList());
```
cryogen, line 1213: I know this is already merged, but this is very wasteful. It would have been better to try and 'diff' the users/games and just send the client removes for them

dmikalova: I'll get that in

PR #5027, opened 2026-05-10
fix: Mirror Shell copying token shows correct card identity

dmikalova:
fixes #3484 fixes #4917 fixes #2908 fixes #3280

When Mirror Shell copies a creature name onto a token card, `tokenCard()` was looking up by name. This could fail when a copyCard effect overrides `this.name`, causing the wrong token definition (or no token) to be returned.

**Fix:** resolve the token directly from `this.owner.tokenCard` instead of searching all players by name.

**Test:** adds a regression test to `MirrorShell.spec.js` verifying that a Grunt token copied via Mirror Shell retains correct identity and abilities.

> [!NOTE]
> CSS card ring highlighting split into a separate PR: #5113

---

> [!NOTE]
> **Medium Risk**
> Touches core card resolution, house availability, and UI state for all tokens; behavior change is targeted but affects shared `Card`/`Player` paths.
> 
> **Overview**
> Fixes token handling when **Mirror Shell** (and similar `copyCard` effects) copy a creature onto a token without breaking game logic or UI.
> 
> **`tokenCard()`** now returns the owning player’s deck token definition (`owner.tokenCard`) instead of scanning all players by `this.name`, which could resolve the wrong card or nothing after `copyCard` overrides the displayed name.
> 
> **House selection** for in-play cards uses `card.getHouses()` on the live instance so tokens respect `copyCard` and enhancements; out-of-play tokens still fall back to the printed token definition.
> 
> **Client summary** sets `copying` only when a non-baseline `copyCard` is active (not the default token self-copy from `MakeTokenCreatureAction`).
> 
> A **Mirror Shell** regression test asserts a Grunt token keeps id/name **Grunt** while reaping with copied abilities.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d3f56cd2f6af9a9514f05f611660df049fcf6d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5113, opened 2026-05-25
feat: add copying card state visual indicator

dmikalova:
Also updated the card border ring:
- If a card is a copy, then it will have a violet ring (mimic gel, mirror shell)
- Removed the token creature ring - does not seem necessary and the gold was very close to the change control yellow
- The highest priority color takes precedence (ready green, copy purple, control yellow)
- Minor tweaks to the size and shape of the ring to match the card shapes

image

---

> [!NOTE]
> **Low Risk**
> Presentation and summary-flag logic only; no auth, payments, or core game rules beyond a narrow `copying` false-positive fix.
> 
> **Overview**
> Adds a **violet ring** on board cards when the server reports `copying` (identity overridden by something other than baseline token/flip self-`copyCard`).
> 
> **Server:** `getSummary` treats `copyEffect === this` like token baseline copies so flip/un-tokenize does not show as copying.
> 
> **Client:** `Card.jsx` applies a `copying` CSS class; card chrome is refactored to a single **ring** via `--c-ring` / `--ring-w` (replacing stacked box-shadow combos). **Token** gold ring styling is removed; exhausted card image offsets use dimension-based centering instead of magic percentages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 612a215fdf31c2d7a6194459d1c11effe1e8eace. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dmikalova:
Draft until the following are merged:

https://github.com/keyteki/keyteki/pull/5112
https://github.com/keyteki/keyteki/pull/5027

server/game/Card.js
```diff
@@ -1481,11 +1481,13 @@ class Card extends EffectSource {
...
             // MakeTokenCreatureAction); ignore that self-copy and only flag
             // `copying` when something else (e.g. Mirror Shell, Mimic Gel)
-            // overrides the card's identity.
+            // overrides the card's identity. FlipAction uses `copyCard(card)`
+            // when un-tokenizing, which is also a baseline self-reference.
             copying: (() => {
                 const copyEffect = this.mostRecentEffect('copyCard');
                 if (!copyEffect) return false;
                 if (this.isToken() && copyEffect === this.tokenCard()) return false;
+                if (copyEffect === this) return false;
```
cryogen: Would be good to fix these missing braces while you're here and not add a new one

PR #5115, opened 2026-05-26
fix: re-handoff lobby users to game after gamenode reconnect

dmikalova:
When a gamenode reconnects to the lobby, re-send the handoff for any players/spectators whose game lives on that node. Without this, players stranded in the lobby after a gamenode restart never get bounced back into their game — they had to refresh or re-login to trigger `doPostAuth`.

---

> [!NOTE]
> **Low Risk**
> Small, targeted change reusing existing handoff logic; no new auth model, mainly improves recovery after node restarts.
> 
> **Overview**
> When a **gamenode** reconnects, `onNodeReconnected` already re-syncs lobby game state from the node but did not push clients back onto the game server. This adds a pass after that sync: for every **started** game on the reconnected node, it calls `sendHandoff` for any **players and spectators** still connected to the lobby.
> 
> That closes the gap where users stayed in the lobby after a gamenode restart until they refreshed or re-authenticated (the only other paths that triggered handoff). Handoff uses the same fresh short-lived JWT as normal start/watch flows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90e5522ee1bfb566089beec3e108cb7655f25b54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/server/lobby.js
+++ b/server/lobby.js
@@ -1299,27 +1299,30 @@ class Lobby {
             }
         }
 
-        // Re-handoff any logged-in users whose started game lives on the
-        // just-(re)connected node. Without this, players stranded in the
-        // lobby after a gamenode restart never get bounced back into
-        // their game — they have to refresh or re-login to trigger the
-        // doPostAuth handoff. The handoff payload is a fresh
-        // JWT and is safe to re-issue if the client is already on /play.
+        // Re-handoff only when this game is still the user's active lobby game,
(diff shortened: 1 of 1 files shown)


cryogen:
That’s not how this is supposed to work. Ai is conflating two things here. This will cause more harm than good

PR #5120, opened 2026-05-28
[WIP] fix: reverse opponent battleline card ordering to match visual expectation

dmikalova:
Reverses the card order in the opponent's battleline rows when `rowDirection === 'reverse'`, so the visual ordering matches the expected left-to-right layout from the opponent's perspective.

image

---

> [!NOTE]
> **Low Risk**
> Small presentation-only change in `PlayerBoard` with no game logic or data flow impact.
> 
> **Overview**
> When rendering the opponent play area (`rowDirection === 'reverse'`), each battleline row now maps cards through a **reversed copy** of `row.cards` before rendering, so left-to-right order matches what players expect from the opponent’s side.
> 
> Your own board (`default`) is unchanged and still uses the original card order.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b1d8d121459a7a2da78bd99352b3c98df45f382. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/Components/GameBoard/PlayerBoard.jsx
+++ b/client/Components/GameBoard/PlayerBoard.jsx
@@ -54,29 +54,33 @@ const PlayerBoard = (props) => {
                 source='play area'
                 manualMode={props.manualMode}
             >
-                {rows.map((row, index) => (
-                    <div className={`card-row ${row.name}`} key={`card-row-${index}`}>
-                        {row.cards.map((card) => (
-                            <Card
-                                key={card.uuid}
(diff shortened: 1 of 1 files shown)


dmikalova:
Putting in draft bc placing creatures in opponent's battleline is the opposite of what you would expect (eg Malison)

Also need to consider with messaging updates what left/right means for flank positioning

PR #5131, opened 2026-05-30
fix(game): reset game owner when player leaves started game

dmikalova:
This PR resolves game message ownership issues by ensuring that game ownership is correctly cleaned up and reset when a player leaves a game that has already started. This fixes the root cause of potential null/undefined owner state on the client side, rather than masking it.

### Key Changes

* **Server (`server/pendinggame.js`)**: Moved the execution of `removeAndResetOwner(playerName)` outside of the `!this.started` condition in the `leave(playerName)` method, ensuring ownership is reset unconditionally when a player leaves.
* **Client (`client/Components/GameBoard/Messages.jsx`)**: Reverted optional chaining on `owner?.name` to `owner.name` in message class calculations to surface any state inconsistencies directly rather than hiding them.


cryogen:
This shouldnt happen so this change will hide a legit bug

dmikalova:
> This shouldnt happen so this change will hide a legit bug

Good catch, updated to fix the root cause (owner leaving game does not set a new game owner) rather than masking.

strib (approved):

PR #5147, opened 2026-06-12
feat: support house enhancements in card house checks

### LINKED ISSUE #5146: A Fair Game discarding a house enhanced card does not count both houses.
I used a Fair Game to discard an untamed enhanced shadows card but only gained aember for the shadows cards in my opponents hand, while I expected to get aember for both the shadows and the untamed card. This could be the correct behavior, but was playing with David M. who also believed it was incorrect behavior.

dmikalova:
Fixes #5146 

This pull request updates the game engine and card logic to respect house enhancements when performing card house checks. Instead of reading only the printed house of a card, the implementation now queries all active houses of a card via `getHouses()`, supporting cards that possess multiple houses due to enhancements or effects.

### Key Changes
* **Game Engine Core**:
  * Updated `Card.getHouses()` and `Player` house-checking logic to flatten results from `changeHouse` effects, supporting multi-house cards.
* **Card Rules Updates**:
  * **A Fair Game**: Refactored card checks to evaluate all houses returned by `getHouses()` of the top card of the deck rather than just its printed house.
  * **Ol' Paddy (Evil Twin)**: Updated the card-matching condition to check all houses of the discarded card.
  * **Immortal Greking**: Stolen creatures now gain all houses of Greking (via `changeHouse(context.source.getHouses())`), and the card's effect text formats multiple houses using "and".
  * **Into the Warp**: Uses `flatMap` on `getHouses()` to match creatures in play with any of the houses of discarded cards.
* **Testing**:
  * Added regression and unit tests for *A Fair Game*, *Ol' Paddy (Evil Twin)*, *Immortal Greking*, and *Into the Warp* verifying successful interaction with house enhancements.
  * Added mock implementation of `getHouses()` in `cancreatecards.spec.js`.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit d6478ecbc7f18b15da7277a894a668886edfc1dd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/server/game/Card.js
+++ b/server/game/Card.js
@@ -685,7 +685,7 @@ class Card extends EffectSource {
         let combinedHouses = [];
 
         if (this.anyEffect('changeHouse')) {
-            combinedHouses = combinedHouses.concat(this.getEffects('changeHouse'));
+            combinedHouses = combinedHouses.concat(this.getEffects('changeHouse').flat());
         } else {
             let copyEffect = this.mostRecentEffect('copyCard');
             combinedHouses.push(copyEffect ? copyEffect.printedHouse : this.printedHouse);
--- a/server/game/cards/01-Core/AFairGame.js
+++ b/server/game/cards/01-Core/AFairGame.js
@@ -19,12 +19,18 @@ class AFairGame extends Card {
                     oppTop || 'nothing',
                     oppHand,
                     oppTop
-                        ? oppHand.filter((card) => card.hasHouse(oppTop.printedHouse)).length
+                        ? oppHand.filter((card) =>
+                              oppTop.getHouses().some((house) => card.hasHouse(house))
+                          ).length
                         : 0,
                     context.player,
                     myTop || 'nothing',
                     myHand,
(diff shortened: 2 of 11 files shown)


server/game/Card.js
```diff
@@ -685,7 +685,7 @@ class Card extends EffectSource {
         let combinedHouses = [];
 
         if (this.anyEffect('changeHouse')) {
```
dmikalova, line 687: If Immortal Greking has multiple houses, then its change house ability should give the creature all of those houses - hence the flat here.

This is meant to be backwards compatible with the current combineHouses.concat() - which combines all changeHouse effects. However, that is not correct, the most recent changeHouse should win. - ticket [here](https://github.com/keyteki/keyteki/issues/5148).

Using .flat() would also set up Greking to forwards compatible with the fixes, rather than applying 2 change house effects in a row.

strib (approved):

PR #5156, opened 2026-06-13
fix(game): prevent cardsInPlay mutation during control updates

dmikalova:
Prevent mutating `player.cardsInPlay` during traversal when card control updates occur, and add safety checks for the new controller's state.

### Key Changes
- **Game Engine (`server/game/game.js`)**:
  - Replaced `_.each` with a `for...of` loop over a copy (`[...player.cardsInPlay]`) to avoid traversal issues when `this.takeControl` modifies the source array in place.
  - Added validation checks to ensure `newController` exists and has `cardsInPlay` defined before transferring control.
- **Card Tests (`test/server/cards/06-WoE/TheAemberRoad.spec.js`)**:
  - Added a readiness assertion `expect(this.player2).isReadyToTakeAction();` to verify the state flow of "The Æmber Road" control transfer.

---

> [!NOTE]
> **Low Risk**
> Small, localized engine loop fix with a regression test; doc edits are cosmetic.
> 
> **Overview**
> **`checkGameState`** no longer walks `player.cardsInPlay` with `_.each` while `takeControl` can remove cards from that array mid-loop. It now uses `for...of` over `[...player.cardsInPlay]` so every in-play card is checked once when modified-controller effects flip ownership (e.g. trade/control artifacts).
> 
> The **The Æmber Road** spec adds `expect(this.player2).isReadyToTakeAction()` after the second omni use and control swap, guarding that the game pipeline still offers a legal next action.
> 
> **Docs** (`card-abilities.md`, `implementing-cards.md`) only adjust markdown list/table alignment; no rule or API text changes beyond whitespace.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90e7452a369924867788625e59fc3a688d6fdbbd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


test/server/cards/06-WoE/TheAemberRoad.spec.js
```diff
@@ -31,6 +31,7 @@ describe('The Æmber Road', function () {
             expect(this.player2.amber).toBe(4);
             expect(this.player1.player.cardsInPlay).toContain(this.theÆmberRoad);
```
dmikalova, line 32: This crash was being caused by aember road in a one player game - however instead of going for a player guard in aember road, the AI noticed that the way _.each works is with an indexed based traversal that would desync in this situation. So now aember road and other cards that work like this will still do as much as they can (eg add a counter), but don't error out on change control in a 1 player game.

All that being said, wasn't sure the best way to test this, since the test harness defaults to 2 players. Tested manually and it works.

strib (approved):

Issue #5160, opened 2026-06-13
My Blorb returns Blorb Hive from Opponents Discard to Their Hand


akomar08:
My opponent was playing a Blorb deck and ended up putting Blorb Hive in the discard early on and then giving me an enraged blorb. I was playing an Aember Skies and thus did not have my own Blorb Hive. But I did have Mars so I fought the Blorb into an enemy creature and after I was prompted to select a Blorb Hive from a discard. I selected my opponents as it appeared to be the only way to get out of the screen and it returned it to their hand.

PR #5185, opened 2026-06-22
fix: Blorb incorrect retrieval (#5160)

akomar08:
The blorb destroyed affect now only checks the discard of the controller instead of any discard.


dmikalova (approved):

dmikalova:
Fixes #5160

PR #5202, opened 2026-07-21
feat: change adaptive to auto pick first deck

### LINKED ISSUE #5167: Change Adaptive Bo1 setup to match official GG setup
Per version 1.0 of the Alternative Formats Guide, the official setup procedure for Adaptive is as follows:
1. Randomly determine which **deck** (not which player) will take the first turn.
2. Reveal decklists. Both players will take 3-5 minutes to asses the decklists.
3. Both players secretly select which deck they would like to play for the match, then reveals their choices to each other simultaneously.
4. If both players chose a different deck, the match begins with the deck that was determined to take the first turn. If both players choose the same deck, they must start bidding.

keyteki currently handles 2-4 and the bidding procedure without issue, however, 1 is not handled at all and instead uses picks a player to go first after 2-4 and bidding is complete.

The AdaptiveDeckSelectionPrompt is entirely separate from the FirstPlayerSelection but happens beforehand and could set **game.activePlayer**. With game.previousWinner unset, the onCompleted() on line 54 should handle properly assigning the first player coming from AdaptiveDeckSelectionPrompt in FirstPlayerSelection.js.

The message saying which deck is going first needs to happen before players have to pick decks from the activePrompt() in AdaptiveDeckSelectionPrompt. The setting of game.activePlayer would need to happen either in the else condition at line 72 in AdaptiveDeckSelectionPrompt.js or the if condition on line 31 of ChainBiddingPrompt.js.

This change would have adaptive players singing the praises of TCO I have no doubt!

Moosemandude:
Fixes #5167 

## Summary
Changes the behavior of the Adaptive Best-of-1 game mode to match official GG procedure by picking a first deck instead of a first player.

## Key Changes
- Adds an `adaptiveFirstDeck` variable to Game which tracks the deck that is chosen to go first. Used as a check in a few spots to opt-in to adaptive specific behavior when needed.
- Add `firstdeck()` to AdaptiveDeckSelectionPrompt.js which is used in the constructor to choose a first deck before the message needs to be displayed to the players.
- Adds a handler to FirstPlayerSelection.js that will set `this.game.activePlayer` before it is randomly decided if `adaptiveFirstDeck` is set.
- Adds an alert to display which deck will go first after the messages that state which player brings what deck to the crucible in setupphase.js.


server/game/gamesteps/setup/AdaptiveDeckSelectionPrompt.js
```diff
@@ -8,6 +9,7 @@ class AdaptiveDeckSelectionPrompt extends AllPlayerPrompt {
         this.adaptive = game.adaptive;
         this.clickedButton = {};
         this.players = game.getPlayers();
+        this.firstDeck();
```
strib: I think this might run for every game type, thus always setting `adaptiveFirstDeck` and affecting the completion condition for every game.  Maybe check `this.game.gameFormat` before calling this?

Moosemandude: I definitely missed how that section is running every game! I've added the check to make sure the format is `adaptive-bo1` and confirmed that the code now only executes in adaptive games and not in standard games.

